PR #2, opened 2016-10-20
Show contacts without fn

irgendwie:
This firstly removes the property filter for syncing addressbooks, meaning contacts without fn are now also synced.
Secondly, the value of n was automatically joined and inserted into the visual fullname, this is now visually improved.

@Henni @jancborchardt

### DIFF
--- a/js/dav/dav.js
+++ b/js/dav/dav.js
@@ -2832,7 +2832,7 @@ var _prop = require('./prop');
 var _prop2 = _interopRequireDefault(_prop);
 
 function addressBookQuery(object) {
-  return '<card:addressbook-query xmlns:card="urn:ietf:params:xml:ns:carddav"\n                          xmlns:d="DAV:">\n    <d:prop>\n      ' + object.props.map(_prop2['default']) + '\n    </d:prop>\n    <!-- According to http://stackoverflow.com/questions/23742568/google-carddav-api-addressbook-multiget-returns-400-bad-request,\n         Google\'s CardDAV server requires a filter element. I don\'t think all addressbook-query calls need a filter in the spec though? -->\n    <card:filter>\n      <card:prop-filter name="FN">\n      </card:prop-filter>\n    </card:filter>\n  </card:addressbook-query>';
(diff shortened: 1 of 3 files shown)


irgendwie:
Fixes some bugs:

https://github.com/owncloud/contacts/issues/485
https://github.com/owncloud/contacts/issues/403

MariusBluem:
This is very cool 👍 ... I'll love to see my macOS adressbook without all these "Without Title" contacts    😜

jancborchardt:
Can you review if this fixes the issues you encountered? :) @bart-1 @steinjer @aspdye @skjnldsv

skjnldsv:
This is a bad Idea! Than means we allows non valid vcards.
Right now we never allowed such thing, and made sure we couldn't generate vcard that are not rfc-compliant. I haven't tested it right now, but we can't allow that in our system.

steinjer:
@skjnldsv: No! It is a bad idea to delete my imported vcards without any warning or error! 
Many client programs ignore the "non valid vcard" and handle empty fn properties. 
If the fn property is necessary, the contact app could fill it with default value like "formatted text is empty".

skjnldsv:
@steinjer It's not deleted, it's juste not synced.

Henni:
@skjnldsv I think we should apply the robustness principle:
> be conservative in what you do, be liberal in what you accept from others.

So in my opinion we should accept "broken" vcards without fn during an import or from webdav (if another client uploaded them) and display them in our app. But we should never create such a vcard ourselves. This means our frontend shouldn't allow to submit a new contact if the fn field isn't filled out.

This should also apply for all other fields, btw. For example this means we should try to handle broken date formats, but always create correct date formats ourselves. (I know that this currently isn't the case, thanks to our vcard lib)

skjnldsv:
@irgendwie Right now with your pr, what does the contact look like in the app if it doesn't have a FN field set?

irgendwie:
@skjnldsv Since we already have https://github.com/nextcloud/contacts/commit/ed1659dd49d475fcf554ad3c3d0bddb9b25b20c5, contacts without a fullname would be loaded, the contents of name would be concatinated and stored in fullname.
So this PR would simply load contacts without a fullname, since we obviously need to take care of this case.

skjnldsv:
In this case bug 👍 🎉 
But we should really be careful about broken vcard anyway :)

PR #8, opened 2016-10-24
Sabre dav 3.2 requires on each vcard to have fn set

MariusBluem:
Downstream from https://github.com/owncloud/contacts/pull/526

cc @irgendwie @Henni

### DIFF
--- a/js/services/contact_service.js
+++ b/js/services/contact_service.js
@@ -93,6 +93,9 @@ angular.module('contactsApp')
 		newContact.uid(newUid);
 		newContact.setUrl(addressBook, newUid);
 		newContact.addressBookId = addressBook.displayName;
+		if (_.isUndefined(newContact.fullName()) || newContact.fullName() === '') {
+			newContact.fullName(t('contacts', 'New contact'));
+		}
 
 		return DavClient.createCard(
 			addressBook,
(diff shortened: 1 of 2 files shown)


MariusBluem:
Anyone 🏜 @nextcloud/contacts

js/services/contact_service.js
```diff
@@ -93,6 +93,9 @@ angular.module('contactsApp')
 		newContact.uid(newUid);
 		newContact.setUrl(addressBook, newUid);
 		newContact.addressBookId = addressBook.displayName;
+		if (_.isUndefined(newContact.fullName()) || newContact.fullName() === '') {
+			newContact.fullName(t('contacts', 'New Contact'));
```
jancborchardt: This should be capitalized like »New contact« – we always only capitalize the first word. ;)

raghunayyar:
@MariusBluem everything else seems okay. Can you add @jancborchardt changes so we can merge it. :)

MariusBluem:
What about the change in the contact_model 😁 https://github.com/owncloud/contacts/pull/526/files

js/services/contact_service.js
MariusBluem: So it should be `New.Contact.fullName` everywhere 🤔 ... this would also affect lines which are not touched by this PR - so I think I am misunderstanding you @jancborchardt

rullzer:
This fixes #29 

LGTM

PR #23, opened 2016-10-29
added headings

PatrycjaPuczko:
I put headings for the sections to make form more readable for users. 
@nextcloud/contacts How can a get the real names for the headings instead of using {{prop.$key}?

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -230,7 +230,7 @@ detailsitem.details-item-adr .icon-delete, detailsitem.details-item-n .icon-dele
 }
 
 detailsitem.details-item-adr {
-	padding: 20px 0;
+	padding: 0;
 }
 
 detailsitem.details-item-n {
--- a/templates/contactDetails.html
+++ b/templates/contactDetails.html
@@ -30,6 +30,7 @@
 		</header>
 		<section>
 			<div ng-repeat="prop in ctrl.contact.props | toArray | orderDetailItems:'$key'">
+				<h3>{{prop.$key}}</h3>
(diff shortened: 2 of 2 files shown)


Henni:
I believe you can at the moment only hotfix this with `{{prop.$key}}`.
My suggestion how to solve this cleanly would be to:

1. create a new angular directive (`detailgroup`) which would encapsulate all `detailitem`s of the same type and display the heading (which is currently only accessible inside the `detailsitem` directive as `{{ctrl.meta.readableName}}`) or
2. restructure the `detailitem` directive to list all entries of a single type.


"Graphically" this means:
**Current State:**
```
contact
|--detailitem (Name)
|--detailitem (Phone Work)
+--detailitem (Phone Home)
```

**Suggestion 1:**
```
contact
|-- detailgroup (Name)
|   +-- detailitem (Name)
+-- detailgroup (Phone)
    |--detailitem (Phone Work)
    +--detailitem (Phone Home)
```

**Suggestion 2:**
```
contact
|-- detailitem (Name)
+-- detailitem (Phone Work + Home)
```

cc @irgendwie thoughts?

jancborchardt:
@Henni @irgendwie do we want to check this out at the next meetup maybe? :)

jancborchardt:
@Henni suggestion 1 seems most fitting, right? @nextcloud/contacts is some of you down for implementing that?

skjnldsv:
We'll need a full restructuration of the details.
Since no activity, closing. We'll work on a full rewriting later :)

jancborchardt:
Why a full restructuring @skjnldsv? :) Can’t we do this step by step? And this is already a good step in the right direction. We just need the actual text value.

jancborchardt:
@Henni let me know as soon as you have the follow-up pull request to this open. :)

PR #35, opened 2016-11-20
change version to current released version

enoch85:
As discussed before. @irgendwie


irgendwie:
We released version 1.5.0, everything should be correct now! :wink:

enoch85:
Finally ;) Thanks! :D

PR #40, opened 2016-11-25
Disable autocorrection

skjnldsv:
No more this!
capture d ecran_2016-11-25_16-07-38

### DIFF
--- a/templates/contactDetails.html
+++ b/templates/contactDetails.html
@@ -11,14 +11,14 @@
 			<avatar data="ctrl.contact" ng-class="{'avatar--missing': ctrl.contact.photo()===undefined}"></avatar>
 			<div id="details-contact-infos">
 				<h2>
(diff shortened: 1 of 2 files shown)


MariusBluem:
I agree with that 👍

irgendwie:
This would leave auto-correction for 'detailed name' and various other fields
and only affect the contact header, correct?
Do we really want to disable this "feature"? What is the common practice for this?
cc @jancborchardt

skjnldsv:
Yes, this is correct. Maybe we should indeed extend it to more values. Like phones, detailed names (which I forgot we had ^^)...

irgendwie:
@skjnldsv It seems logical for me to expand the disabled auto-correct to detailed name but not more fields. Which means name and organization in the header and detailed name in the details view. Should be enough for now - would you mind making the required changes?

skjnldsv:
Your wish is my command! 🚀

irgendwie:
LGTM :+1:

jancborchardt:
Good detail fix @skjnldsv! :+1:

PR #43, opened 2016-11-29
Fix export and version comparison

### LINKED ISSUE #39: Currently not possible to download group of contacts.
At the moment it is not possible to select a single group of contacts and export/download it as a vcard file.
This used to be possible with the older contact app (last year) 
Can this feature be re-added again?

Regards,

Bert

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39513291-currently-not-possible-to-download-group-of-contacts?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
We can have the export back! 🎉 

fix #39 
@irgendwie @brjhaverkamp

### DIFF
--- a/js/components/addressBook/addressBook_controller.js
+++ b/js/components/addressBook/addressBook_controller.js
@@ -4,8 +4,22 @@ angular.module('contactsApp')
 
 	ctrl.showUrl = false;
 	/* globals oc_config */
+
+	function compareVersion(version1, version2) {
+		for(var i = 0; i < Math.max(version1.length, version2.length); i++) {
+			a = version1[i] || 0;
+			b = version2[i] || 0;
+			if(Number(a) < Number(b)) {
+				return true
+			}
+			if(version1[i] !== version2[i]) {
+				return false
+			}
+		}
+		return false;
+	}
 	/* eslint-disable camelcase */
-	ctrl.canExport = oc_config.version.split('.') >= [9, 0, 2, 0];
+	ctrl.canExport = compareVersion([9, 0, 2, 0], oc_config.version.split('.'));
 	/* eslint-enable camelcase */
 
 	ctrl.toggleShowUrl = function() {


js/components/addressBook/addressBook_controller.js
```diff
@@ -4,8 +4,22 @@ angular.module('contactsApp')
 
 	ctrl.showUrl = false;
 	/* globals oc_config */
+
+	function compareVersion(version1, version2) {
+		var result=false;
+		for(var i=0;i<(Math.max(version1.length, version2.length));i++) {
```
ChristophWurst: shouldn't it be Math.min? You don't want to access an index greater than the string length, do you? :-)

js/components/addressBook/addressBook_controller.js
```diff
@@ -4,8 +4,22 @@ angular.module('contactsApp')
...
+			if(Number(version1[i])<Number(version2[i])) {
+				result=true;
```
ChristophWurst: simple ``return true;``?

js/components/addressBook/addressBook_controller.js
```diff
@@ -4,8 +4,22 @@ angular.module('contactsApp')
...
+				break;
+			}
+			if(version1[i]!==version2[i]) {
```
ChristophWurst:
```js
if(version1[i] !== version2[i]) {
   return false; // version 2 is higher
}
```

Otherwise: do nothing and continue with the next digit.

skjnldsv: Yeah, I snapped this code from somewhere and didn't even realised how poorly it was written...
Please forgive me! 😢

js/components/addressBook/addressBook_controller.js
skjnldsv: Yes we do.
If we're comparing 9.0.0 and 9.0.0.1 we still need to check the last value. But I need to add a safety conversion if NaN or undefined.

skjnldsv:
@ChristophWurst Fixed and rewritten 😊

ChristophWurst:
nice! code looks good now

ChristophWurst:
you could add a unit test for this function 😉

skjnldsv:
I coulc I could... 😂

ChristophWurst:
merge? 🙈 🙊 🙉

skjnldsv:
Since it's a function inside a controller, I don't think I can do tests directly :)
Soooo, let's merge? 😄

ChristophWurst:
> Since it's a function inside a controller, I don't think I can do tests directly :)

You could export but prefix it with an underscore ;-)

skjnldsv:
But is it really necessary? :D 
Let's not over-complicate things!

brjhaverkamp:
Thanks for fixing this. I'm still trying to figure out how this version comparison will solve the export of single address groups. But I am eager to give it a try. Will this be released alongside NC 11? Or how can I try this best?

PR #54, opened 2016-12-07
Add avatar download, view and remove

### LINKED ISSUE #30: Can't view, remove, or download contact photo
User clicks on contact photo

Expected: User can view photo large, download photo, remove photo, or replace photo
With icons - down arrow, pencil, - sign, + sign

Actual: User can only change photo.

There is no obvious way to
a) delete photo (and leave contact with no photo) 
b) view the photo large (often necessary or desirable)
c) download the photo (again, often desirable)

skjnldsv:
Fix #30

capture d ecran_2016-12-07_21-08-33

capture d ecran_2016-12-11_16-21-27

Features:
1. click to download image
2. click to zoom image and click outside of the avatar to quit or press escape
3. delete the avatar

### DIFF
--- a/js/components/avatar/avatar_controller.js
+++ b/js/components/avatar/avatar_controller.js
@@ -39,7 +39,18 @@ angular.module('contactsApp')
 	};
 
 	ctrl.openPhoto = function() {
-
+		$('avatar').toggleClass('maximized');
+		$('avatar').click(function() {
+			$('avatar').removeClass('maximized');
+		});
+		$('avatar img, avatar .avatar-options').click(function(e) {
+			e.stopPropagation();
+		});
+		$(document).keyup(function(e) {
+			if (e.keyCode == 27) {
+				$('avatar').removeClass('maximized');
+			}
+		});
 	};
 
 });
--- a/js/components/avatar/avatar_controller.js
+++ b/js/components/avatar/avatar_controller.js
@@ -47,7 +47,7 @@ angular.module('contactsApp')
 			e.stopPropagation();
 		});
 		$(document).keyup(function(e) {
-			if (e.keyCode == 27) {
+			if (e.keyCode === 27) {
 				$('avatar').removeClass('maximized');
 			}
 		});
--- a/js/components/avatar/avatar_controller.js
+++ b/js/components/avatar/avatar_controller.js
@@ -7,6 +7,7 @@ angular.module('contactsApp')
 	ctrl.removePhoto = function() {
 		ctrl.contact.removeProperty('photo', ctrl.contact.getProperty('photo'));
 		ContactService.update(ctrl.contact);
+		$('avatar').removeClass('maximized');
 	};
 
 	ctrl.downloadPhoto = function() {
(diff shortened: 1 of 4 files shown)


js/components/avatar/avatar_controller.js
```diff
@@ -4,4 +4,53 @@ angular.module('contactsApp')
...
+
+	ctrl.openPhoto = function() {
+		$('avatar').toggleClass('maximized');
+		$('avatar').click(function() {
+			$('avatar').removeClass('maximized');
+		});
+		$('avatar img, avatar .avatar-options').click(function(e) {
+			e.stopPropagation();
+		});
+		$(document).keyup(function(e) {
```
ChristophWurst: what if this method is called more than once? Will there be more than one registered keyup handler on document?

skjnldsv:
@ChristophWurst Thanks, make sense!
Review?

ChristophWurst:
Nice!
Two little issues:
1. the icons are only visible on hover, therefore this could be problematic on phones or tablets where you don't have a mouse
2. if no avatar is set the upload icon is not centered anymore:
bildschirmfoto von 2016-12-11 19-08-21

skjnldsv:
1. Dr. Firefox I presume? 😩
2. Yeah, this is an issue I'm aware of, but need to be fixed in another pr since the upload icon already has this issue.

skjnldsv:
Fixed.

ChristophWurst:
> Dr. Firefox I presume? 😩

Right 😉 

The upload icon is centered now 👍

jancborchardt:
Sorry I didn’t see this earlier, but now there’s a bit much going on here. 4 icons without label :\

Deleting and uploading a different photo makes a lot of sense, but viewing and downloading is waaay less important and takes attention from the really important actions.

skjnldsv:
Is it really that bad?
I think we can let this like that until we figure out a solution for a better system than just a hover for icons (for mobile users for example) :)

jancborchardt:
Yes, I think it went from a really simple, »just upload« function to a mass of actions where you don’t really know what is for what and the most important options are overshadowed. This is not how we do design in Nextcloud. ;)

If we find something is missing, we gradually add things. The single most important action is uploading the image, everything else could go into a 3-dot menu next to it, or rather be integrated. Because:
- Viewing an image in large could just be done by clicking on it, no extra action necessary
- Downloading the image could be done from that detail view
- Deleting an image is very rare, usually you just update it. But if you really need to, deleting could also be done from that detail view

We can do better than cramming the actions all inside that small circle. So what do you think about that – having the upload action show on the bottom part of the circle (on hover), and the rest of the actions, including also upload, only when viewing the image on click of the image itself?

jancborchardt:
And in the view of the image, the actions should of course have a text label.

skjnldsv:
We should indeed find something else!
I fully agree with the menu, but I need to think of a way to correctly implement it.
A three dot under the avatar?

capture d ecran_2016-12-14_22-11-25

jancborchardt:
As said I think it’s better to have the primary action displayed directly. And the primary is upload:
capture du 2016-12-14 22-37-06
The rest can be displayed when either clicking on the image directly, as is now (although centered and with proper labels).

jancborchardt:
Oh and that should of course only show on hover/focus. It’s fine for mobile also since when you click/tap the image, the detail view will show Upload and the other actions too.

PR #59, opened 2016-12-14
Contacts app is actually in the "Organization" category

LEDfan:
Thanks to Tweakers.net user for reporting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 ## Installation
 
-In your Nextcloud, simply navigate to »Apps«, choose the category »Productivity«, find the Contacts app and enable it.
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Actually, it's in three categories: https://github.com/nextcloud/contacts/blob/4c3537983a6dfeb3fcae7546f66c2c56fab6a985/appinfo/info.xml#L17-L19

LEDfan:
I know. For the readme would it be better for to choose one, or list them all?

ChristophWurst:
Yes, that makes sense. Listing all three would be confusing. Nvm my comment then :-)

irgendwie:
Thanks!

Issue #41, opened 2016-11-27
Use common timestamp format for REV


rfc2822:
Hello,

In 44b8edd4f2223bf96cfd9ef891f40a37fb489ed9, setting `REV` from JavaScript was introduced like this:

```
contact.rev(new Date().toISOString());
````

The format generated by `new Date().toISOString()` is `2016-11-27T16:53:30.027Z`, which currently causes problems with VCard libraries like ez-vcard (https://github.com/mangstadt/ez-vcard/issues/73). While the format seems to be correct according to ISO.8601.2004, it would reduce problems with CardDAV clients like DAVdroid if the timestamp could be printed in an easier format, like `REV:19951031T222710Z` (without milliseconds).

Also, `VALUE=timestamp` is redundant and could be removed.

Would that be possible? Currently, this change (released with nextcloud/contacts 1.5.2) [breaks DAVdroid synchronization (more information here)](https://forums.bitfire.at/topic/1259/rev-must-not-appear-more-than-once-in-a-vcard-component/).

**Comment by irgendwie** [comment]
Will look into this, thank you for reporting!

**Comment by dervomsee** [comment]
Hi,
i'm sorry, same error message again:
```
SYNCHRONIZATION INFO
Synchronization phase: 4
Account name: post@***.de
Authority: com.android.contacts

HTTP REQUEST:
PUT /remote.php/dav/addressbooks/users/conny/kontakte1/2e4f2d8d8a.vcf
If-Match: "d37068c3c4463c735d23d79133b899b7"

BEGIN:VCARD[CR][LF]
VERSION:4.0[CR][LF]
PRODID:+//IDN bitfire.at//DAVdroid/1.3.4.1-ose vcard4android ez-vcard/0.10.[CR][LF]
 0[CR][LF]
REV;VALUE=timestamp:2016-11-19T17:44:22.057Z[CR][LF]
UID:2e4f2d8d8a[CR][LF]
FN:Name Name[CR][LF]
N:Name;Name;;;[CR][LF]
TEL;TYPE=voice:+49123456[CR][LF]
TEL;TYPE=fax:+49123456[CR][LF]
TEL;TYPE=voice:+49123456[CR][LF]
REV:20161130T181514Z[CR][LF]
END:VCARD[CR][LF]


HTTP RESPONSE:
http/1.1 415 Unsupported Media Type
Cache-Control: no-store, no-cache, must-revalidate
Connection: keep-alive
Content-Security-Policy: default-src 'none';
[...]

PR #65, opened 2016-12-21
Mark REV values as UTC (see #41)

rfc2822:


skjnldsv:
We need to be sure we're following rfc 3339
```
Z           A suffix which, when applied to a time, denotes a UTC
            offset of 00:00; often spoken "Zulu" from the ICAO
            phonetic alphabet representation of the letter "Z".
```
https://www.ietf.org/rfc/rfc3339.txt

zeehio:
I am all for the "Z".

Dropping the milliseconds does not go against the rfc3339. In fact:

```
5.3. Rarely Used Options

   A format which includes rarely used options is likely to cause
   interoperability problems.  This is because rarely used options are
   less likely to be used in alpha or beta testing, so bugs in parsing
   are less likely to be discovered.  Rarely used options should be made
   mandatory or omitted for the sake of interoperability whenever
   possible.

   The format defined below includes only one rarely used option:
   fractions of a second.  It is expected that this will be used only by
   applications which require strict ordering of date/time stamps or
   which have an unusual precision requirement.
```

So either there is a rationale for keeping milliseconds or we should avoid using it.

PR #74, opened 2016-12-29
Fix avatar options visibility and layout

skjnldsv:
- Don't show any option except the avatar if it is set
- Show only the upload button if no avatar is set
- Clicking the image open the avatar options

capture d ecran_2016-12-29_10-40-55

Following #54 
@jancborchardt


skjnldsv:
capture d ecran_2016-12-29_18-24-39

capture d ecran_2016-12-29_18-25-45

skjnldsv:
🎉 💟

PR #85, opened 2017-01-05
Fix contacts group split and added contact failure notification

skjnldsv:
- Remove the ability to add comma into groups name for better readability
- Automatically split unescaped comma into groups to repair broken vcard groups
- Added notification for failed vcard
- Added a save button in case of corrupted vcard
- Added function for future vcard correction with automatic error notification

Broken vcard for review: 
[6cc082f2-17a7-4730-99e1-c9870f072d9d.txt](https://github.com/nextcloud/contacts/files/723757/6cc082f2-17a7-4730-99e1-c9870f072d9d.txt)

(should have three groups, broken master show 2 with one not separated by the comma)
capture d ecran_2017-02-02_07-40-41

----------------------

capture d ecran_2017-01-05_12-33-48
capture d ecran_2017-01-05_12-34-06
capture d ecran_2017-01-05_12-34-19

### DIFF
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -157,9 +157,14 @@ angular.module('contactsApp')
 						return [];
 					}
 					if (angular.isArray(property.value)) {
+						// Avoid and fix unescaped commas
+						if(property.value.join(';').indexOf(',') !== -1) {
+							property.value = property.value.join(',').split(',');
+							this.setProperty('categories', { value: property.value });
+						}
 						return property.value;
 					}
-					return [property.value];
+					return property.value.split(',');
 				}
 			},
 
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -320,9 +320,6 @@ angular.module('contactsApp')
 						}
 					}
 					break;
-
-				default:
-					break;
 				}
 				return property;
 			}
@@ -342,7 +339,7 @@ angular.module('contactsApp')
 
 		var property = this.getProperty('categories');
 		if(!property) {
-			this.categories('');
+			this.categories([]);
 		} else {
 			if (angular.isString(property.value)) {
 				this.categories([property.value]);
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -5,6 +5,7 @@ angular.module('contactsApp')
 
 			data: {},
 			props: {},
+			failedProps: [],
 
 			dateProperties: ['bday', 'anniversary', 'deathdate'],
 
@@ -152,19 +153,14 @@ angular.module('contactsApp')
 					return this.setProperty('categories', { value: value });
 				} else {
 					// getter
-					var property = this.getProperty('categories');
+					var property = this.validate('categories', this.getProperty('categories'));
 					if(!property) {
 						return [];
 					}
 					if (angular.isArray(property.value)) {
-						// Avoid and fix unescaped commas
-						if(property.value.join(';').indexOf(',') !== -1) {
-							property.value = property.value.join(',').split(',');
-							this.setProperty('categories', { value: property.value });
-						}
 						return property.value;
 					}
-					return property.value.split(',');
+					return [property.value];
 				}
 			},
 
@@ -269,6 +265,16 @@ angular.module('contactsApp')
 
 				// keep vCard in sync
 				self.data.addressData = $filter('JSON2vCard')(self.props);
+
+				// Revalidate all props
+				_.each(self.failedProps, function(name, index) {
+					if (!_.isUndefined(self.props[name]) && !_.isUndefined(self.props[name][0])) {
+						// Set dates again to make sure they are in RFC-6350 format
+						self.failedProps.splice(index, 1);
+						self.validate(name, self.props[name][0]);
+					}
+				});
+
 			},
 
 			matches: function(pattern) {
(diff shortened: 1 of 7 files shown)


js/models/contact_model.js
```diff
@@ -281,6 +292,29 @@ angular.module('contactsApp')
 					return false;
 				});
 				return matchingProps.length > 0;
+			},
+
+			validate: function(prop, property) {
```
ChristophWurst: we really should start unit testing stuff like this 😉

skjnldsv: You're right!

js/tests/models/contact_model.js
```diff
@@ -27,4 +27,11 @@ describe('contactModel', function() {
 		var d = contact.getISODate(new Date('2016-09-01T09:07:05Z'));
 		expect(d).to.equal('20160901T090705Z');
 	});
+
+	it('should fix invalid group array', function() {
+		var contact = new $Contact({displayName: 'test'});
+		contact.categories(['Test 1', 'Test 2\,Test 3']);
+		var categories = contact.categories();
+		expect(categories).to.equal(['Test 1', 'Test 2', 'Test 3']);
```
skjnldsv: @ChristophWurst This doesn't work. Any idea why?
`expected [ 'Test 1', 'Test 2', 'Test 3' ] to equal [ 'Test 1', 'Test 2', 'Test 3' ]`

skjnldsv:
Testing fixed.
Please review again.

js/tests/models/contact_model.js
skjnldsv: I needed to use deep.equal

js/models/contact_model.js
```diff
@@ -291,6 +302,29 @@ angular.module('contactsApp')
...
+				switch(prop) {
```
ChristophWurst: since there's only one condition here, it would be easier to simply change this to an if statement

js/models/contact_model.js
```diff
@@ -291,6 +302,29 @@ angular.module('contactsApp')
...
+						}
+					} else if (angular.isString(property.value)) {
+						if(property.value.indexOf(',') !== -1) {
+							this.failedProps.push(prop);
+							property.value = property.value.split(',');
+						}
+					}
+					break;
+
+				default:
```
ChristophWurst: this branch is not needed

js/models/contact_model.js
skjnldsv: I'm planning to put more testing in there :)

js/models/contact_model.js
skjnldsv: Indeed! Reliques from an old test :p

jancborchardt:
There’s starting to be too many icons without text in the top right. Can we change that to only show one icon (the Save icon) with text right next to it, and then the 3-dot-icon as more menu, with Download and Delete in it?

skjnldsv:
The save icon is only shown when there is an issue with the card. Otherwise, this is an autosave :)

skjnldsv:
@jancborchardt bump!

ChristophWurst:
> Broken vcard for review:
> 6cc082f2-17a7-4730-99e1-c9870f072d9d.txt

I downloaded it, renamed to .vcf and tried to import. The ajax request resulted in a HTTP 500 and this reponse:
```xml
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Doctrine\DBAL\Exception\DriverException</s:exception>
  <s:message>An exception occurred while executing 'INSERT INTO `oc_cards` (`carddata`, `uri`, `lastmodified`, `addressbookid`, `size`, `etag`) VALUES(?, ?, ?, ?, ?, ?)' with params ["BEGIN:VCARD xxxxxxxxxxxxx END:VCARD", "dc0b2664-9da4-4059-b531-1e49fc786c3a.vcf", 1485158019, "1", 1322594, "534fac2deb9cb3642a73870bd732e901"]:

SQLSTATE[08S01]: Communication link failure: 1153 Got a packet bigger than 'max_allowed_packet' bytes</s:message>
</d:error>
```

skjnldsv:
@ChristophWurst This comes from your sql settings 😜 

Here's a simple vcard, I had left the picture in the previous one 😞 
[6cc082f2-17a7-4730-99e1-c9870f072d9d.txt](https://github.com/nextcloud/contacts/files/723757/6cc082f2-17a7-4730-99e1-c9870f072d9d.txt)

ChristophWurst:
Fixes the contact as expected, but the warning is not shown on FF

skjnldsv:
@ChristophWurst I have ti on ff ! 😕 
kazam_screenshot_00000

irgendwie:
> Remove the ability to add comma into groups name for better readability

works, until you start to add already existing groups:
=> adding "test" and then "test,test2" leads to visually broken vcards and odd bugs, try it: [example vcard](https://github.com/nextcloud/contacts/files/749001/test.txt)
@skjnldsv different issue for that?

> Automatically split unescaped comma into groups to repair broken vcard groups

works
> Added notification for failed vcard

works
> Added a save button in case of corrupted vcard

works, but nearly didn't notice it was there - maybe we should think about adding it next to the repaired field
> Added function for future vcard correction with automatic error notification

nice! :+1:

skjnldsv:
> adding "test" and then "test,test2" leads to visually broken vcards and odd bugs, try it: example vcard

I don't really get it, do you have a group with a coma in the left sidebar?

> works, but nearly didn't notice it was there - maybe we should think about adding it next to the repaired field

Could be a good idea!

irgendwie:
@skjnldsv what happens, if you import the example vcard I provided?

skjnldsv:
@irgendwie #112

jancborchardt:
By the way @skjnldsv just wanted to give you a _huge_ thanks for fixing this issue! :)

skjnldsv:
❤️

Issue #82, opened 2017-01-03
Translation issue for 'Anniversary' in french


kao98:
### Steps to reproduce
1. Use the contact app in french
2. Create a new contact
3. Add a new field: birtday ("Anniversaire" in french)

### Expected behaviour
To have a single field "Anniversaire" corresponding to the birthday of the contact

### Actual behaviour
There are two fields "Anniversaire", one for the birthday, one for another anniversary. But in french we only have one word for those two terms, "Anniversaire". It should correspond only to "birthday". The "Anniversary" field should be translated something else (see some proposition below).

### Server configuration

**Contacts version:** 1.5.2

**Fresh install**

### Translation proposition
Android translated the "Anniversary" field like this: "Fête". That correspond to the name day.
But actually, I prefer the more generic translation of Microsoft in the windows 10 contacts app: "Autre date" that correspond to "Another date".

Are there some french guys that could give their opinion in there?
Otherwise if everyone agree I'll submit a pull request with "Autre date" instead of "Anniversaire" for the "Anniversary" field.

skjnldsv:
Translations are done over transifex: https://www.transifex.com/nextcloud/nextcloud/
But as far as I'm concern, "Autre Date" seems more appropriate to me.

skjnldsv:
@kao98 Any news?

kao98:
Hi,

I registered on transifex.com and ask to join the nextcloud organization.
It is still pending yet.

2017-01-10 15:02 GMT+01:00 John Molakvoæ <notifications@github.com>:

skjnldsv:
Fixed

PR #86, opened 2017-01-05
Fix issue #82

kao98:
Anniversary now translated from "Anniversaire" to "Autre date" in french.
More info in issue #82


skjnldsv:
Like I said,
> Translations are done over transifex: https://www.transifex.com/nextcloud/nextcloud/

You can't edit it directly from here.

Issue #12, opened 2016-10-24
Introduce lazy loading to shorten waiting time on open


eppfel:
I open this issue, because I think this would be a huge improvement.
As discussed here: owncloud/contacts#176
### Steps to reproduce
1. Create an adressbook with more than 300 contacts
2. Open the contacts app in your browser
3. Stare at mesmerizing loading animation, until you fall into a state of deep trance

daniellandau:
The way I'm reading the spec (Example in https://tools.ietf.org/html/rfc6352#section-8.6.5, definition in https://tools.ietf.org/html/rfc6352#section-10.6) it seems that there isn't proper paging support in CardDav. You can ask for less results, but not results starting from the next page (that is, "give me results 1-10, then 11-20, etc").

I think the real thing we want is not lazy loading, but good performance and good UX in general. To that end we need to profile the app to see whether it's for example

1. Taking a long time to parse the vcard data → we can ask for minimal fields to render listing and make a backend request for more when user clicks on a contact
2. Taking a long time to render the list → we can render only the beginning of the list and render more when the user scrolls to the end or is idle.

If that fails to get results, we could (3) make the user click a button to fetch the full list if they want to or alternatively use the search to find contacts. (Talking about search, it took me reading the source code to realize that the search button is for contact search. I had always assumed it was a global search. Maybe it could be expanded and have a place holder text "Search contacts").

eppfel:
That sounds reasonable to me. :+1:

silverhook:
As someone with 1k+ contacts in the address book, running on an ARM server, I would **very** much like to see this happen :smiley:

daniellandau:
So, results.

### Problem

First of all, here's a full profile of initial load with ~1100 contacts, about ~25 of them having profile pictures.

![kuvakaappaus - 2017-01-07 23-00-21](https://cloud.githubusercontent.com/assets/1044056/21745139/1f386c1a-d52d-11e6-859b-ab9ac5bd4e5c.png)

The startup takes in total roughly 30 seconds with my i7 machine and the Chromium dev tools set to throttle transfer bandwidth at 10Mbits per second (the same I have at home). Most of the time (~21 seconds) is spent waiting for the whole addressbook with everyone's pictures downloading.

After that significant time is spent on other various things.

* Parsing the response and creating `Contact` objects, about 1.3 seconds

[...]

PR #88, opened 2017-01-07
Introduce lazy loading

daniellandau:
This is related to https://github.com/nextcloud/contacts/issues/12

### DIFF
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -21,7 +21,7 @@ angular.module('contactsApp')
 		clearInterval(ctrl.intervalId);
 		ctrl.intervalId = setInterval(
 			function () {
-				if (!ctrl.loading && ctrl.contacts.length > ctrl.limitTo) {
+				if (!ctrl.loading && ctrl.contacts && ctrl.contacts.length > ctrl.limitTo) {
 					ctrl.limitTo += 25;
 					$scope.$apply();
 				}
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -1,5 +1,5 @@
 angular.module('contactsApp')
-.controller('contactlistCtrl', function($scope, $filter, $route, $routeParams, ContactService, SortByService, vCardPropertiesService, SearchService) {
+.controller('contactlistCtrl', function($scope, $filter, $route, $routeParams, $timeout, ContactService, SortByService, vCardPropertiesService, SearchService) {
 	var ctrl = this;
 
 	ctrl.routeParams = $routeParams;
@@ -56,7 +56,7 @@ angular.module('contactsApp')
 	ctrl.loading = true;
 
 	ContactService.registerObserverCallback(function(ev) {
-		$scope.$apply(function() {
+		$timeout(function () { $scope.$apply(function() {
 			if (ev.event === 'delete') {
 				if (ctrl.contactList.length === 1) {
 					$route.updateParams({
@@ -82,7 +82,7 @@ angular.module('contactsApp')
 				});
 			}
 			ctrl.contacts = ev.contacts;
-		});
+		}); });
 	});
 
 	// Get contacts
@@ -96,7 +96,45 @@ angular.module('contactsApp')
 		}
 	});
 
-	// Wait for ctrl.contactList to be updated, load the first contact and kill the watch
+	var getVisibleNames = function getVisibleNames() {
+		function isScrolledIntoView(el) {
+			var elemTop = el.getBoundingClientRect().top;
+			var elemBottom = el.getBoundingClientRect().bottom;
+
+			var bothAboveViewport = elemBottom < 0;
+			var bothBelowViewPort = elemTop > window.innerHeight;
+			var isVisible = !bothAboveViewport && !bothBelowViewPort;
+			return isVisible;
+		}
+
+		var elements = Array.prototype.slice.call(document.querySelectorAll('.contact__icon:not(.ng-hide)'));
+		var names = elements
+				.filter(isScrolledIntoView)
+				.map(function (el) {
+					var siblings = Array.prototype.slice.call(el.parentElement.children);
+					var nameElement = siblings.find(function (sibling) {
+						return sibling.getAttribute('class').indexOf('content-list-item-line-one') !== -1;
+					});
+					return nameElement.innerText;
+
+				});
+		return names;
+	};
+
+	var timeoutId = null;
+	document.querySelector('.app-content-list').addEventListener('scroll', function () {
+		clearTimeout(timeoutId);
+		timeoutId = setTimeout(function () {
+			var names = getVisibleNames();
+			ContactService.getFullContacts(names);
+		}, 250);
+	});
+
+	// Wait for ctrl.contactList to be updated, load the contact requested in the URL if any, and
+	// load full details for the probably initially visible contacts.
+	// Then kill the watch.
+	// Don't blindly load the first contact as that is a heavy operation that hangs the browser, preventing
+	// for example searching while the DOM is updated
 	var unbindListWatch = $scope.$watch('ctrl.contactList', function() {
 		if(ctrl.contactList && ctrl.contactList.length > 0) {
 			// Check if a specific uid is requested
@@ -108,6 +146,8 @@ angular.module('contactsApp')
 					}
 				});
 			}
+			var firstNames = ctrl.contactList.slice(0, 20).map(function (c) { return c.displayName(); });
+			ContactService.getFullContacts(firstNames);
 			ctrl.loading = false;
 			unbindListWatch();
 		}
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -133,7 +133,6 @@ angular.module('contactsApp')
 	// Wait for ctrl.contactList to be updated, load the contact requested in the URL if any, and
 	// load full details for the probably initially visible contacts.
 	// Then kill the watch.
-	// Don't blindly load the first contact as that is a heavy operation
 	var unbindListWatch = $scope.$watch('ctrl.contactList', function() {
 		if(ctrl.contactList && ctrl.contactList.length > 0) {
 			// Check if a specific uid is requested
@@ -144,6 +143,10 @@ angular.module('contactsApp')
 						ctrl.loading = false;
 					}
 				});
+			}
+			// No contact previously loaded, let's load the first of the list if not in mobile mode
+			if(ctrl.loading && $(window).width() > 768) {
+				ctrl.setSelectedId(ctrl.contactList[0].uid());
 			}
 			var firstNames = ctrl.contactList.slice(0, 20).map(function (c) { return c.displayName(); });
 			ContactService.getFullContacts(firstNames);
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -133,8 +133,7 @@ angular.module('contactsApp')
 	// Wait for ctrl.contactList to be updated, load the contact requested in the URL if any, and
 	// load full details for the probably initially visible contacts.
 	// Then kill the watch.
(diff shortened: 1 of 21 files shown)


eppfel:
Thanks, for that detailed work. I would move the detailed explanation here, because you not only the explain the problem, but also your solution.

I ping @nextcloud/contacts @ChristophWurst for review!

templates/group.html
```diff
@@ -1 +1 @@
-<a ng-href="#/{{ctrl.group}}">{{ctrl.group}}</a>
+<a ng-href="#/{{ctrl.group}}">{{ctrl.group}} ({{ctrl.groupCount}})</a>
```
skjnldsv: Please format menus according to guidelines: https://docs.nextcloud.com/server/11/developer_manual/app/css.html#menus

skjnldsv: Please format menus according to guidelines: https://docs.nextcloud.com/server/11/developer_manual/app/css.html#menus

daniellandau: Done.

js/filters/counterFormatter_filter.js
```diff
@@ -0,0 +1,12 @@
+// from https://docs.nextcloud.com/server/11/developer_manual/app/css.html#menus
+angular.module('contactsApp')
```
skjnldsv: Please add tests (seehttps://github.com/owncloud/contacts/blob/31e403af1db859f85e84c2337134e20e5719c825/js/tests/filters/counterFormatter_filter.js)

templates/group.html
```diff
@@ -1 +1 @@
-<a ng-href="#/{{ctrl.group}}">{{ctrl.group}} ({{ctrl.groupCount}})</a>
+<a ng-href="#/{{ctrl.group}}">{{ ctrl.group }} ({{ ctrl.groupCount | counterFormatter }})</a>
```
skjnldsv: Better, but you need to use the li element
`<li class="app-navigation-entry-utils-counter">{{ count | counterFormatter }}</li>`

capture d ecran_2017-01-08_17-43-40

templates/groupList.html
```diff
@@ -1 +1 @@
-<li ng-repeat="group in ctrl.groups" group data="group" ng-click="ctrl.setSelected(group)" ng-class="{active: group === ctrl.getSelected}"></li>
+<li ng-repeat="group in ctrl.groups" group group-name="group[0]" group-count="group[1]" ng-click="ctrl.setSelected(group[0])" ng-class="{active: group[0] === ctrl.getSelected}"></li>
```
skjnldsv: you could do something like `<li ng-repeat="(group, count) in ctrl.groups | orderGroupList" group data="group" count="count" ng-click="ctrl.setSelected(group)" ng-class="{active: group === ctrl.getSelected}"></li>` for better readibility :)

skjnldsv: you could do something like `<li ng-repeat="(group, count) in ctrl.groups" group data="group" count="count" ng-click="ctrl.setSelected(group)" ng-class="{active: group === ctrl.getSelected}"></li>` for better readibility :)

See my previous work: https://github.com/owncloud/contacts/pull/341/files

skjnldsv: you could do something like `<li ng-repeat="(group, count) in ctrl.groups" group data="group" count="count" ng-click="ctrl.setSelected(group)" ng-class="{active: group === ctrl.getSelected}"></li>` for better readibility :)

See my previous work: https://github.com/owncloud/contacts/pull/341/files

js/components/contactList/contactList_controller.js
```diff
@@ -92,10 +139,8 @@ angular.module('contactsApp')
 					}
 				});
 			}
-			// No contact previously loaded, let's load the first of the list if not in mobile mode
-			if(ctrl.loading && $(window).width() > 768) {
-				ctrl.setSelectedId(ctrl.contactList[0].uid());
-			}
+			var firstNames = ctrl.contactList.slice(0, 20).map(function (c) { return c.displayName(); });
+			ContactService.getFullContacts(firstNames);
```
skjnldsv: We still don't want to load the first contact in mobile mode. 
We need a condition to prevent that like we had before ;)

templates/groupList.html
daniellandau: Oh, did not know that Angular does list destructuring, will fix.

skjnldsv:
If no contacts, I get errors on the console:
```
Uncaught TypeError: Cannot read property 'length' of undefined
    at contactList_controller.js:22
```

templates/groupList.html
daniellandau: Hmm, it seems that that syntax is only used for going through the keys and values in an object.

js/components/contactList/contactList_controller.js
daniellandau: We now never blindly load the first contact, so not on mobile either.

skjnldsv: By selecting the first uid after load, we display the first contact.
In mobile, since the list view and detail views are split, we need to stay on the contact list only. :)

daniellandau:
I added a defense in contactList_controller.

js/filters/counterFormatter_filter.js
daniellandau: Done.

templates/group.html
daniellandau: In order to get right-aligning instead of parentheses I gather. Sure thing, fixed.

js/components/contactList/contactList_controller.js
daniellandau: I don't think we are selecting the first uid at all. I tested with both a responsive design mode and an actual mobile browser.

skjnldsv: So the right app panel is blank on your contact installation? On fresh click of `your.domain/index.php/apps/contacts`, the app doesn't load anything?

daniellandau: Exactly.

js/services/contact_service.js
```diff
@@ -35,8 +62,10 @@ angular.module('contactsApp')
 								if (addressBook.objects[i].addressData) {
 									var contact = new Contact(addressBook, addressBook.objects[i]);
 									contacts.put(contact.uid(), contact);
+									var oldList = urlsByDisplayname.get(contact.displayName()) || [];
+									urlsByDisplayname.put(contact.displayName(), oldList.concat(contact.data.url));
 								} else {
-									// custom console
+									// eslint-disable-next-line no-console
```
skjnldsv: Nice! 😆

js/components/contactList/contactList_controller.js
skjnldsv: Strange, it changes the url with an uid and then a contact.
Anyway, we need to keep the old mechanic in place.
1. If mobile, do not load contact details and saty on the contacts list EXCEPT if there is a matching uid specified in the url
2. If not in mobile, load the first contact

daniellandau: Ok, added that back in. It adds ~400 ms rendering time, but in the scheme of things that's probably fine.

skjnldsv: You're awesome! :)

daniellandau:
@skjnldsv did you want me to still do things? Like rebase with signed off commits?

templates/groupList.html
skjnldsv: Isn't there an alternative?

daniellandau: The order of keys in an object is arbitrary, so I don't want to have the whole `ctrl.groups` be an object. We want to guarantee `All Contacts` is first.

skjnldsv:
Not really. But we need to wait for the 3rdparty patch/sabre release.

templates/groupList.html
skjnldsv: Too bad, it would have been great to improve the readability. But it's clearly not that important :)

jancborchardt:
Any idea when this sabre patch release is coming? cc @evert

daniellandau:
I don't know. I asked on their IRC channel a couple of weeks ago and got no reply. I asked two days ago on Github (https://github.com/fruux/sabre-dav/issues/936), but haven't gotten any reply yet.

Henni:
I'd suggest to split this up into two separate pull requests:
* one for the the partial carddav requests, which is blocked by sabre-dav
* and another one which loads the contacts list piece by piece

I'd like to merge the second one as soon as possible, because it might introduce a nice performance improvement and also to reduce merge conflicts in the future.

@daniellandau what do you think about this? would you do the job of splitting this PR up?

daniellandau:
Do you mean as the second part the thing that renders the contact list piece by piece? I'd rather just merge the sabre-dav patch and merge this whole thing. All in all that would be much less maintenance work in my opinion, and the sabre-dav patch is merged in all branches upstream, so it shouldn't cause conflicts or problems later on either.

Henni:
@daniellandau *shouldn't* cause any conflicts. But I also understand that the guys over at nextcloud/server don't want to patch anything unreleased into their 3rdparty repo.

Let's wait a bit longer for a response by sabre/dav.

daniellandau:
Just a quick update, the sabre/dav release has come, and the current PR on 3rdparty is https://github.com/nextcloud/3rdparty/pull/40

daniellandau:
I rebased the PR to current master, ping @skjnldsv.

irgendwie:
If I'm not mistaken, the REPORT functionality (https://github.com/fruux/sabre-dav/issues/889) is only included starting from sabre-dav 3.0.10 (https://github.com/fruux/sabre-dav/blob/master/CHANGELOG.md#3010-2016--) which isn't released (by now?) and only sabre-dav 3.2.1 (https://github.com/fruux/sabre-dav/blob/master/CHANGELOG.md#321-2017-01-28) seems to support the REPORT functionality .

sabre-dav in Nextcloud 9 was only bumped to sabre-dav 3.0.9 (https://github.com/nextcloud/3rdparty/tree/stable9/sabre), which means we wouldn't support Nextcloud 9 anymore by now.

And sabre-dav in Nextcloud 10 was also only bumped to sabre-dav 3.0.9 (https://github.com/nextcloud/3rdparty/tree/stable10/sabre), which also means no support for Nextcloud 10 anymore by now.

And sabre-dav in Nextcloud 11 was also only bumped to sabre-dav 3.2.0 (https://github.com/nextcloud/3rdparty/tree/stable11/sabre), which also means no support for Nextcloud 11 anymore by now.

:-1: I'm not willing to merge this PR into the master branch, which would basically mean that the next version of contacts would only be compatible with Nextcloud 12.
We need a solution for this, before we merge this.

---> After some testing, if the REPORT feature is not supported, it still seems to work (requested properties are ignored?)

daniellandau:
> ---> After some testing, if the REPORT feature is not supported, it still seems to work (requested properties are ignored?)

Yes, the properties are simply ignored. There's some extra network traffic when the browsers asks for the visible cards again (as they are already fully downloaded in the beginning).

jancborchardt:
@LukasReschke @MorrisJobke care to check this out? :) Would be great to get your opinion on this.

MorrisJobke:
> @LukasReschke @MorrisJobke care to check this out? :) Would be great to get your opinion on this.

Sadly I don't have time for this :/

jancborchardt:
Anyone else who can help here? @rullzer @schiessle @nickvergessen @georgehrke

skjnldsv:
🎉

jancborchardt:
Great work everyone! :)

timreeves:
Does "renders the contact list piece by piece" translate to "infinite scrolling", 'cos that would be really great. I have 4 adress books shared with different people, totalling around 1300 contacts, and the load time is lousy, even on a fast linux PC with fast cable internet. It would be great, really great, if I could select which of the adress books should be displayed (BEFORE they are) - I would have thought that an elementary feature.

And a client of mine uploaded about 15.000 contacts to one adress book; since then it has been unusable and would have to be deleted manually from the DB (igitt).

So yes, infinite scrolling / lazy loading or whatever we call it would imho be **neccessary** for professional use. Or even/also the possibility to limit the display to first name / last name beginning with...

Many thanks to all working on this!

PR #92, opened 2017-01-15
Fix url-link icon position.

MG2R:
Previously, the url-link icon would overlap the delete icon (.icon-delete). This change moves the url-link icon to the right.

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -261,7 +261,6 @@ detailsitem .icon-delete {
 detailsitem .url-link {
 	position: absolute;
 	padding: 8px 10px;
-	margin-left: -30px;
 	opacity: .5;
 }
 


skjnldsv:
Hum, it's still broken for me :/
Does ths works for you?
``` css
detailsitem .url-link {
    position: absolute;
    padding: 8px 10px;
    opacity: .5;
    top: 2px;
    right: 33px;
}
```

MG2R:
That works too, just puts it inside the text field, which means that a long link will obscure the icon again.

On my end, my original change puts the 'go to URL' icon to the right of the 'delete field' button, which seems to make more sense to me.

skjnldsv:
It was meant to be inside at first.
Not as a secondary button :)

PR #99, opened 2017-01-23
Fix url icon inline, closes #92

skjnldsv:
capture d ecran_2017-01-23_05-44-08


css/public/style.css
```diff
@@ -263,6 +263,8 @@ detailsitem .url-link {
 	padding: 8px 10px;
 	margin-left: -30px;
 	opacity: .5;
+    top: 2px;
```
ChristophWurst: indentation looks odd :-)

Issue #14, opened 2016-10-24
Can't scroll through shared addressbook list


SIRSteiner:
If I get many shared contacts (addressbooks) then in settings it is not possible to scroll them.
Please look at the screenshot
![screenshot](https://cloud.githubusercontent.com/assets/3488638/19664088/3b005120-9a3f-11e6-98a8-11cb0ecf9a7d.png)

irgendwie:
@jancborchardt Already told you about that at the meetup! :blush:  Would be cool to have a good looking solution for this. Also @skjnldsv

jancborchardt:
Right! 3 things we need to do:
- [ ] restrict the height of the settings area (should be done in core/css/apps.css), for example to 500px or max 80% height.
- [ ] put the actions in a 3-dot menu because four icons next to each other are difficult to understand
- [ ] ellipsize the names if they are too long so it doesn’t break

cc @nextcloud/designers

jakobsack:
tasks 1 and 3 are ready to be merged. Great work, @pgys !

skjnldsv:
for task 2, I can suggest this again:
![gif](https://cloud.githubusercontent.com/assets/14975046/18564110/787871dc-7b8a-11e6-9a62-90c06acccbaa.gif)

from https://github.com/owncloud/contacts/pull/518#issuecomment-247414277

jancborchardt:
@skjnldsv let’s do it like in the other apps with the popover icons+text. Without the text, 4 icons next to each other are very difficult to understand.

skjnldsv:
Okay, but without texts. Otherwise, we won't have enough space to display the popup and I don't really want to add a big bottom padding 😉

jancborchardt:
What do you mean? The icons need the text for sure, like in all the other menus. :) If there’s not enough space to display the popup menu downwards then we have to make room downwards. We can add the padding depending on if you click the menu so it’s not always there. :)

skjnldsv:
I completely change my view now. I fully agree with you.
I was waiting for my work on the css guidelines. :)
A pr is on its way.

jancborchardt:
@skjnldsv :heart:  :) high five

skjnldsv:
@jancborchardt look at other opened pr on contacts pls :)

skjnldsv:
Anyhow, the popover menu is a bit broken on stable11. Should I backport my apps.css edits?

jancborchardt:
[...]

PR #101, opened 2017-01-24
Addressbooklist in a popover

skjnldsv:
Testing: 
[contacts.tar.gz](https://github.com/nextcloud/contacts/files/749683/contacts.tar.gz)

- [x] Popover structure
    - [x] Css guidelines
    - [x] Don't close settings after click
- [x] Clipboard
    - [x] Tooltips
    - [x] Fallback to input if no clipboard support
- [x] Shares design 
    - [x] List
    - [x] Checkboxes
    - [x] Don't close settings after click

Fix and finishes #14
capture d ecran_2017-02-18_18-14-11 capture d ecran_2017-02-18_18-15-23 capture d ecran_2017-02-18_18-15-47


jancborchardt:
Very nice work! Just 3 details which are probably related to the core styles:
- top right border radius is not set
- there is a bit too much bottom padding on the popover
- »Share« and »Delete« are much lighter than the other item. Were you hovering »Download«? Cause they should be the same lightness.

And one thing: Because sharing is inside the menu, »Copy link« should be too. Because otherwise the icon without text will cause confusion because it’s used in the context of sharing usually. Yes, for sharing a link, but mostly in combination with the other sharing. :)

skjnldsv:
@jancborchardt Thanks! :)
- I wasn't hovering it, there is some fix to do. 
- Top right indeed comes form core, will fix.
- Makes sense for the copy link/share. Should we put the download button outside then?

skjnldsv:
This is what it does on stable 11 @jancborchardt 
capture d ecran_2017-01-25_15-55-44

skjnldsv:
Backport on nextcloud/server#3286

MorrisJobke:
For me those icons don't appear in a popover but inline in the row. 😕 So the three dots menu is not there.

skjnldsv:
@MorrisJobke I added a contact build for testing. :)
But we need this backport! https://github.com/nextcloud/server/pull/3286

skjnldsv:
Rebased.

We now should wait for 11.0.2

skjnldsv:
All clear, rebased and fixed! (see updated screenshots on first post)
Please review @nextcloud/contacts :)

jancborchardt:
@skjnldsv can you post a screenshot of how it looks like now? Am at a fair the week and can't test well :/

How it should be: Sharing icon outside of the popover, but the Copy link function inside. And it should ideally trigger a copy directly on clicking the entry.

skjnldsv:
@jancborchardt The first post screenshots are up-to-date :)

You want me to move the share url into the popover? Also we could completely get rid of the text input for the link and simply copy the link by clicking the entry IN the popovermenu :)

jancborchardt:
@skjnldsv yes, exactly what you said :) move the "Copy sync link" to the menu and make it copy on click. You can use tooltip.js to show a "Link copied" confirmation.

skjnldsv:
Can't get the clipboard to work, don't know why :/
Can somebody help me there? Is it working for you?

I added some feedback on the console as a debug, but even if everything seems normal, my clipboard stays empty 😢

js/components/addressBook/addressBook_controller.js
```diff
@@ -141,6 +141,12 @@ angular.module('contactsApp')
 	};
 
 	ctrl.onSelectSharee = function (item) {
+		// Prevent settings to slide down
+		$('#app-settings-header > button').data('apps-slide-toggle', false);
```
skjnldsv: This is an ugly trick. But core is broken for popups in the settings area.
Here is the issue: https://github.com/nextcloud/server/blob/25acce7e0d992b10b930472b8138240b7bd9af13/core/js/apps.js#L108

skjnldsv:
@nextcloud/designers @nextcloud/contacts All clear.
Screenshots are updated on first post.

Please review! 🐤

irgendwie:
@skjnldsv The 'Copy URL' tooltip is a bit glitchy for me. When clicking 'Copy URL' the 'Copied!' tooltip appears shortly in the top left corner of the screen. When hovering the 'Copy URL' (after closing and reopening the settings) the tooltip still shows 'Copied!' - only after some time it changes back to the original help text.

jancborchardt:
@skjnldsv any update? Would be great if we can get this in time for Nextcloud 12 :)

@irgendwie I would say the tooltip glitch we can fix separately? :) Let’s move forward and not block this PR too long.

skjnldsv:
@jancborchardt this update is ready to be merged.
Just the tooltip bug @irgendwie had.

irgendwie:
@jancborchardt Would love to merge, still it would break the contacts app for nc 9 and nc 10 and would only be compatible with nc 11 and 12. Are we willing to do this?

skjnldsv:
For the record, I am :)

jancborchardt:
cc @LukasReschke @MorrisJobke @jospoortvliet for second opinions.

I would agree with @skjnldsv it’s ok for the Contacts app to only support the two latest releases. It’s a big testing workload otherwise and also contacts syncing is in core anyway. And the app in its current state is still available for Nextcloud 9 and 10.
We want people to update to new versions also for security reasons, so getting proper updated apps is another push to do so.

Henni:
@jancborchardt I agree. But we should really support the latest two versions and thus might want to wait for the nc 12 release.

jancborchardt:
> @jancborchardt I agree. But we should really support the latest two versions and thus might want to wait for the nc 12 release.

Yes, good call – do you mean after or before the Nextcloud 12 release? And for when is the next release of the Contacts app planned? Nextcloud 12 will come soon. :)

Henni:
@jancborchardt how about the same day as Nextcloud 12? :wink: 

The current milestone has a due date of April 1st. But it might make sense to delay it to the nc release.

jancborchardt:
@Henni cool, let’s do that! :) But then let’s also merge this change already so it can be used and tested as early as possible. :) Ok @irgendwie @Henni @skjnldsv?

PR #104, opened 2017-01-25
Fix grouplist height

skjnldsv:
Might not be needed if this only occurs on core master since nextcloud/server#3260 has a fix for that.

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -503,7 +503,7 @@ ul.addressBook-share-list li {
 }
 
 #app-navigation > ul {
-	height: calc(100% - 24px);
+	height: calc(100% - 68px);
 }
 
 /* styling for contact count in contact list */


ChristophWurst:
> Might not be needed if this only occurs on core master since nextcloud/server#3260 has a fix for that.

Then let's wait for the server PR and close this then? :-P

skjnldsv:
It's also on stable11 apparently, so we still need this I think?

ChristophWurst:
Okay, I'd rather fix this here then instead of backporting too much.

ChristophWurst:
How can I test this?

skjnldsv:
Well, to really a test, but since we have the new contact button that is 68px high, we need to calc the 100% minus the new button height! ^^
capture d ecran_2017-01-26_19-02-56

skjnldsv:
capture d ecran_2017-01-26_19-03-36

PR #114, opened 2017-02-05
Photo warning

### LINKED ISSUE #113: Add error notification if the photo has been dismissed
following #102 
reference #85

skjnldsv:
Fix #113

- [x] Validate function
- [x] Debug logs
- [x] Design
- [x] Testing
- [x] ~~Why is the groups test failing with the `console.debug`~~ Will see later

### DIFF
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -357,13 +357,13 @@ angular.module('contactsApp')
 						if(property.value.join(';').indexOf(',') !== -1) {
 							this.failedProps.push(prop);
 							property.value = property.value.join(',').split(',');
-							//console.warning(this.uid()+': Categories split: ' + property.value);
+							//console.warn(this.uid()+': Categories split: ' + property.value);
 						}
 					} else if (angular.isString(property.value)) {
 						if(property.value.indexOf(',') !== -1) {
 							this.failedProps.push(prop);
 							property.value = property.value.split(',');
-							//console.warning(this.uid()+': Categories split: ' + property.value);
+							//console.warn(this.uid()+': Categories split: ' + property.value);
 						}
 					}
 					if(property.value.length !== 0) {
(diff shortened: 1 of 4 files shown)


irgendwie:
Works! The only problem that I have is, that it's not obvious, that the image got removed. Maybe we should information about that in the tool-tip from the little ! sign. For me that's the only obvious information source of failed properties (apart from the red mark of input fields).

skjnldsv:
The image is removed on purpose. We can't guess what types the image is. So we have an unusable binary base64 string :/

js/models/contact_model.js
```diff
@@ -339,10 +339,17 @@ angular.module('contactsApp')
...
-							this.failedProps.push(prop);
-							this.removeProperty('photo', property);
-							property = undefined;
-							console.debug(this.uid()+': Photo removed');
+							if (MimeService.b64mime(property.value) !== '') {
+								this.failedProps.push(prop);
+								property.meta.type = MimeService.b64mime(property.value);
+								console.debug(this.uid()+': Photo detected as ' + property.meta.type);
+							}
+							else {
```
skjnldsv: please on the same line :p

js/models/contact_model.js
```diff
@@ -307,6 +309,7 @@ angular.module('contactsApp')
 				return matchingProps.length > 0;
 			},
 
+			/* eslint-disable no-console */
```
Henni: This, the two code comments (L321, L327) and the `console.debug` calls (L351, L356) should be removed.

Henni: Ah, this seems to be intended. Should we hide this behind some flag?
(`debug` sounds a bit wrong, maybe `console.error`, `console.warning` or `console.info`might be better.)

skjnldsv: Warning seems good to me :)

PR #121, opened 2017-02-06
Sort functionality (visible in settings, bottom left)

irgendwie:
Downstream from https://github.com/owncloud/contacts/pull/569

* Added sort functionality (visible in settings, bottom left)

* Fallback to fullName if the sort criteria is not set within the contact; displayName is now the default sort criteria

* Changing display name based on sort order; sorting takes both first and second name in consideration


michaelletzgus:
Nice. :-)

Small enhancement:
When sorting by last name and there is no first name available a comma ',' is shown in the contact list.
IMHO this ',' should be suppressed when first name is empty (or white space only?)

I think a better location would be top left above the "new contact" button or top right near the search input field.

PR #123, opened 2017-02-07
Allow user to import contacts into custom addressbook

### LINKED ISSUE #55: Allow vCard importing into custom AddressBook
### Steps to reproduce
1. Create a new AddressBook.
2. Import a `.vcf` file.
3. Note that the AddressBook of all of the contacts is the default `Contacts` addressbook. 

### Expected behaviour
There should be a field or other option to import into a custom AddressBook

### Actual behaviour
There is no such option.

### Server configuration
**Nextcloud version:** (see Nextcloud admin page)
`10.0.1`

**Contacts version:** (see Nextcloud apps page)
`1.4.0.0`

**Updated from an older Nextcloud or fresh install:**
Fresh install

### LINKED ISSUE #89: Importing large contacts file is very heavy
What we should do:
- [x] Disable contact switching when importing contacts
- [x] Disable render of templates 
- [ ] ~~Only show a progress with a nice animation~~
- [x] Show the first contact of the list when done

skjnldsv:
fix #55, fix #89 

- [x] Ui-select
- [x] Design
- [x] Controller edit
- [x] Directive fix
- [x] Disabled addressbook input autotomplete, autocorrect and spellcheck
- [x] Settings closing when clicking on select

Vcard test example of 500 contacts: [500.txt](https://github.com/nextcloud/contacts/files/1211762/500.txt)

capture d ecran_2017-02-07_15-07-43

@nextcloud/contacts please review


tcitworld:
IMHO, the UI would be much better if a popup opened after selecting the file to choose the addressbook in which to import it, just like the calendar app does.

skjnldsv:
And implementing #89 ?
Could be great, but require far much work ^^

skjnldsv:
@tcitworld I disagree with you. The popup on calendar just add a useless step on choosing where you want to import. Having everything on the same location is much more easier UX wise that having multiple windows that open forcing the user to look away and move his mouse somewhere else.

tcitworld:
You have a point, but it doesn't feel practical with if the list of addressbooks is too long. Let's say most people won't have many addressbooks.

skjnldsv:
Yes most of them. But you just made a nice point too, I need to enable the search in case of too many addressbooks 😝

Tost69:
May be most of the people are using one addressbook only, but what if it's not enough?
I have already two: one for private contacts, one for work. Additionally I have two more shared addressbooks in my account. More than 400 entries totally and for now there is no way to view only work contacts, I have to fiddle them out of the big list one by one. And if I want to export some of them there is only the possibility to export the single contacts one after another or I can choose to export the whole selected addressbook (pretty much ignoring that I only need some members of a contact group to be exported).
How about to bring back the addressbook selection checkboxes next to the download/link/share/delete buttons in the config section like in the Owncloud contacts app and a checkbox in the contact item panel for further actions like export or adding to a group? That would be really nice (and yes, I know it's not a oneliner).
Nevertheless thanks for your great work, kind regards, Tom

skjnldsv:
- [ ] FIX: After importing contacts, clicking a contact doesn't show anything until reload
- [ ] Better progress animation

jancborchardt:
Cool, good progress @skjnldsv! :) Say when it’s fully ready to review :)

skjnldsv:
@jancborchardt, @ChristophWurst, @MorrisJobke, @Henni, @irgendwie This is finally done, please review! 😄 💟 

Tests: Import the big file from the first post
 - See the loading animation with the spinner
 - See the percentage of completion increasing
 - See no change on the url nor the contacts while the import is in background
 - See the load of the first contact once everything is done
 - Try the search (if more than 5 addressbooks)

skjnldsv:
We will need something to block the user from clicking a group or a contact in the detail list and show only a spinner in background. Should we do that here or in another pr?

skjnldsv:
Okay, I want to change my mind about the loading animation. 
I will create a progress bar with some data and a cancel button (if possible) in the main content section (replacing the contacts list and the contact details) and disable the group selection and the settings section because we need to prevent the user from interacting with the app since it can lead to errors like :
- deleting the addressbook we're importing in
- switching contacts (which won't work because the import requests are sent before)

jancborchardt:
Can you review this one @jonatoni @xh3n1? :)

skjnldsv:
@jancborchardt wait a bit, I want to add an import progress screen. This is not ready for review yet, We'll ship this with the next version thought :)

skjnldsv:
Aaaaaaaand, we've got an import screen (with theming support)
contacts nextcloud

skjnldsv:
@nextcloud/contacts @jancborchardt Any change wanted? What do you think of it?

skjnldsv:
Percentage added
kazam_screencast_00001

skjnldsv:
With name 
capture d ecran_2017-08-13_21-42-17

- [ ] Prevent failing of displayname if xhr fail

jancborchardt:
Cool! :) I'd prefer if we don't block interaction with the app, for example if import takes some time? I still want to see my existing contacts and not be blocked by import. (Just lile File upload doesn't block everything.)

We should have the progress bar be in the botton left, the "Importing" bar itself with the spinner. I think we used to have it like that, no?

jancborchardt:
And the text could just say "Importing 34% …"

skjnldsv:
You didn't read my message correctly. Clicking anything on the page don't work if an import is in progress. Your browser is adding all of its requests chronologically, so if your vcf file has 200 contacts and you decide to click a group (for example), your group requests to the server will be added AFTER the 200 contacts creation request. 

So there is no need allowing the user clicking anything. He, won't understand why the app isn't responding.

jancborchardt:
Hhmm ok. Bummer. :/

Well then we should go with the usual class="emptycontent" view. Check how it's layed out in the files app.

Basically the layout should be:
> big icon-loading as icon
> **Importing 34% …** (as h2)
> Below a small progress bar looking like the small quota bar bottom left in the files app 
> Below that maybe: `Name, 20 of 255`

skjnldsv:
@jancborchardt do you mean adding all of that in the settings section?
I'll strongly advice against that since there is no need to hide this. It's not a small information, this is a main screen which needs to be on front. Unlike the quota bar, this is a full view. :)

If you didn't meant that, forget what I just said ;)

Anyhow, I don't see the point of having a loader icon next to a progress bar since there are basically the exact same thing, a loader :)

Aside from the design of my import screen, which I'm willing to argue about, I think I maxed out the most of the essential information without having unneeded stuff! :)

jancborchardt:
No, all of what I wrote should be in the main view. Like the emptycontent view for empty folders.

Basically what you do right now, but with the design slightly improved. :)

skjnldsv:
Here we go:
kazam_screencast_00004

jancborchardt:
Nice! :) Just details:

- When there's only one address book, the button should be called "Import into Test 1" without the dropdown below. Otherwise there are too many elements and it looks complicated.
- The "Currently" in the importing title can be cut
- There should be a big icon-loading over the heading
- The heading is usually centered
- The percentage is probably better placed in the bottom right, under the bar (cause it's styled like the contact name also)

skjnldsv:
This is what it looks on mobile
kazam_screencast_00001

- [ ] Auto hide sidebar if mobile

xh3n1:
It looks really good :+1:

jancborchardt:
Nice! As said I would just add an icon above the text like in the usual emptycontent views. :) `icon-upload`

skjnldsv:
Done 
capture d ecran_2017-08-15_08-44-01

jancborchardt:
:+1: all good from my side now, looks very good! :)

jancborchardt:
Another review please @nextcloud/javascript? :)

skjnldsv:
@jancborchardt Still waiting for the sidebar block to avoid user interaction too :)

skjnldsv:
Okay, I need to fix the addressbook deletion update, But since this is a bigger work, I will do it in another pr.

Please review!! 😄 ❤️ 

**Test**: _Import the big file from the first post_
- See the loading animation with the spinner
- See the percentage of completion increasing in the main screen
- See the imported user switch under the progress bar
- See no change on the url nor the contacts while the import is in background
- See the load of the first contact/group once everything is done
- Try the search (if more than 5 addressbooks)
- See the auto-hide of the sidebar on mobile when the import starts

skjnldsv:
Okay, I need to fix the addressbook deletion update, But since this is a bigger work, I will do it in another pr. 

@jancborchardt, @ChristophWurst, @MorrisJobke, @Henni, @irgendwie, @jonatoni, @xh3n1 
Please review!! 😄 ❤️ 

**Test**: _Import this big file_: [500vcards.txt](https://github.com/nextcloud/contacts/files/1211762/500.txt)

- See the loading animation with the spinner
- See the percentage of completion increasing in the main screen
- See the imported user switch under the progress bar
- See no change on the url nor the contacts while the import is in background
- See the load of the first contact/group once everything is done
- Try the search (if more than 5 addressbooks)
- See the auto-hide of the sidebar on mobile when the import starts

nursoda:
Works :-)

I spotted one glitch:
- have one **a**ddress**b**ook 'first'
- create another ab 'second'
- 'import into' shows dropdown, one may select 'first' and 'second'
- switch that dopdown to 'second'
- delete ab 'second' through it's context menu (three dots)
- 'import into' still shows 'second'(!)

Nice to have:
- an 'abort import' button, especially for big imports

skjnldsv:
@nursoda all the bug you found are known, I created a problem for the contacts/groups list update after deletion, and will also fix the selection at the same time! 
Thank you very much for your review, it's perfect! 

@jonatoni & @xh3n1 thanks!! 😉

devillemereuil:
I just tested this PR on Nextcloud 12. I've found two issues:
- When imported the first time, the contacts cannot be displayed. I get "No Contacts Here". If I quit Contacts (i.e. change to another Nextcloud App) and come back again, then they are displayed.
- Once they are displayed, the first contact to be displayed belong is assigned to the correct AddressBook (field at the bottom), but all the others are assigned to the first, pre-existing addressbook. Doesn't seem that they actually are misassigned though, because removing the addressbook does remove all the imported contacts. So, probably just a display issue.

I tested the 500 contacts provided by @skjnldsv and another one of mine waiting for this awesome feature to be imported! ;)

skjnldsv:
@devillemereuil It's possible that your server is having trouble loading the contact you just clicked since it has received a lot of requests after the import.
I can't reproduce here :/

devillemereuil:
Possible. The contacts do not seem to be sync with CardDAV either... Weird. I'll let it settle a bit and report back if it stays the same...

devillemereuil:
A few hours after uploading my VCFs, during which I even rebooted the server, I'm still having the same issues as described above... Also the contacts are not visible through CardDAV (the address books are synced as empty...). Weird. Might be something wrong in my install...

PR #125, opened 2017-02-08
Fix typo in "Show URL"

nickvergessen:
See question at https://www.transifex.com/nextcloud/nextcloud/translate/#de/contacts/104496499

I don't think the missing space was intended?


skjnldsv:
Indeed ^^

PR #127, opened 2017-02-09
fix labels in detail items

Henni:
Fixes double ids in case of multiple occurrences of the same detail item (multiple addresses).


skjnldsv:
👍 🙈

PR #159, opened 2017-03-09
Missing border in Contacts

Dennis1993:
I use your demo of NextCloud and founded a missing border. On the left is a grey border for the sidebar. On the next 2 divs is no border anymore. It would be good, if a border is here too to seperate it.

unbenannt-1

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -435,6 +435,7 @@ contactlist .tooltip {
 	height: 100%;
 	overflow-x: hidden;
 	overflow-y: auto;
+	border-right: 1px solid #eee;
 }
 
 .app-content-detail {


MariusBluem:
Example Screenshot with your Fix: 

bildschirmfoto 2017-03-09 um 20 15 56

irgendwie:
Thank you! :+1:

Dennis1993:
Thank you for your quick feedback!

jancborchardt:
@Dennis1993 thanks for your contribution! :) It would be cool if you can fix it in the server too so it’s also applied to other apps which use this layout (like Mail https://github.com/nextcloud/mail)

You just need to add the code blocks `.app-content-list` (and all below starting with that) as well as `.app-content-detail` into this file: https://github.com/nextcloud/server/blob/master/core/css/apps.scss :)

PR #172, opened 2017-03-19
Added rename addressbook functionality.

irgendwie:
Needs #101 

Adds the possibility to rename addressbooks and can be found in the settings menu.

rename
2017-03-31-132144_239x243

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -304,15 +304,6 @@ avatar.failed {
 }
 
 /* addressbook settings */
-input.renameAddressBookInput {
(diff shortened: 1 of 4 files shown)


jancborchardt:
As you can see from the second screenshot, it’s a bit confusing with the »Address book name« input field.

It would be good to put the »Add new address book« function behind a button to clear up the confusion, also in general when you open the settings. cc @skjnldsv @irgendwie @nextcloud/contacts what do you think?
(Simply a button, which on click reveals this input field in its place, being focused.)

PR #173, opened 2017-03-20
Add contacts menu integration

ChristophWurst:
WIP - requires https://github.com/nextcloud/server/pull/3233

TODO
- [x] Correct namespacing
- [x] Routing
- [x] Hide *other* contacts, e.g. from the system address book https://github.com/nextcloud/server/pull/3233#issuecomment-292185893


jancborchardt:
@ChristophWurst is this already up for first review or too much in-development state? :)

ChristophWurst:
still WIP

jancborchardt:
A few bugs so far:

- [ ] The list in the Contacts menu shouldn’t show yourself at all.
- [ ] Neither the »i« Details icon for internal users. :) (At least not yet where we don’t have a detail page for them.)
- [ ] When clicking on a details icon for a contact, it opens the Contacts app and loads the first contact as usually when opening the app. It should load that specific contact though.

cc @nextcloud/contacts help appreciated. :)

ChristophWurst:
> Hide other contacts, e.g. from the system address book nextcloud/server#3233 (comment)

@nextcloud/contacts any idea how I can distinguish between vcards from the system address book and the user address book? The data structures looks similar unfortunately, there's nothing specific to contacts from the contacts app it seems 🤔

nickvergessen:
If you have access to the URI, it should contain `system/system`

ChristophWurst:
> If you have access to the URI, it should contain system/system

Thanks. Apparently the URI for system contacts is their UID. However, I've found what I was looking for: `isLocalSystemBook` is set for system contacts 😂

lib/ContactsMenu/Providers/DetailsProvider.php
```diff
@@ -60,6 +57,11 @@ public function process(IEntry $entry) {
 			return;
 		}
 
+		if ($entry->getProperty('isLocalSystemBook') === true) {
+			// Cannot show details -> ignore
+			return;
```
jancborchardt: Will this still show their email address though? Because that can be set in the personal details. :) The only thing which should be hidden is the details icon, not the whole entry.

ChristophWurst: email actions are added by the server, also for internal users

ChristophWurst: it doesn't hide the entry. it just doesn't add any action 😉

ChristophWurst:
> When clicking on a details icon for a contact, it opens the Contacts app and loads the first contact as usually when opening the app. It should load that specific contact though.

@skjnldsv you mentioned that it's easy to add that. Could you please tell me where I have to adjust the code? I presume it starts with a dedicated route in `main.js`. But how do I get the `gid` of that contacts to be able to *redirect* to `/gid/uid`? Is that even possible?

ChristophWurst:
> @skjnldsv you mentioned that it's easy to add that. Could you please tell me where I have to adjust the code? I presume it starts with a dedicated route in main.js. But how do I get the gid of that contacts to be able to redirect to /gid/uid? Is that even possible?

Kinda figured out how to possibly do that: https://github.com/nextcloud/contacts/pull/173/commits/026d509336775b9d6bf4c1d10849f03a2e9e5692
Still, that route doesn't seem to work when I'm already in the contacts app. It then always shows the first entry.
cc @nextcloud/contacts

irgendwie:
@ChristophWurst found the reason (https://github.com/nextcloud/contacts/blob/contactsmenu-integration/js/components/contactList/contactList_controller.js#L142) and fixed it with a commit.

ChristophWurst:
@irgendwie thanks a lot \o/

PR #182, opened 2017-04-03
Use PSR-4 naming for classes

ChristophWurst:


rullzer:
shouldn't it be in `/lib`?

ChristophWurst:
> shouldn't it be in /lib?

Seems to work outside too. But yes, let's move it.

ChristophWurst:
Moved the controller to `lib` and adjusted the makefile

Issue #110, opened 2017-02-02
add an option to enable callto link after phone numbers


adsworth:
it'd be nice if a callto link could be displayed after phone numbers. This would allow users to initiate calls directly from the contacts app.

adsworth:
the same could be said for email addresses. :)

skjnldsv:
Good idea indeed!

skjnldsv:
Should we display it only in mobile view?

adsworth:
I'd say yes definitely.

j-ed:
@skjnldsv Why should you limit the availability of callto links only to the mobile view?
I could imagine that users would use the auto dial feature on a desktop PC to, especially
if Nextcloud is used in a business environment.

adsworth:
I'd definitely say on both views. Personally I don't need it in the mobile view as I sync to my android address book.

skjnldsv:
@j-ed I guess you're right! And since this is a protocol, other apps may be using it.

skjnldsv:
Fixed by #193

PR #193, opened 2017-04-19
Added tel/mail link to tel/mail field

### LINKED ISSUE #187: Add tel: links for phone numbers
I regularly look up phone numbers from my computer. To place the call I have to copy&paste the number in skype/hangouts/etc (or use Join to send it to my phone in my case).

A simple link with 'href="tel:PHONENUMBER"' would suffice to be able to call numbers with any application by just clicking it but I'm not sure what would be the best place to put it. The options are:

- add a toggle to switch between editing the number and showing a clickable link (adds an extra click to many workflows)
- add a "call" button next to the trash icon
- somehow make the number in the input clickable (possible but not recommended)
- add a modifier for clicks on the number (shift+click = place call)

In my opinion, adding a "call"-button in [templates/detailItems/tel.html](https://github.com/nextcloud/contacts/blob/02ceb6732857e15c9b905a7390ee4506d29fbcd4/templates/detailItems/tel.html) should be the best solution since it wouldn't interfere with any logic and there doesn't seem to be any space restrictions/wrapping issues around the trash icon.

Since this is my first project for nextcloud, it would be great to get some input if this is possible, wanted and what might be the best approach.

irgendwie:
Adds the tel link to the tel field and fixes #187

TODO Although ng-if is specified, the icon is shown when the field is empty.

@juzim @Henni @skjnldsv

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -271,6 +271,12 @@ detailsitem .item-action {
 	right: 33px;
 }
 
+detailsitem .item-action i {
+	display: block;
+	height: 20px;
+	width: 20px;
+}
+
 detailsitem.details-item-adr .icon-delete, detailsitem.details-item-n .icon-delete {
 	vertical-align: middle;
 	left: 251px;
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -262,8 +262,7 @@ detailsitem .icon-delete {
 	opacity: .2;
 }
 
-detailsitem .url-link,
-detailsitem .tel-link {
+detailsitem .item-action {
 	position: absolute;
 	padding: 8px 10px;
 	margin-left: -30px;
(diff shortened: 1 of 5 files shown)


css/public/style.css
```diff
@@ -262,7 +262,8 @@ detailsitem .icon-delete {
 	opacity: .2;
 }
 
-detailsitem .url-link {
+detailsitem .url-link,
+detailsitem .tel-link {
```
Henni: can we find something more generic, which we could use for both?

irgendwie: e.g. link-action-icon?

irgendwie:
ref. #110

Additionally there is now an icon next to e-mail addresses with mailto links.

Issue #190, opened 2017-04-17
Can't create or import contacts


Woefdram:
### Steps to reproduce
1. Open Contacts
2. Click "New Contact"
3.

### Expected behaviour
New, empty card should appear.

### Actual behaviour
"Contact could not be created" appears at the top of the screen, nothing more. Devel console shows message "TypeError: contact is undefined"

### Server configuration
Debian Testing ("Stretch"), kernel 4.9.18-1 (amd64)

**Web server:**
Apache 2.4.25

**Database:**
MariaDB 10.1.22

**PHP version:**
7.0.16

**Nextcloud version:** (see Nextcloud admin page)
11.0.2 (stable)

**Contacts version:** (see Nextcloud apps page)
1.5.3

**Updated from an older Nextcloud or fresh install:**
Fresh install.

**Signing status:**
?

```
Login as admin user into your Nextcloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
```
No errors have been found.

**List of activated apps:**
  - activity: 2.4.1
  - bookmarks: 0.9.1
  - calendar: 1.5.2
  - comments: 1.1.0
  - contacts: 1.5.3
  - dav: 1.1.1
  - federatedfilesharing: 1.1.1
  - federation: 1.1.1
  - files: 1.6.1
  - files_pdfviewer: 1.0.1
  - files_sharing: 1.1.1
  - files_texteditor: 2.2
  - files_trashbin: 1.1.0
  - files_versions: 1.4.0
  - files_videoplayer: 1.0.0
  - firstrunwizard: 2.0
  - gallery: 16.0.0
  - gpxedit: 0.0.5
  - gpxpod: 2.1.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.0.0
  - nextcloud_announcements: 1.0
  - notes: 2.2.0
  - notifications: 1.0.1
  - ojsxc: 3.1.1
  - password_policy: 1.1.0
  - provisioning_api: 1.1.0
  - serverinfo: 1.1.1
  - sharebymail: 1.0.1
  - spreed: 1.2.0
  - spreedme: 0.3.8
  - survey_client: 0.1.5
  - systemtags: 1.1.3
  - tasks: 0.9.5
  - theming: 1.1.1
  - twofactor_backupcodes: 1.0.0
  - updatenotification: 1.1.1
  - user_ldap: 1.1.2
  - workflowengine: 1.1.1

**The content of config/config.php:**
$CONFIG = array (
[...]

PR #197, opened 2017-04-19
Fix dav lib to send corrent content-type when using PUT requests.

### LINKED ISSUE #196: Send correct content-type in AJAX
Adding a new contact sends a PUT request with the following header:

```
Content-Type: application/xml;charset=UTF-8
```

The PUT body, however, is a text/vcard document. 

This is a problem with mod_security which rejects the requests because it cannot parse XML in a request body that's supposed to be XML, so it's a violation of the HTTP protocol.

I think this could be solved by adding:

```
xhr.setRequestHeader("text/vcard;charset=UTF8");
```

somewhere in the AJAX flow, but I was unable to locate the right place.

irgendwie:
should fix #196

@gjedeer you can easily test this by replacing the js/dav/dav.js file

### DIFF
--- a/js/dav/dav.js
+++ b/js/dav/dav.js
@@ -938,7 +938,7 @@ exports.createAccount = _co2['default'].wrap(regeneratorRuntime.mark(function ca
 }));
 
 // http redirect.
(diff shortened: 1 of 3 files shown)


irgendwie:
@Woefdram since you also had an issue with ModSecurity, maybe you can leave it enabled with this change. (ref. #190)

PR #198, opened 2017-04-22
Fix url fields without http

irgendwie:
- Clicking the url action icon can lead to relative links
- This PR is a continuation of https://github.com/owncloud/contacts/pull/366
- It fixes this by adding the http protocol to the URI if no other protocol is specified


jancborchardt:
@nickvergessen @MorrisJobke @nextcloud/contacts can you review this? :)

js/tests/parsers/urlModel_directive.js
```diff
@@ -0,0 +1,46 @@
...
+		$scope.model.url = '3';
+		expect(form.url.$modelValue).to.equal('http://3');
+	});
+	it('Don\'t add http protocol', function() {
+		form.url.$setViewValue('http://owncloud.org');
+		expect($scope.model.url).to.equal('http://owncloud.org');
+		form.url.$setViewValue('https://owncloud.org');
+		expect($scope.model.url).to.equal('https://owncloud.org');
+	});
+	it('should recognize different protocols, apart from http(s)', function() {
```
skjnldsv: Should we allow ftp or were you testing against it?

js/tests/parsers/urlModel_directive.js
```diff
@@ -0,0 +1,46 @@
...
+		expect($scope.model.url).to.equal('http://3');
+
+		$scope.model.url = '3';
+		expect(form.url.$modelValue).to.equal('http://3');
+	});
+	it('Don\'t add http protocol', function() {
+		form.url.$setViewValue('http://owncloud.org');
+		expect($scope.model.url).to.equal('http://owncloud.org');
+		form.url.$setViewValue('https://owncloud.org');
+		expect($scope.model.url).to.equal('https://owncloud.org');
```
jancborchardt: Can we have different testing urls, like nextcloud.com? ;)

js/tests/parsers/urlModel_directive.js
irgendwie: I was just adding additional test cases since I was thinking about a generic url recognition
May be overpowered 🤔

irgendwie:
If we assume one would just add websites (http, https) we should check for the short format `http:` and the normal `http://` (and both with the ssl extension) - if none is given simply add it.
This potentially destroys other URIs. Officially allowed are much more formats (see https://tools.ietf.org/html/rfc3986#section-3)

@jancborchardt What says the designer? We could also simply leave the url untouched. 🐳

js/tests/parsers/urlModel_directive.js
skjnldsv: I think we should only allow ftp and http, much easier.
Other protocols like facebook, twitter... goes into X-SOCIAL :)

jancborchardt:
If no protocol is specified, I’d say prepending either https:// or http:// should be fine. Otherwise it might not be linkified, no?

skjnldsv:
@Henni Yes! :)
Like I commented earlier: 
>Should we allow ftp or were you testing against it?

js/components/parsers/urlModel_directive.js
```diff
@@ -0,0 +1,34 @@
+angular.module('contactsApp')
+.directive('urlModel', function() {
+	return{
+		restrict: 'A',
+		require: 'ngModel',
+		link: function(scope, element, attr, ngModel) {
+			ngModel.$formatters.push(function(value) {
+				if (typeof value !== 'string') {
```
skjnldsv: @Henni I get an issue here. All tests using the formater assignment are null at this point! :/
```
$scope.model.url = 'formatters.nextcloud.com';
expect(form.url.$modelValue).to.equal('https://formatters.nextcloud.com');
```
https://github.com/nextcloud/contacts/pull/198/files#diff-19123d57bdd04dbac14aca8a241ae1d2R29

Henni: Adding `$scope.$digest(); after [this line](https://github.com/nextcloud/contacts/pull/198/files#diff-19123d57bdd04dbac14aca8a241ae1d2R29) fixes it for me.

skjnldsv: @Henni Yeah, I knew it was going to be something like that  --"

skjnldsv:
@Henni somehow two tests keeps failing. Can you take a look? :)

PR #199, opened 2017-04-22
Fix :active background for contactdetails header input fields.

irgendwie:
When **clicking** on an input field in the contactdetails header, the background becomes white (Nextcloud, current master).

2017-04-22-162454_436x105

This PR fixes this, by setting the background to transparent when `:active`.

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -19,6 +19,9 @@
 	display: flex;
 	flex-wrap: wrap;
 }
+.contactdetails__header input[type=text]:active {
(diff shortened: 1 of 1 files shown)


jospoortvliet:
@jancborchardt @MorrisJobke @nickvergessen - it looks sane to me but better have somebody else have a look, too ;-)

jancborchardt:
Yup, looks way nicer transparent as it was before. :) :+1: cc @nextcloud/designers @jonatoni @xh3n1

Issue #108, opened 2017-01-27
provide vcard GENDER field


hirschrobert:
Thanks for the great work.

Is it possible to provide the GENDER field (https://tools.ietf.org/html/rfc6350#section-6.2.7)? This can be further used to generate email/letter salutation (Dear Mr./Dear Ms.).

PR #214, opened 2017-05-13
provide gender field

hirschrobert:
tried to contribute to my issue #108 . If you could have a look at it. am new here and hope it is the right track. many thanks


MariusBluem:
Well ... Do you have some example clients who render the gender in your vCard? 

Especially the non-male/female part 🤔 ... I think we should replace this with something like "other" or whatever those CardDAV-clients do ;)

hirschrobert:
when i download it in nextcloud it gives this file: [a94d959f-a9e9-4d18-b887-4664370e4eda-4.vcf.txt](https://github.com/nextcloud/contacts/files/998564/a94d959f-a9e9-4d18-b887-4664370e4eda-4.vcf.txt)

According to https://tools.ietf.org/html/rfc6350#section-6.2.7 that seem to be right.

l10n/de_DE.js
```diff
@@ -70,6 +70,10 @@ OC.L10N.register(
     "Voice" : "Anruf",
     "Social network" : "Soziales Netzwerk",
     "Details" : "Details",
-    "Settings" : "Einstellungen"
+    "Settings" : "Einstellungen",
+    "Gender" : "Geschlecht",
+    "Male" : "männlich",
+    "Female" : "weiblich",
+    "intersex" : "intersexuell"
```
MariusBluem: Translations are handled over Transifex 🙈

hirschrobert: thanks. do i understand [this](https://docs.nextcloud.com/server/11/developer_manual/core/translation.html#automated-synchronization-of-translations) right? I don't need to take care of translation? Your server is doing the scanning of new labels, pushes it to transifex, your crew will translate it and pushes it to github? All automatically? Thanks!

irgendwie: @hirschrobert That's correct - you just have to add "t('contacts', ...)" in the code and the rest is handled automatically.

skjnldsv:
No, don't delete the files, delete the edit from the commit :)

xh3n1:
Female
Male
None
Other
Unknown

MariusBluem:
Male
Female
Other

None/Unknown can be done by simply not adding this gender-field.

jonatoni:
thanks @nickvergessen 😜

MariusBluem:
Thanks for contributing 🙈

Issue #253, opened 2017-07-04
Show fields for phone, email, address and groups by default for new contact


jancborchardt:
When pressing the »New contact« button, you are faced with only the header for picture and name, as wel as the groups dropdown.

Phone, email and address should be displayed for any new contact. It was like that before too.

xh3n1:
The problem with adding the 3 props tel email and adr to the template is that after they are created in the controller they do not have a $$hashKey. I tried multiple things, such as adding track by value option to the foreach loop but no good because it said it contained dupes. Therefore, I'm stuck atm and don't know how to proceed. Any advice would be appreciated.

MorrisJobke:
> The problem with adding the 3 props tel email and adr to the template is that after they are created in the controller they do not have a $$hashKey. I tried multiple things, such as adding track by value option to the foreach loop but no good because it said it contained dupes. Therefore, I'm stuck atm and don't know how to proceed. Any advice would be appreciated.

@xh3n1 Do you have the code change available somewhere? Then I can have a look. Angular is sometimes a bit tricky to tame 😉 

cc @georgehrke

xh3n1:
@MorrisJobke There is no code change. The issue is that when the 3 properties are added on line 11 the created objects do not have a hashkey and therefore the template is not updated. What do you suggest?

georgehrke:
@xh3n1 Can you point us to the code in question? :)

georgehrke:
For reference: https://github.com/nextcloud/contacts/blob/master/js/components/newContactButton/newContactButton_controller.js#L11

xh3n1:
hey @MorrisJobke, I just did a little test by adding $$hashKey property to the created tel object/prop and to the value level. It was still unresponsive, meaning it did not update the template accordingly. What I also suspect now is that the contact object on line 11 mentioned above is not the same with contact object being displayed in the template. Could you look into my assumptions? cc @nickvergessen  Thank you!

xh3n1:
cc @Henni @skjnldsv @irgendwie

georgehrke:
This is no angular issue, but simply a bug introduced by https://github.com/nextcloud/contacts/commit/028df62242f9756c043af4faaff0f4359baaf892

The contact object is created in this function, created on the server and stored in the cache:
https://github.com/nextcloud/contacts/blob/c506cd69e25325cc11d77b2da163e8a655af9734/js/services/contact_service.js#L161
Let's call it `contact1`.

[...]

PR #259, opened 2017-07-20
Show fields for phone, email, address and groups by default 

xh3n1:
Issue #253

### DIFF
--- a/js/services/contact_service.js
+++ b/js/services/contact_service.js
@@ -9,7 +9,6 @@ angular.module('contactsApp')
 	var observerCallbacks = [];
 
 	var loadPromise = undefined;
-
 	var newContactJustAdded = false;
 
 
@@ -210,13 +209,11 @@ angular.module('contactsApp')
 				filename: newUid + '.vcf'
 			}
 		).then(function(xhr) {
-			if (!(_.isUndefined(newContact.fullName()) || newContact.fullName() === '')) {
-				newContact.setETag(xhr.getResponseHeader('ETag'));
-				contacts.put(newUid, newContact);
-				notifyObservers('create', newUid);
-				$('#details-fullName').select();
-				return newContact;
-			}
+			newContact.setETag(xhr.getResponseHeader('ETag'));
+			contacts.put(newUid, newContact);
+			notifyObservers('create', newUid);
+			$('#details-fullName').select();
+			return newContact;
 		}).catch(function() {
 			OC.Notification.showTemporary(t('contacts', 'Contact could not be created.'));
 		});
--- a/js/services/contact_service.js
+++ b/js/services/contact_service.js
@@ -9,6 +9,7 @@ angular.module('contactsApp')
 	var observerCallbacks = [];
 
 	var loadPromise = undefined;
+
 	var newContactJustAdded = false;
 
 
@@ -209,11 +210,13 @@ angular.module('contactsApp')
 				filename: newUid + '.vcf'
 			}
 		).then(function(xhr) {
-			newContact.setETag(xhr.getResponseHeader('ETag'));
-			contacts.put(newUid, newContact);
-			notifyObservers('create', newUid);
-			$('#details-fullName').select();
-			return newContact;
+			if (!(_.isUndefined(newContact.fullName()) || newContact.fullName() === '')) {
+				newContact.setETag(xhr.getResponseHeader('ETag'));
+				contacts.put(newUid, newContact);
+				notifyObservers('create', newUid);
+				$('#details-fullName').select();
+				return newContact;
+			}
 		}).catch(function() {
 			OC.Notification.showTemporary(t('contacts', 'Contact could not be created.'));
 		});
--- a/js/services/contact_service.js
+++ b/js/services/contact_service.js
@@ -1,5 +1,5 @@
 angular.module('contactsApp')
-.service('ContactService', function(DavClient, AddressBookService, Contact, $q, CacheFactory, uuid4) {
+.service('ContactService', function(DavClient, AddressBookService, Contact, $q, CacheFactory, uuid4, vCardPropertiesService ) {
 
 	var cacheFilled = false;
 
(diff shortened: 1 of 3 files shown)


MorrisJobke:
Hi @xh3n1 two little remarks:

1. could you revert the changes in `js/dav/dav.js` because they change the whitespace on over 1000 lines of code 😉 (and nothing else)
2. if you say in the comment "fixes #NUMBER" then github automatically closes this linked issue once the PR gets merged 😉 (see https://help.github.com/articles/closing-issues-using-keywords/ for details)

nickvergessen:
See https://help.github.com/articles/closing-issues-using-keywords/ if you hate the word fixes ;)

georgehrke:
@jancborchardt

jancborchardt:
@georgehrke yes that is the expected behavior. :) As these are the 3 most common properties. It's also the behavior of Android and iOS.

skjnldsv:
@jancborchardt the field isn't saved empty. @xh3n1'commit will add empty ones in the vcard.
On my android the fields are displayed empty, but nothing is saved on the vcard until I write somthing into those fields.

jancborchardt:
Right, of course it shouldn’t be saved empty, just displayed empty. A virtual field if you wish. ;)

xh3n1:
@skjnldsv I'm not sure I understand what you are saying, However I think a solution to what I understand is the problem is that there should be no fields saved with empty values. What do you think about this?

skjnldsv:
@xh3n1 when you click the new contact button, a new contact is created with the default REQUIRED fields according to the vcard rfc.
A basic contact MUST have one FN field set, but it cannot be empty, so we need it to be filled with something on creation or the server will refuse it.

That being said, if we create a temporary contact without any fields displayed BUT waiting to be synced as soon as the FN (name) field is filled, we could have something viable! :)

xh3n1:
So when I'm creating a contact now the behavior is that it is added to the list of contacts with empty fields and with the name of New contact. The expected behavior is that the new contact is not added until the name is changed from New contact to something else and no fields are saved in it unless they have a non empty string in them as values. Am I correct @skjnldsv ?

skjnldsv:
You're correct! 😉

jancborchardt:
@jonatoni @xh3n1 is this now changed as @skjnldsv asked? Then we can review. :)

xh3n1:
hey @skjnldsv  can you review it now?

js/services/contact_service.js
```diff
@@ -12,6 +12,8 @@ angular.module('contactsApp')
 
 	var newContactJustAdded = false;
 
+	var createflag=true;
```
skjnldsv: Add some space between the `=` :)

js/services/contact_service.js
```diff
@@ -210,14 +211,21 @@ angular.module('contactsApp')
 				filename: newUid + '.vcf'
 			}
 		).then(function(xhr) {
-			newContact.setETag(xhr.getResponseHeader('ETag'));
-			contacts.put(newUid, newContact);
-			notifyObservers('create', newUid);
-			$('#details-fullName').select();
-			return newContact;
+			if (_.isUndefined(newContact.fullName()) || newContact.fullName() === '') {
```
skjnldsv: Why not use the condition directly in the if line 218?

js/services/contact_service.js
```diff
@@ -185,14 +210,17 @@ angular.module('contactsApp')
 				filename: newUid + '.vcf'
 			}
 		).then(function(xhr) {
-			newContact.setETag(xhr.getResponseHeader('ETag'));
-			contacts.put(newUid, newContact);
-			notifyObservers('create', newUid);
-			$('#details-fullName').select();
-			return newContact;
+			if (!(_.isUndefined(newContact.fullName()) || newContact.fullName() === '')) {
```
skjnldsv: The contact will still have a valid fn set at this point. See line 202
```javascript
if (_.isUndefined(newContact.fullName()) || newContact.fullName() === '') {
    newContact.fullName(t('contacts', 'New contact'));		 			
}
```

js/services/contact_service.js
```diff
@@ -149,7 +156,25 @@ angular.module('contactsApp')
 				});
 				return addressBook
 					? DavClient.getContacts(addressBook, {}, [ contact.data.url ]).then(
-						function (vcards) { return new Contact(addressBook, vcards[0]); }
+						function (vcards) {
+
```
skjnldsv: Do not add unnecessary empty spaces :)

js/services/contact_service.js
```diff
@@ -10,6 +10,9 @@ angular.module('contactsApp')
 
 	var loadPromise = undefined;
 
+	var newContactJustAdded = false;
+
```
skjnldsv: Do not add unnecessary empty spaces :)

skjnldsv:
Okay, I'll list the problem with the creation of a local contact only:
- `contact_services.js`:
  - The create function is used by the **user input** AND the **import button**, so we can't remove the xhr (XMLHttpRequest) request part from this function or we'll break the import.
  - The create function is returning a **promise**, used by `ctrl.createContact` in the `newContactButton_controller.js`, so we need to **return something else** (*like the contact for example*) and edit the controller to use it instead.
  - My guess will be to create a new function for the local contacts, something like `createLocalContact()` which don't save a thing and that is used by `newContactButton_controller.js`.
  - When opening a contact, the function getById is called and generate a xhr request to retrieve the full vcard. As we don't send it to the server on creation, the card won't be found. We need to create a function to load the contact **from cache ONLY** and not from the server. We need to create another function that use the same system that `getById` but for local contacts and without server requests.
  - We could probably add a value to the contact model that indicate if the contact is local or not. This way we can indicate `contactDetails_controller.js` to use either `getById` or our freshly created function (`getLocalById`?) instead. As soon as a contact is synced, we should set this value to false since it won't be a local contact anymore.

- `newContactButton_controller.js` 
  - Don't use the `then` function since we're not returning a promise anymore and set the `create` function as a variable instead.

skjnldsv:
presentation sans titre

jancborchardt:
Seems to not work at the moment? Any updates or known issues? :)

jancborchardt:
Also was thinking about showing the fields for IM, Social network and Notes by default too. Then we have a very nice 2-column layout:
screenshot from 2017-09-15 15-10-14

PR #137, opened 2017-02-18
Validate rev prodid version and validation edit

skjnldsv:
Fix #129 
- Keep only one rev/version/prodid field in case of multiple ones
    - [x] Autosave
    - [x] Keep first item only
- [x] Added some animation for the save button

PR #264, opened 2017-07-26
Better photo error notification

skjnldsv:
Replace the ugly red background.

kazam_screenshot_00000

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -305,8 +305,13 @@ detailsitem.details-item-note label {
 detailsitem.failed {
 	box-shadow: inset 2px 0 #c5173a;
 }
-avatar.failed {
-	background-color: #c5173a;
+avatar .icon-error {
+	position: absolute;
+	width: 100%;
+	height: 100%
+}
+avatar .icon-error + img {
+	opacity: 0.5;
 }
 
 /* Prevent delete for last adr/mail/tel item  */
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -310,7 +310,7 @@ avatar .icon-error {
 	width: 100%;
 	height: 100%
 }
-avatar .icon-error + img {
+avatar:not(.maximized) .icon-error + img {
 	opacity: 0.5;
 }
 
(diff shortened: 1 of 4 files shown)


skjnldsv:
@jancborchardt that's all we've been doing on the contact app lately when we encounter an invalid vcard.
We fix it but we don't save the card until the user manually check it. We made the choice to not automatically fix the errors with @irgendwie 

So the photo has been potentially fixed, but we still need the user approval :)
Do you want to remove the tooltip and leave the warning icon alone instead?

jancborchardt:
We fix it but we don't save the card until the user manually check it. We made the choice to not automatically fix the errors with @irgendwie 

Hm … for some cases it's definitely nicer to just automatically fix it and not require people to check every single contact. Otherwise it seems like it's always very buggy and if that happens with half of your 500 contacts (or more) it's really annoying to check. :\

skjnldsv:
@jancborchardt totally agree, some process are automatically processed like #137.

For now we need to test correctly our fix and trust them to allow automatic processing.
For now I say we go for semi-auto fix and if we don't have issues about users manually fixing their contacts, we should be able to go full-auto 🔫 ! 😆

skjnldsv:
@jancborchardt removed tooltip, please review!

irgendwie:
As a side note, @jancborchardt I want to avoid the situation, that the contacts app just magically 🦄  fixes vcards without any notice to the user and without any possibility to revert those changes.
I can see a lot of situation where this could lead to data loss and/or changes, which makes the data somehow unusable.

When the user has to manually check the vcards and the cards aren't changed automatically, we give the control back to the user to approve those changes or leave the vcards untouched.

skjnldsv:
Yes, I removed it, too much info for lambda user :)

jancborchardt:
@irgendwie I understand. However I very often had the case that half of my contacts have a warning icon next to them. That makes the trust in an app crumble very fast. Fixing automatically is way better than that. :)

irgendwie:
Also understandable - then maybe we can log the changes somewhere and make it possible to revert them (, also maybe at a later point then).

PR #288, opened 2017-08-15
Addressbook deletion update contacts list and groups

### LINKED ISSUE #287: Deleting addressbook don't update the contact list
A reload is required

skjnldsv:
Fix #287 

- [x] Working if the selected contact doesn't belong to the ab we deleted
- [x] Fix: update uid in url after deletion
- [x] Fix: too many request in background on contact deletion
- [x] Fix: update groups as well
- [x] Fix: lazy loading getContacts addressbooks

**Testing**:
 1. Add contacts to various addressbooks
 2. Delete one addressbook
 3. See everything being updated to left only the existing contacts

### DIFF
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -56,41 +56,45 @@ angular.module('contactsApp')
 	ctrl.loading = true;
 
 	ContactService.registerObserverCallback(function(ev) {
-		$timeout(function() { $scope.$apply(function() {
-			if (ev.event === 'delete') {
-				ctrl.getFirstContact(ev.uid);
-			}
-			else if (ev.event === 'create') {
-				$route.updateParams({
-					gid: $routeParams.gid,
-					uid: ev.uid
-				});
-			}
-			else if (ev.event === 'importend') {
-				$route.updateParams({
-					gid: t('contacts', 'All contacts')
-				});
-			}
-			if(ev.contacts.length !== 0) {
-				ctrl.contacts = ev.contacts;
-			}
-		}); });
+		$timeout(function() {
+			$scope.$apply(function() {
+				if (ev.event === 'delete') {
+					ctrl.getFirstContact(ev.uid);
+				}
+				else if (ev.event === 'create') {
+					$route.updateParams({
+						gid: $routeParams.gid,
+						uid: ev.uid
+					});
+				}
+				else if (ev.event === 'importend') {
+					$route.updateParams({
+						gid: t('contacts', 'All contacts')
+					});
+				}
+				if(ev.contacts.length !== 0) {
+					ctrl.contacts = ev.contacts;
+				}
+			});
+		});
 	});
 
 	AddressBookService.registerObserverCallback(function(ev) {
-		$timeout(function() { $scope.$apply(function() {
-			if (ev.event === 'delete') {
-				// Get contacts
-				ctrl.loading = true;
-				ContactService.updateDeletedAddressbook(function() {
-					ContactService.getAll().then(function(contacts) {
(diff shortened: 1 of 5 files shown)


js/components/contactList/contactList_controller.js
```diff
@@ -92,6 +77,22 @@ angular.module('contactsApp')
 		}); });
 	});
 
+	AddressBookService.registerObserverCallback(function(ev) {
+		$timeout(function() { $scope.$apply(function() {
```
ChristophWurst: I'd suggest to wrap the line after the curly brace

js/components/contactList/contactList_controller.js
```diff
@@ -235,4 +236,24 @@ angular.module('contactsApp')
 		return $routeParams.uid;
 	};
 
+	ctrl.getFirstContact = function(contactId) {
```
ChristophWurst: the method's name is a bit misleading. From the name I would expect it to return a contact, but apparently it doesn't really return anything but change another state. For the sake of readability I'd suggest to rename this accordingly, maybe to something like `selectFirstContact`? :)

skjnldsv:
@ChristophWurst thanks!

skjnldsv:
@irgendwie @Henni 🔔 😉

daniellandau:
Code looks fine, but couldn't really test it as I couldn't figure out address books at all or they were really buggy (rifling through three contacts in three different address books, the address book field in the detail view doesn't update correctly to reflect the address book that contact is in for example).

skjnldsv:
@daniellandau you mean the dropdown?
Should be another issue then.

Aside from that, anything else?
What do you mean by 'buggy'? :)

daniellandau:
Well, mainly that. I don't know if the problem is only with the dropdown, or if state is getting clobbered too. If I do a hard refresh and then immediately (without clicking through to different contacts) remove the address book of the currently selected contact, it gets removed from the list and focus moves to the nearest contact (yay, that's correct!).

If on the other hand I click a bunch of contacts and the dropdowns get out of sync the story is different. For example I have two contacts, `contact`, and `test` in address books `contacts` and `test1`. I click around in them until both contacts show `contacts` as the address book. Then I remove `test1` address book with `test1` contact selected. Focus moves correctly to `contact`, but `test1` contact is not removed from the list. Clicking on `test1` gives "No contacts in here" in the detail view.

Now that I tested some more, it seems that the "select nearest" functionality works as intended, but `this.updateDeletedAddressbook` has some issues (quite possibly not related to this PR, I don't know).

skjnldsv:
@daniellandau I don't understand what you say by "the dropdown gets out of sync", could you do a screencast or shot of what you encounter please?

daniellandau:
peek 2017-08-24 12-34

skjnldsv:
@daniellandau this is coming from your lazy loading I think:
https://github.com/nextcloud/contacts/commit/028df62242f9756c043af4faaff0f4359baaf892#diff-c61a7fd614f46bffdec17a09aaa20fa6R28

skjnldsv:
Yep, that was it, I'm not done yet, but pretty close. 
It will greatly increase the app load too since the defective function was requesting ALL addressbooks the first 20 contacts on load. Even if the contact wasn't in the request addressbook. That was resulting on some mixed-up addressbooks between contacts since all of the contacts were asigned to the last requested addressbook.

So we basically reduced from addressbookCount*20 requests to 20 requests! 🙌

skjnldsv:
Allclear!
@irgendwie @ChristophWurst @daniellandau @Henni @jonatoni @xh3n1 please review :)

irgendwie:
LGTM :+1:

PR #299, opened 2017-08-25
Read-only addressbook improvements 

### LINKED ISSUE #294: Create a read-only view of a contact
... for read-only shared addressbooks

Bugs fixed by this:
- Contacts edition is possible on some fields 
- No move to r/o addressbook allowed

skjnldsv:
- [x] Remove addressbook from the move select
- [x] Remove addressbook from the import select
- [x] Disable import if the only addressbook available is in read-only mode
- [x] Create read-only view for contacts in read-only addressbooks, fix #294

### DIFF
--- a/css/public/style.scss
+++ b/css/public/style.scss
@@ -220,7 +220,23 @@ avatar.maximized .avatar-options > [class^="icon-"] {
 }
 
 /* General details item styles */
-
+contactdetails {
+	select, button, input, textarea {
+		&:disabled {
+			background-color: transparent !important;
+			border-color: transparent !important;
+			opacity: 1 !important;
+			color: #545454 !important;
+		}
+	}
+	detailsitem > select:disabled {
+		background-image: none !important;
+		&:first-child {
+			opacity: 0.5 !important;
+			color: inherit !important;
+		}
+	}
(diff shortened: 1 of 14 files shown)


skjnldsv:
This is done on purpose since the read only details page for those type of user will completely remove this dropdown :)

kazam_screenshot_00000

skjnldsv:
Hum, 1d1ccdaee724091f5c1000ded07f5dafd20028f6 broke the import

skjnldsv:
Added readOnly details
contacts nextcloud 1
contacts nextcloud

Henni:
I'm a bit late to the party, but why did you create a separate readonly template for every existing template?
Couldn't we simply integrate this into the existing templates?

Something like:
```html
<input [...] ng-disabled="ctrl.addressBook.readOnly">
```

skjnldsv:
Tried that already, didn't liked the idea since a user can manually edit the field to remove the disabled state. But yes, I can push those commit if you prefer. I kind of like not having two templates too. :)

skjnldsv:
@Henni @irgendwie Done! Switch to previous commits.
Works fine too now! :)

PR #318, opened 2017-09-03
Fixes hidden fields taking up space

Henni:
Before:
image

After:
image

~~Still broken for small screens. I'm on it.~~


Henni:
Should work now.
This looks like it would be worth to give the detailitems some proper structure (css-classes, etc). I might take a look at this in a separate PR.

skjnldsv:
capture d ecran_2017-09-06_15-56-50

PR #320, opened 2017-09-10
Set FN as ORG, if ORG is present but neither FN nor N is given.

### LINKED ISSUE #241: Imported vCard Org contacts are all displayed as "New Contact"
### Steps to reproduce
1. Import vCards that include Organisations (so no given/surname)

### Expected behaviour
Display name/title should be the organisation name (probably "FN" property?)

### Actual behaviour
"FN" property is set to "New contact", so all organisational contacts are titled "New contact". Note that vCard (exported from G Suite/Google) have FN: field, but it is empty, so "New contact" is being explicitly generated and populated in the cards_properties table. "N" field is also a sequence of four semicolons, since no personal names are present in the export.

**Nextcloud version: 12

**Contacts version: 1.5.3

irgendwie:
Fixes #241 by setting FN to ORG, if neither FN nor N present (and ORG is present).

Some remarks:
- Currently if FN isn't present but N is, then the FN property doesn't get set
- We need to come up with a better way of name handling in general --> how do we handle updates to FN and N properties?

Edit: I'd also like to handle the FN/N/ORG stuff in a nicer way -> should we integrate the change in the contact_model?

### DIFF
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -420,7 +420,7 @@ angular.module('contactsApp')
 		} else {
 			angular.extend(this.props, {
 				version: [{value: '3.0'}],
-				fn: [{value: ''}]
+				fn: [{value: t('contacts', 'New contact')}]
 			});
 			this.data.addressData = $filter('JSON2vCard')(this.props);
 		}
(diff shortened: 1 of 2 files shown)


skjnldsv:
Are you stuck on this @irgendwie ? :)

PR #334, opened 2017-09-16
Notify on addressbook creation failure

### LINKED ISSUE #237: Ampersand (&) in Adressbook Name causes HTTP Error 400
> Migration and CardDAV issues belong in the server repo!

Filing this here because I think the error is coming from the contacts app. Looks like a missing excape in the Ajax request. 

### Steps to reproduce
1. Add new Adressbook with ampersand (&) in name  
2. Does not save, no visible feedback in the interface 

### Expected behaviour
It should work :) and it should report failure if something goes wrong, so that the user does need to open devtools. 

### Actual behaviour
Error 400 - this appears in devtools: 

```
MKCOL https://cloud.foo.bar/remote.php/dav/addressbooks/users/mt/a&b 400 ()
send @ dav.js?v=6904cf9…-41:3641
callee$2$0$ @ dav.js?v=6904cf9…-41:3156
tryCatch @ dav.js?v=6904cf9…-41:121
invoke @ dav.js?v=6904cf9…-41:294
Gp.(anonymous function) @ dav.js?v=6904cf9…-41:330
onFulfilled @ dav.js?v=6904cf9…-41:3737
(anonymous) @ dav.js?v=6904cf9…-41:3726
co @ dav.js?v=6904cf9…-41:3722
send @ dav.js?v=6904cf9…-41:3134
createCollection @ dav.js?v=6904cf9…-41:3453
createAddressBook @ dav.js?v=6904cf9…-41:1696
createAddressBook @ dav.js?v=6904cf9…-41:1492
(anonymous) @ addressBook_service.js:58
dav.js?v=6904cf9…-41:3649 Uncaught (in promise) Error: Bad status: 400
    at XMLHttpRequest.request.onreadystatechange (dav.js?v=6904cf9…-41:3649)
```

Server logfile: 
```json
{
  "reqId": "k4P13g7cmlCRSrWPCRVj",
  "level": 4,
  "time": "2017-06-20T22:01:17+00:00",
  "remoteAddr": "xxx",
  "user": "mt",
  "app": "webdav",
  "method": "MKCOL",
  "url": "/remote.php/dav/addressbooks/users/mt/A&B",
  "message": "Exception: {\"Message\":\"HTTP\\/1.1 400 EntityRef: expecting ';'\\n on line 9, column 37\",\"Exception\":\"Sabre\\\\DAV\\\\Exception\\\\BadRequest\",\"Code\":0,\"Trace\":\"#0 [intern
al function]: Sabre\\\\DAV\\\\CorePlugin->httpMkcol(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#1 \\/var\\/www\\/nextcloud\\/3rdparty\\/sabre\\/event\\/lib\\/EventEmitte
rTrait.php(105): call_user_func_array(Array, Array)\\n#2 \\/var\\/www\\/nextcloud\\/3rdparty\\/sabre\\/dav\\/lib\\/DAV\\/Server.php(479): Sabre\\\\Event\\\\EventEmitter->emit('method:MKCOL', Array)
\\n#3 \\/var\\/www\\/nextcloud\\/3rdparty\\/sabre\\/dav\\/lib\\/DAV\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#4
 \\/var\\/www\\/nextcloud\\/apps\\/dav\\/lib\\/Server.php(253): Sabre\\\\DAV\\\\Server->exec()\\n#5 \\/var\\/www\\/nextcloud\\/apps\\/dav\\/appinfo\\/v2\\/remote.php(33): OCA\\\\DAV\\\\Server->exec
()\\n#6 \\/var\\/www\\/nextcloud\\/remote.php(162): require_once('\\/var\\/www\\/nextcl...')\\n#7 {main}\",\"File\":\"\\/var\\/www\\/nextcloud\\/3rdparty\\/sabre\\/dav\\/lib\\/DAV\\/CorePlugin.php\
",\"Line\":570,\"User\":\"foo@bar\"}",
  "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3071.104 Safari/537.36",
  "version": "12.0.0.29"
}
```
### Server configuration
Nextcloud 12 - current Contacts app (1.5.3), PHP 7.1, nginx, PHP-FPM 

### Client configuration
**Browser:**
Chrome 59, Win10 

Looks like it's just a missing escape that passes the ampersand directly to the DAV and this fails. Other apps like Calendar or Groups work fine with the Ampersand. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46419869-ampersand-in-adressbook-name-causes-http-error-400?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
capture d ecran_2017-09-16_18-01-20

Fix #237

### DIFF
--- a/js/components/addressBookList/addressBookList_controller.js
+++ b/js/components/addressBookList/addressBookList_controller.js
@@ -29,6 +29,8 @@ angular.module('contactsApp')
 					ctrl.addressBooks.push(addressBook);
 					$scope.$apply();
 				});
+			}).catch(function() {
+				OC.Notification.showTemporary(t('contacts', 'AddressBook could not be created.'));
 			});
 		}
 	};
--- a/js/components/addressBookList/addressBookList_controller.js
+++ b/js/components/addressBookList/addressBookList_controller.js
@@ -21,7 +21,7 @@ angular.module('contactsApp')
 
 	ctrl.t = {
 		addressBookName : t('contacts', 'Address book name'),
-		regexError : t('contacts', 'Error, special characters allowed: -_.!?#|()')
+		regexError : t('contacts', 'Only these special characters are allowed: -_.!?#|()')
 	};
 
 	ctrl.createAddressBook = function() {
@@ -32,7 +32,7 @@ angular.module('contactsApp')
 					$scope.$apply();
 				});
 			}).catch(function() {
-				OC.Notification.showTemporary(t('contacts', 'AddressBook could not be created.'));
+				OC.Notification.showTemporary(t('contacts', 'Address book could not be created.'));
 			});
 		}
 	};
--- a/js/components/addressBookList/addressBookList_controller.js
+++ b/js/components/addressBookList/addressBookList_controller.js
@@ -20,7 +20,8 @@ angular.module('contactsApp')
 	});
 
 	ctrl.t = {
-		addressBookName : t('contacts', 'Address book name')
+		addressBookName : t('contacts', 'Address book name'),
+		regexError : t('contacts', 'Error, special characters allowed: -_.!?#|()')
 	};
 
 	ctrl.createAddressBook = function() {
--- a/js/components/addressBookList/addressBookList_controller.js
+++ b/js/components/addressBookList/addressBookList_controller.js
@@ -4,6 +4,7 @@ angular.module('contactsApp')
 
 	ctrl.loading = true;
 	ctrl.openedMenu = false;
+	ctrl.addressBookRegex = /^[a-zA-Z0-9À-ÿ\s-_.!?#|()]+$/i;
 
 	AddressBookService.getAll().then(function(addressBooks) {
 		ctrl.addressBooks = addressBooks;
--- a/css/_settings.scss
+++ b/css/_settings.scss
@@ -3,8 +3,8 @@
 }
 
 .addressBookList form {
-	position: relative;
 	width: 100%;
+	display: inline-flex;
 }
 
 .addressBookList form input.ng-invalid-pattern {
@@ -16,6 +16,14 @@ ul.addressBookList > li {
 	display: flex;
 	align-items: center;
 	flex-wrap: wrap;
+}
+
+ul.addressBookList > li.newAddressBookContainer {
+	position: initial !important;
+}
+
+ul.addressBookList > li.newAddressBookContainer .tooltip {
+	width: 175px;
 }
 
 ul.addressBook-share-list {
@@ -68,8 +76,7 @@ ul.addressBookList li .action > a {
 addressBookList input[type='submit'].inline-button,
 addressBookList input[type='button'].inline-button {
 	position: absolute;
-	right: 0;
-	top: 0;
+	right: 14px;
 	padding: 6px 15px;
 	background-color: transparent;
 	border: none;
@@ -119,6 +126,12 @@ div.addressBookShares ul.dropdown-menu li > a {
 	height: 30px !important;
 	min-height: 30px !important;
 	line-height: 30px !important;
+}
+
+ul.dropdown-menu li {
+	width: 100%;
+	padding: 3px 7px 4px;
+	cursor: pointer;
 }
 
 ul.dropdown-menu li.active {
--- a/css/_settings.scss
+++ b/css/_settings.scss
@@ -7,11 +7,8 @@
 	width: 100%;
 }
 
-ul.addressBookList > li {
-	padding: 6px 0;
-	display: flex;
-	align-items: center;
-	flex-wrap: wrap;
+.addressBookList form input.ng-invalid-pattern {
+	border-color: $color-error !important;
 }
 
 ul.addressBookList > li {
--- a/templates/addressBookList.html
+++ b/templates/addressBookList.html
@@ -1,13 +1,13 @@
 <i ng-show="ctrl.loading" style="display: block" class="icon-loading-small"></i>
 <ul class="addressBookList">
 	<li ng-repeat="addressBook in ctrl.addressBooks" addressbook data="addressBook" list="ctrl.addressBooks"></li>
-	<li>
+	<li class="newAddressBookContainer">
 		<form ng-submit="ctrl.createAddressBook()" name="newAddressBookForm">
 			<input id="newList" placeholder="{{ctrl.t.addressBookName}}" class="newAddressBookInput"
 				   ng-model="ctrl.newAddressBookName" type="text" ng-minlength="1"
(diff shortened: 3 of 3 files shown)


jancborchardt:
We should directly show an inline tooltip for the field when an unallowed character is used in the address book name. :)

skjnldsv:
@jancborchardt indeed!!

skjnldsv:
Okay, There is faaaar more characters that break things.
Like if you put a smiley, this gets inserted in the db, but you can't delete it after. ⚠️ 
Or you're allowed to create an ab with an hashtag in it, but this break the contact creation in it once in a while...

We could use a regex that matches all alphabet characters including Japanese, Chinese, Norwegian, Turkish... but this require a special libs as javascript regex does NOT support the magic selector: `\p{L}`
We could also only allow European char, but this seems like a drastic change, and I'm not quite fond of it.

@nextcloud/contacts, what do you think?

jancborchardt:
I'd say simply only allow a-z, A-Z, 0-9 and -, maybe space? It's just an address book name, and if it can break things, let's rather keep it safe.

mburnicki:
@jancborchardt Isn't contacts just a "frontend" to nextcloud's caldyv service? So IMO the rule for valid characters for the name of an addressbook should be in nextcloud, and the contacts app should just apply to that rule.

Alternatively, a addressbook could be created with a different tool like the cardbook addon for Thunderbird, and the contact app should be able to handle this correctly.

jancborchardt:
@mburnicki that's true, however the technical details will have to be weighed by @skjnldsv @irgendwie @Henni :)

skjnldsv:
We should sanitize the server with this.

skjnldsv:
I added some pattern and feedback if the name isn't accepted..
The pattern include some accented char, and some safe punctuation like `-_.!?#|()`. That should covers our needs.

irgendwie:
Don't forget, that there is a difference between the address book represented by a collection "object" and the name of an address book. The collection object is restricted in the terms of usable characters, but the displayname of the address book isn't or shouldn't afaik.
Optimally we would take the address book name the user entered when creating an address book, adapt it such that it is as close to the original name as possible and replace characters which aren't allowed. Therefore the collection object can be created and we can "instantly" rename the address book to the user desired value.

skjnldsv:
@irgendwie A lot of special char are breaking our dav server, so as far as I'm concerned, I think we should definitely prevent sending wrong data to it :)

irgendwie:
@skjnldsv I'm on your side :wink:
What I meant is, we can make it easier for the user by not throwing an error message, but instead decide for an (random) collection name (or try to take all the "legal" characters of the chosen name) and create the address book with that name, while setting the "displayname" of that address book to the original user input.

irgendwie:
The error message is fine for now but for the future we should increase usability by generating valid collection names for any user input leading to no errors and therefore no error message! 😊

skjnldsv:
I'm not sure to understand. You want us to have a 'storename' which include simple char and the displayname that the user as chosen?

If so, this could be a good idea, but how do we store the array to match the collectionNames with the dislayNames?

irgendwie:
See this example: https://tools.ietf.org/html/rfc5689#section-3.4

An address book is created on the server under `/home/special/` while the displayname is set to `<D:displayname>Special Resource</D:displayname>`.

Currently we are directly using the user input from the new address book field as collection and displayname.

skjnldsv:
Oooooooooooh!! 🙌 
This is an amazing idea! I never thought of that!!! Care to code it??
Amazing catch!! Far better that my approach! Well done!!!! ❤️

irgendwie:
We'll do this for the next release, since we need to come up with a way of creating collection names. For now let's use your PR! Would you mind adding a tooltip, explaining the allowed character for an address book? The red border is good but I still have no idea what I am allowed to input! 😀

skjnldsv:
Maybe angular provides a way to indicate which characters is breaking the input. Let me take a look! :)

irgendwie:
@skjnldsv Any news? Would like to integrate this into the 2.0.0 release! 😊

skjnldsv:
Wait for a small update. Will try to add a pop-up. When would you like to release? :)

irgendwie:
I'm currently looking for the last 'papercuts' and this is more or less the last PR I'd like to merge into this release.

skjnldsv:
@irgendwie I should be at my flat in like 30 min. I will finish this pr asap. :)

skjnldsv:
@irgendwie DONE!
I'm not quite sure about the error message. If you have something better, please edit! 😉

skjnldsv:
Suggestion:
capture d ecran_2017-09-24_19-55-41

jancborchardt:
@skjnldsv inline errors like that should be displayed as tooltip so the whole content below doesn't shift cause of a newly appeared div. :)

skjnldsv:
@jancborchardt Rolled back the old div commit.
- [ ] TOFIX:  Tooltip doesn't disappear when tooltip is shown and settings gets closed

skjnldsv:
capture d ecran_2017-09-25_18-49-52
All clear! @jancborchardt @irgendwie 🚂 🚂

jancborchardt:
@skjnldsv only one thing: Let’s put the tooltip above the box because then it’s also gonna work on mobile and be more visible and related to the field. :)

(Btw, I know dark mode needs to invert some colors – but a _white_ shadow? ;)

skjnldsv:
@jancborchardt done!
capture d ecran_2017-09-26_15-31-12

js/components/addressBookList/addressBookList_controller.js
```diff
@@ -21,7 +21,7 @@ angular.module('contactsApp')
 
 	ctrl.t = {
 		addressBookName : t('contacts', 'Address book name'),
-		regexError : t('contacts', 'Error, special characters allowed: -_.!?#|()')
+		regexError : t('contacts', 'Only these special characters are allowed: -_.!?#|()')
```
skjnldsv: Thx, far better! :)

jancborchardt:
@skjnldsv ;D

skjnldsv:
Where can I buy one!?

PR #342, opened 2017-09-20
Remove italic style from import feedback

jancborchardt:
We never ever use italic in any interface. :) The contact names shown on import progress were italic, quick fix & review @nextcloud/contacts @nextcloud/designers


skjnldsv:
Don't know why I approved twice :p

jancborchardt:
:D didn't know that is possible

PR #347, opened 2017-09-20
Only show delete icon on hover or focus

jancborchardt:
This cleans up the interface and doesn’t make the rarely used delete function that distracting. Similar to how we do it in the theming app.

~It doesn’t work for the address set of inputs yet because the delete icon is not next to the inputs.~ FIXED

### DIFF
--- a/css/_details.scss
+++ b/css/_details.scss
@@ -309,7 +309,23 @@ detailsitem .icon-delete {
 	padding: 16px 10px;
 	background-color: transparent;
 	border: none;
+	opacity: 0;
+}
+
+detailsitem input:hover + .icon-delete,
+detailsitem input:focus + .icon-delete,
+detailsitem input:active + .icon-delete,
+detailsitem select:hover + .icon-delete,
+detailsitem select:focus + .icon-delete,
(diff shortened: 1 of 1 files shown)


css/_details.scss
```diff
@@ -310,6 +310,13 @@ detailsitem .icon-delete {
 	background-color: transparent;
 	border: none;
 	opacity: .2;
+	visibility: hidden;
+}
+
+detailsitem input:hover + .icon-delete,
```
skjnldsv: Don't you want to use the hover state on the detailsitem element directly?
This would be a solution for the address issue :)

jancborchardt: Good enhancement, however there's no :active state for detailsitem, so it won't show when you are in a field of the address?

skjnldsv: Good call! :/
But I really would like to avoid js!

jancborchardt: Alright, I fixed it, check it out! :) Is also now accessible via keyboard cause I switched from `visibility: hidden;` to `opacity: 0;` :rocket:

PR #348, opened 2017-09-20
Only show input border on hover/focus

jancborchardt:
Also cleaning the interface a bit. It looks as simple as an interface which only show data, but it’s also super quick to edit.

Before & after:
screenshot from 2017-09-21 01-39-33
screenshot from 2017-09-21 01-35-47

Please review @nextcloud/contacts

### DIFF
--- a/css/_details.scss
+++ b/css/_details.scss
@@ -265,6 +265,13 @@ detailsitem textarea,
 .add-field {
 	width: 245px;
 	flex-grow: 1;
+	border-color: transparent;
+
+	&:hover,
+	&:focus,
+	&:active {
+		border-color: $color-primary;
+	}
 }
 
 .add-field {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Yeah, this is also the change I'm most skeptical about. :)

skjnldsv:
Lower the contrast of the border? :)
We can use the CSS selector to detect if the input is empty?

jancborchardt:
> Lower the contrast of the border? :)

Right, will play around with that.

> We can use the CSS selector to detect if the input is empty?

Good point – but for which fields do we then hide the border, empty or filled ones?

skjnldsv:
> Good point – but for which fields do we then hide the border, empty or filled ones?

I suggest to hide it for filled ones as the text is already an indicator! :)

Jab2870:
This is obviously a personal preference but I much prefer the original - with the borders showing all the time.

I have noticed a lot of forms recently doing away with the borders and, personally, I find them much harder to use. I would be surprised if I am the only person who feels this way

irgendwie:
If you want to reduce the clutter remove every box and show an underline or only lower border on mouse over. But don't completely remove every hint of the fields being editable.
For mobile always show the lower border/underline since we don't have a mouse pointer.

jancborchardt:
I’ll just close this – was an experiment anyway. ;))

PR #349, opened 2017-09-21
Add loading in details view on contact switch and initialisation 

skjnldsv:
kazam

### DIFF
--- a/js/components/contactDetails/contactDetails_controller.js
+++ b/js/components/contactDetails/contactDetails_controller.js
@@ -22,12 +22,12 @@ angular.module('contactsApp')
 		placeholderName : t('contacts', 'Name'),
 		placeholderOrg : t('contacts', 'Organization'),
 		placeholderTitle : t('contacts', 'Title'),
-		selectField : t('contacts', 'Add field ...'),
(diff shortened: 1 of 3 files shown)


js/components/contactDetails/contactDetails_controller.js
```diff
@@ -25,7 +26,8 @@ angular.module('contactsApp')
 		download : t('contacts', 'Download'),
 		delete : t('contacts', 'Delete'),
 		save : t('contacts', 'Save changes'),
-		addressBook : t('contacts', 'Address book')
+		addressBook : t('contacts', 'Address book'),
+		loading : t('contacts', 'Waiting for the contact list to load...')
```
jancborchardt: I’d say »Loading contacts« here, that doesn’t sound like we’re complaining about the slowness of our own app ;D

skjnldsv: Ahahha, good point!! 😂

jancborchardt: Just fixed it ;)

skjnldsv:
You pushed a second before me! 😉👍

PR #352, opened 2017-09-21
Always refresh contactList

irgendwie:
Deleting the last contact leads to the following issue:

2017-09-21-153550_825x147

@skjnldsv Is there a reason we only update the list if `length !== 0`?

### DIFF
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -72,9 +72,7 @@ angular.module('contactsApp')
 						gid: t('contacts', 'All contacts')
 					});
 				}
-				if(ev.contacts.length !== 0) {
-					ctrl.contactList = ev.contacts;
-				}
+				ctrl.contactList = ev.contacts;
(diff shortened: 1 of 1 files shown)


skjnldsv:
I added this line because of the importend broadcast. Please test the import function :)

irgendwie:
@skjnldsv import seems fine on my side 🙈

jonatoni:
Works even for me

skjnldsv:
Dang it! I knew something was wrong! ^^
Now you have an error when you upload and have no contacts.

irgendwie:
@skjnldsv Upload? You mean import with no contacts? No error for me

skjnldsv:
How it happens on my side:
1. Remove all contacts
2. Have 2 addressbooks
3. reload
4. select second ab
5. import single card contact
6. see error in console and app crash

irgendwie:
Ok, was able to reproduce this - will take a look!

PR #355, opened 2017-09-21
Fix alignment of header content to input fields

jancborchardt:
The text used to be left-aligned, must have gotten lost at some point. Fixed. :)

Before & after (disregard the invalid text in the input fields, was just to align it)
screenshot from 2017-09-21 16-59-34 screenshot from 2017-09-21 16-55-27

Please review @nextcloud/contacts @nextcloud/designers

### DIFF
--- a/css/_details.scss
+++ b/css/_details.scss
@@ -15,7 +15,7 @@
 
 .contactdetails__header {
 	height: 100px;
-	padding-left: 20px;
+	padding-left: 44px;
 	display: flex;
 	font-weight: bold;
 	align-items: center;


skjnldsv:
The left margin of avatar looks too big no?

jancborchardt:
@skjnldsv it looks that way in the screenshot, but actually it’s quite nice when in normal view. And as a bonus on mobile the menu icon doesn’t overlap the face anymore.

jancborchardt:
The avatar belongs to the name, so it should be closer to it than the left side.

skjnldsv:
Hum, I see your point! LGTM then! :)

PR #357, opened 2017-09-21
adjust the space between avatar and name in the contacts list

xh3n1:
Before:
before
After:
after

### DIFF
--- a/css/_app-content-list.scss
+++ b/css/_app-content-list.scss
@@ -31,6 +31,7 @@
 	height: 68px;
 	border-top: 1px solid nc-darken($color-main-background, 8%);
 	cursor: pointer;
+	padding-left: 7px;
 }
 
 .app-content-list-item:hover,


skjnldsv:
This is going into server though. :)

irgendwie:
This leads to the contactList being scroll-able in the horizontal direction (left and right) 🙈
My css pros, how do we fix this? @xh3n1 @skjnldsv @jancborchardt @jonatoni

xh3n1:
@irgendwie I don't think it induces the contactList to being scrollable in the horizontal direction :confused: 
image

irgendwie:
@xh3n1 will recheck

irgendwie:
@xh3n1 It only happens when the scrollbar is visible, meaning you need a bigger number of contacts, then it is scroll-able in the horizontal direction

jancborchardt:
@xh3n1 the element width of the name has to be adjusted. :) Maybe best to do this all in core as @skjnldsv mentioned.

irgendwie:
@xh3n1 Btw, today I found out this problem only occurs in chrome and *not* in firefox - you are using firefox, right?

xh3n1:
@irgendwie I tested it in Firefox and Chromium and this problem does not occur :/

irgendwie:
@jancborchardt @skjnldsv @jonatoni Are also not able to reproduce this issue?

PR #365, opened 2017-09-30
Add getFullContacts and update to ObserverCallback.

irgendwie:
The events getFullContacts and update need to update the ctrl.contactList in order to work correctly.

Before:
- contactList didn't react on contact updates
- contactList didn't react on changing sorting order

fixes #364 which was introduced by #359

PR #366, opened 2017-09-30
Select first contact after import if available

### LINKED ISSUE #343: Issues after import
Directly after import, there are 2 issues:
- The content view still shows »No contacts in here«. The first contact should simply be shown as usual
- Clicking any other group doesn’t bring up the contacts and just leaves the view white.

Can you confirm @nextcloud/contacts? I don’t get any console logs. Stuff works only after a hard refresh.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49558078-issues-after-import?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

irgendwie:
Should fix #343

- In callback, the contactList gets refreshed if 'importend' was notified, in order to select the first contact
- If import was successful we should have at least one contact, otherwise the routeParams UID is set to undefined
- 'importend' now only gets notified once - it was done twice before the change

Edit: May need #365 to correctly work!

### DIFF
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -56,6 +56,13 @@ angular.module('contactsApp')
 	ctrl.loading = true;
 
 	ContactService.registerObserverCallback(function(ev) {
+		/* after import at first refresh the contactList */
+		if (ev.event === 'importend') {
+			$scope.$apply(function() {
+				ctrl.contactList = ev.contacts;
+			});
+		}
+		/* update route parameters */
 		$timeout(function() {
 			$scope.$apply(function() {
 				switch(ev.event) {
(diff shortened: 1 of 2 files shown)


irgendwie:
@skjnldsv Any idea where this comes from?

2017-09-30-153731_1690x530

This happens when you try to import a file without any contacts.
(It was already there before this PR.)

skjnldsv:
@irgendwie I thought we fixed that! 🤔 

Basically we have some apply within another apply I think. This issue was already mentioned by me not long ago. I can't recall in what pr we had this discussion. :/

irgendwie:
@skjnldsv Single contact is also enough... looking into it

irgendwie:
@skjnldsv The bug description of you isn't introduced by this PR, but rather a regression of the lazy loading PR and the initial code of the contacts app (createAddressbook isn't working 100% correctly).
I'd like to leave the PR as it is and fix the problem you described in a different PR.

jancborchardt:
Thank you @irgendwie! :) :tada:

PR #408, opened 2017-10-26
Add group and contactfilter model, implement group ordering by count

### LINKED ISSUE #344: Sort order of the contact groups
Currently there seems to be no specific sort order of the contact groups in the navigation:
screenshot from 2017-09-21 00-58-05
It’s neither alphabetically nor by amount of people in it.

Ideal would be to sort the groups by number of people in it, biggest groups on top.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49558174-sort-order-of-the-contact-groups?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Henni:
Besides implementing sorting of groups (fixes #344 ), this also adds two models:
* Group Model, which replaces the array representation of groups:
`['name', 42]` -> `new Group({name: 'name', count: 42})`
* ContactFilter

'All Contacts' and 'Not grouped' are now Contact Filters instead of groups.
This will also allow us to add additional filters in the future, which are kept separate from the groups.

Filters are always shown above the groups.

### DIFF
--- a/js/components/contactFilter/contactFilter_controller.js
+++ b/js/components/contactFilter/contactFilter_controller.js
@@ -1,0 +1,5 @@
+angular.module('contactsApp')
+.controller('contactfilterCtrl', function() {
+	// eslint-disable-next-line no-unused-vars
+	var ctrl = this;
+});
--- a/js/components/contactFilter/contactFilter_directive.js
+++ b/js/components/contactFilter/contactFilter_directive.js
@@ -1,0 +1,13 @@
+angular.module('contactsApp')
+.directive('contactFilter', function() {
+	return {
+		restrict: 'A', // has to be an attribute to work with core css
+		scope: {},
+		controller: 'contactfilterCtrl',
+		controllerAs: 'ctrl',
+		bindToController: {
+			contactFilter: '=contactFilter'
+		},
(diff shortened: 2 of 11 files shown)


skjnldsv:
Not sure if relative to this pr, but creating a contact when a group is selected does not add it to the active group by default anymore! :/

Henni:
I believe this was broken before...

Contactfilters are  separate from groups.
First all contactfilters are displayed. Afterwards the groups are listed, ordered by count (desc).

skjnldsv:
Nice! Do you plan adding the different sorting options in this pr?

Henni:
@skjnldsv nope, that is stuff for a follow-up

skjnldsv:
Could you add tests for this in another pr? :)

DesktopMasters:
So.. it has been a few days.  Are we going to be able to sort the groups Alphabetically like a human would want to see them?

PR #409, opened 2017-10-26
Adding tests for models

Henni:
is there any other way to get coverage results, besides publishing a PR?


js/models/addressBook_model.js
```diff
@@ -8,16 +8,6 @@ angular.module('contactsApp')
...
 			readOnly: data.data.props.readOnly === '1',
-
-			getContact: function(uid) {
-				for(var i in this.contacts) {
-					if(this.contacts[i].uid() === uid) {
-						return this.contacts[i];
-					}
-				}
-				return undefined;
-			},
```
Henni: This code isn't used, so I removed it instead of adding tests for it.

PR #410, opened 2017-10-27
Exclude already shared groups from the sharees search

### LINKED ISSUE #140: Already shared groups aren't excluded from the search when sharing
And we can still add them multiple times
capture d ecran_2017-02-19_16-54-53

@irgendwie do you know if this was on purpose?
@Henni ?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/42202813-already-shared-groups-aren-t-excluded-from-the-search-when-sharing?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
Fix #140

### DIFF
--- a/js/components/addressBook/addressBook_controller.js
+++ b/js/components/addressBook/addressBook_controller.js
@@ -84,17 +84,18 @@ angular.module('contactsApp')
 				itemType: 'principals'
 			}
 		).then(function(result) {
-			// Todo - filter out current user, existing sharees
 			var users   = result.ocs.data.exact.users.concat(result.ocs.data.users);
 			var groups  = result.ocs.data.exact.groups.concat(result.ocs.data.groups);
 
 			var userShares = ctrl.addressBook.sharedWith.users;
 			var userSharesLength = userShares.length;
+
+			var groupsShares = ctrl.addressBook.sharedWith.groups;
+			var groupsSharesLength = groupsShares.length;
 			var i, j;
 
 			// Filter out current user
-			var usersLength = users.length;
-			for (i = 0 ; i < usersLength; i++) {
+			for (i = 0 ; i < users.length; i++) {
 				if (users[i].value.shareWith === OC.currentUser) {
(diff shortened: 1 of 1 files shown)


js/components/addressBook/addressBook_controller.js
```diff
@@ -103,16 +104,26 @@ angular.module('contactsApp')
 
 			// Now filter out all sharees that are already shared with
 			for (i = 0; i < userSharesLength; i++) {
-				var share = userShares[i];
-				usersLength = users.length;
-				for (j = 0; j < usersLength; j++) {
-					if (users[j].value.shareWith === share.id) {
+				var shareUser = userShares[i];
+				for (j = 0; j < users.length; j++) {
+					if (users[j].value.shareWith === shareUser.id) {
```
Henni: How about using `Array.filter` instead of the `for splice` construct to improve readability?

skjnldsv: We could indeed, but we need to create the filter function anyway, so it would basically be the same :)

skjnldsv: Also: basic loop seems more performant: https://jsperf.com/javascript-array-filter-versus-loop

Henni: Splice seems to defeat the performance improvement:
https://jsperf.com/splice-vs-filter

But let's keep it as it is for now.
We don't need to overengineer this.

skjnldsv: We could use an auto test of performance? Is such thing possible?

jancborchardt:
Sounds good! :+1:

PR #412, opened 2017-10-27
Enable/disable addressbooks

### LINKED ISSUE #36: Possibility to filter by account
It would be very useful to be able to only display contacts from a certain address book in Nextcloud's UI.

In our setup, each user has his own address book plus access to a organization-wide address book (shared from a user created just for sharing organization-wide filed, appointments, contacts, etc.).
One of the use cases is to review contacts from the personal address book and move them to the company-wide because mobile devices that cannot distinguish between the two address books will just put contacts created on the device into the personal address book.

Currently we do this work using Thunderbird's address book with the SoGo connector. Both personal and org-wide address books are set up and they can be accessed independently and contacts moved around.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39465792-possibility-to-filter-by-account?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
capture d ecran_2017-10-27_22-32-32

Fix #36


templates/contactList.html
```diff
@@ -1,6 +1,6 @@
 <div class="contacts-list" ng-class="{loading: ctrl.loading, 'mobile-show': ctrl.show}">
         <div class="app-content-list-item"
-			 ng-repeat="contact in ctrl.filteredContacts = (ctrl.contactList | contactGroupFilter:ctrl.routeParams.gid | localeOrderBy:ctrl.sortBy | filter:query | limitTo:ctrl.limitTo ) as filtered track by contact.uid()"
+			 ng-repeat="contact in ctrl.filteredContacts = (ctrl.contactList | filter:{'data':{'addressBook':{'enabled': true}}} | contactGroupFilter:ctrl.routeParams.gid | localeOrderBy:ctrl.sortBy | filter:query | limitTo:ctrl.limitTo) as filtered track by contact.uid()"
```
skjnldsv: For some reason it doesnt work! :/
@Henni ?

js/models/addressBook_model.js
```diff
@@ -2,12 +2,17 @@ angular.module('contactsApp')
 .factory('AddressBook', function()
 {
 	return function AddressBook(data) {
+		var storageKey = btoa(data.data.href.split('/').splice(-3, 2).join('_'));
```
Henni: This variable isn't used later on. Also stuff like this asks for a short comment.

templates/contactList.html
Henni: As far as I can see, you only store the state in localstorage and don't update it inside the addressbook itself.
This update is only applied on page reload (https://github.com/nextcloud/contacts/pull/412/files#diff-adfefcaa5c468a1b21ca840b249cadcfR14).

Why don't you store it on the server in the first place?

js/models/addressBook_model.js
```diff
@@ -2,12 +2,17 @@ angular.module('contactsApp')
...
 		angular.extend(this, {
 
 			displayName: '',
 			contacts: [],
 			groups: data.data.props.groups,
 			readOnly: data.data.props.readOnly === '1',
+			// use only owner and url, convert ''/remote.php/dav/addressbooks/users/admin/Test1/'' to 'admin.Test'
+			key: data.data.href.split('/').splice(-3, 2).join('_'),
+			enabled: window.localStorage.getItem('contacts_ab_'+this.key) !== false,
```
Henni: this doesn't work, as localstorage only stores Strings.
=> this statement always evaluates to true.

skjnldsv: Indeed thx

templates/contactList.html
skjnldsv: I don't think we can.

js/models/addressBook_model.js
skjnldsv: Hum, it's strange since it actually remembers the state, so It seems to work :)

js/models/addressBook_model.js
skjnldsv: Fixed

js/models/addressBook_model.js.orig
```diff
@@ -0,0 +1,72 @@
...
+					this.sharedWith.groups.push({
+						id: href.substr(28),
+						displayname: href.substr(28),
+						writable: readWrite
+					});
+				}
+			}
+		}
+	};
+});
```
Henni: this file shouldn't be in here ;)

templates/contactList.html.orig
```diff
@@ -0,0 +1,18 @@
...
+			 ng-click="setSelected(contact.uid())"
+			 ng-class="{active: contact.uid() === ctrl.getSelectedId()}">
+	</div>
+	<div ng-show="!ctrl.filteredContacts.length && !ctrl.loading && ctrl.searchTerm !== ''">
+	        <div id="emptycontent" class="emptycontent-search">
+		        <div class="icon-search"></div>
+			<h2>{{ctrl.t.emptySearch}}</h2>
+		</div>
+	</div>
+</div>
```
Henni: this as well

templates/contactList.html
Henni: Should be possible, calendar does it. But they use a property in the 'owncloud' namespace.
```
<d:propertyupdate xmlns:d="DAV:"><d:set><d:prop><o:calendar-enabled xmlns:o="http://owncloud.org/ns">0</o:calendar-enabled></d:prop></d:set></d:propertyupdate>
```
@irgendwie Is such a thing possible with our current dav lib?

js/components/addressBook/addressBook_controller.js
```diff
@@ -178,4 +181,9 @@ angular.module('contactsApp')
 		});
 	};
 
+	ctrl.toggleState = function() {
+		ctrl.enabled = AddressBookService.toggleState(ctrl.addressBook);
+		ContactService.fillCache();
```
Henni: Why do you call `fillCache`?

skjnldsv: It was to force a full refresh of the cache. This pr is still wip, I'm not convinced yet! :)

templates/contactList.html
skjnldsv: Oooh yeah, I remember this! I don't see where it's stored in the db though! :/

skjnldsv: IIRC, the settings weren't saved.

skjnldsv: Okay, I needed a patch for the dav lib.
Easy fix now! :)

All good :)

skjnldsv:
Bump! :)
@nextcloud/contacts

skjnldsv:
Bump! :)
@nextcloud/contacts

Henni:
Works mostly fine.

For some reason the currently selected contact is changed when enabling/disabling addressbooks:
peek 2017-11-13 13-42

Also moving contacts seems to have a few issues at the moment:
* Moving a contact to an addressbook and disabling & enabling this addressbook hides the contact but doesn't reshow it.
* Also moving contacts to a different addressbook breaks opening them in the details view.

Both of these issues are fixed after reloading the page.

@skjnldsv Can you reproduce this?

skjnldsv:
Let me fix this, and then publish a new version! :)

skjnldsv:
@Henni Should be fixed :)

skjnldsv:
- [ ] Enabling ab is re-adding the same contacts inside the original ab

skjnldsv:
God damnit, Every time I fix a bug, I find a new unrelated one!! 😭 
@henni could you take a look at https://github.com/nextcloud/contacts/pull/412/files#diff-c61a7fd614f46bffdec17a09aaa20fa6R391 please? I can't figure why the cache doesn't get updated! :/

skjnldsv:
- [ ] Disabling still select nearest contact

skjnldsv:
Pfiouuu! Rebased and ready to merge!
@nextcloud/contacts @jancborchardt @MorrisJobke

skjnldsv:
Bump?

private-nc-user:
I just noticed, that i can't disable a shared adressbook. Since I'm a neewbie can you just give me an advice how to open an issue the right way. Or is just writing this comment enough?

timreeves:
I have also just noticed the same problem as @private-nc-user . It goes like this: 
- Request = https://owncloud.<hidden>/remote.php/dav/addressbooks/users/<hidden>/gabriele_aktiv_shared_by_Gabriele/ (Method = PROPPATCH)
- Response = 403 Forbidden

skjnldsv:
Yes, you can't disable a shared addressbook as the disabled permission is for the owner of this addressbook :/

timreeves:
Thanks for the explanation John @skjnldsv. It's just not what we wanted to hear )-: Look at it from the user pov, not the core hacker pov: The address book shared with me by someone else is more likely the one I want to be able to blend out the most, as it does not contain the contacts which I myself have gathered.
I understand that the backend features you build on do not allow to do what you want; but in that case, either there is a feature lacking, or a layer of abstraction somewhere. Or the app needs to take more control itself and add the neccessary abstraction, or whatever.
Don't take this too bad - I really really do appreciate all the work contributors do, really - but right now I'm feeling like a guy who bought a car and then found the motor built in onto the passenger seat - sorry mate, it was the only place we could fit it in :)

mburnicki:
I also think that it would benice that if you "disable" a specific address book then the contacts from that address books are simply not **shown**. I'm currently maintaining about 30 address books that are shared with different groups of users, so it would be nice to "enable" just a single address book  to see only the contacts in that address book to maintain those contacts.

This is how e.g. the "cardbook" addon for Thunderbird
https://addons.mozilla.org/en-US/thunderbird/addon/cardbook/
works, and this is also possible in the "kontact" app that ships with KDE.

Also, a contact search should only be done in address books that are "enabled".

If you disable the address books on the server then other clients will also be unable to sync the contacts from the disabled address books. There may be cases where this is also desired, but usually you can even configure in Android's contact app which address books to synchronize, and addresses from which address books to be displayed in the contacts list.

timreeves:
Very lucid comment from @mburnicki - he says it better than I did. Seconded! What we need is a clear distinction between disabling an address book completely (I never imagined I might not be able to sync it from my Thunderbird if it is disabled in NC GUI display - igitt); and simply removing its data from the current GUI display settings.

I would personally not even mind if it were not displayed (or even prefer) but still searchable when I type in part of a name or email, e.g. in the Mail app.

What we need to remember is that for some users this hiding of address books in the NC GUI only ever came up as an issue because of the way the app is designed - to always fetch ALL the contacts data to the browser and then implement user display choices in JS. This gives you nifty features for privaze-size address books, but breaks down at around 1000 contacts, becomes too slow to use. So it is not workable for professional-size / team / company address books.

Thus I still advocate a choice of "front-end" OR "backend with *real* infinite scrolling" solutions to display and manage the contacts, [see this issue](https://help.nextcloud.com/t/handling-a-large-number-1000-of-contacts-in-the-web-ui/22064). Quoting from that: _What I really miss is the option to select WHICH of those three books are to be displayed in the GUI: I would only really need one in everyday use, the other two are my archive and my girlfriends book, both of which I only seldom need._

So imho it was never about disabling an address book completely (although that is also a nice feature), but rather about choosing which to display in the GUI. But a seldom-used address book, which I do not want shown in the GUI, should still be available in the background for syncing or more intensive use by others. Put simply, I just want to choose for myself not to see my girlfriends contacts - I could never tell her to disable her address book.

The very minimum we need here is a clear notice in the GUI that disabling an address book will also prevent it from being synced (if I have understood that right).

Anyway, thanks to all involved, as ever!

skjnldsv:
I'll see what I can do, it's maybe doable, I don't know. Can one of you guys open an issue about this? :)

Henni:
@timreeves @mburnicki thanks for the constructive comments. as @skjnldsv said, please open an issue about this and we see what we can do :)

skjnldsv:
> @timreeves @mburnicki thanks for the constructive comments

Yes, it's always a pleasure to have people like you being so nice and friendly about issues issues! :)

timreeves:
To make sure we don't open two: I'm starting on the new issue now.

mburnicki:
@timreeves Thanks. Yesterday I was too busy to do it.

PR #424, opened 2017-11-02
Fix click on popover

### LINKED ISSUE #420: Can't use the addressbook popovermenu on mobile
When using a mobile device, we can't use the popovermenu as it goes away half a sec before the user click.

skjnldsv:
Fix #420 
Please test on mobile.


skjnldsv:
@nextcloud/contacts Can one of you test on ios as well? :)

PR #425, opened 2017-11-02
Fixed contact update after propertyGroup refactor

skjnldsv:
Signed-off-by: John Molakvoæ <skjnldsv@protonmail.com>

### DIFF
--- a/templates/detailItems/adr.html
+++ b/templates/detailItems/adr.html
@@ -5,32 +5,32 @@
 <div>
 
 	<label for="details-pobox-{{ctrl.index}}">{{ctrl.t.poBox}}</label>
-	<input type="text" id="details-pobox-{{ctrl.index}}" name="address" ng-model="ctrl.data.value[0]" ng-model-options="{ debounce: 500 }" ng-change="ctrl.model.updateContact()"  value=""
+	<input type="text" id="details-pobox-{{ctrl.index}}" name="address" ng-model="ctrl.data.value[0]" ng-model-options="{ debounce: 500 }" ng-change="ctrl.updateContact()"  value=""
 	ng-disabled="ctrl.model.addressBook.readOnly" />
 </div>
 <div>
 	<label for="details-address-{{ctrl.index}}">{{ctrl.t.address}}</label>
-	<input type="text" id="details-address-{{ctrl.index}}" name="address" ng-model="ctrl.data.value[2]" ng-model-options="{ debounce: 500 }" ng-change="ctrl.model.updateContact()" value=""
+	<input type="text" id="details-address-{{ctrl.index}}" name="address" ng-model="ctrl.data.value[2]" ng-model-options="{ debounce: 500 }" ng-change="ctrl.updateContact()" value=""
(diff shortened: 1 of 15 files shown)


templates/contactDetails.html
```diff
@@ -42,7 +42,7 @@
 			</div>
 		</header>
 		<section>
-			<propertyGroup ng-repeat="prop in ctrl.contact.props | toArray | orderDetailItems:'$key'" model="ctrl.contact" data="prop" name="prop.$key"></propertyGroup>
+			<propertyGroup ng-repeat="prop in ctrl.contact.props | toArray | orderDetailItems:'$key'" model="ctrl.contact" addressBook="ctrl.addressBook" data="prop" name="prop.$key"></propertyGroup>
```
skjnldsv: @Henni this doesn't work and fails with a `Error: [$compile:multidir]`
Can you help?

Henni: Adding the attribute `addressBook` breaks this.
We have a directive which is called `addressBook`, so Angular doesn't know if it should now render the addressBook or the propertyGroup directive.
You have to choose a different attribute name.

js/components/detailsItem/detailsItem_directive.js
```diff
@@ -8,6 +8,7 @@ angular.module('contactsApp')
 			name: '=',
 			data: '=',
 			contact: '=model',
+			addressBook: '=addressBook',
```
Henni: Just FYI this line is to specific: `addressBook: '='` would suffice.
You only need to write something after the `=` if the attribute should be named differently to the controller variable.

templates/detailItems/adr.html
```diff
@@ -1,36 +1,36 @@
-<select ng-model="ctrl.type" ng-model-options="{ debounce: 500 }" ng-change="ctrl.changeType(ctrl.type)" ng-disabled="ctrl.model.addressBook.readOnly">
+<select ng-model="ctrl.type" ng-model-options="{ debounce: 500 }" ng-change="ctrl.changeType(ctrl.type)" ng-disabled="ctrl.addressBook.readOnly">
 	<option ng-repeat="option in ctrl.availableOptions" value="{{option.id}}">{{option.name}}</option>
 </select>
-<button ng-click="ctrl.deleteField()" class="icon-delete" title="{{ctrl.t.delete}}" ng-if="!ctrl.model.addressBook.readOnly"></button>
+<button ng-click="ctrl.deleteField()" class="icon-delete" title="{{ctrl.t.delete}}" ng-if="!ctrl.addressBook.readOnly"></button>
```
Henni: I would suggest that we offer a readOnly property for contacts, which reflects the readOnly property of the addressbook.
This way we also don't have to pass the addressbook to the propertyGroup

skjnldsv:
All clear! @irgendwie @Henni Please review! :)
cc @MorrisJobke @jancborchardt

MorrisJobke:
> Code changes seem reasonable, every field update gets pushed to the server and read-only view is also working again! 👍

Then merge it ;)

irgendwie:
@MorrisJobke Was waiting for another :+1: 😉

PR #460, opened 2018-01-22
Added the possibility to upload a Public GPG key.

tacruc:
This PR is depending on https://github.com/nextcloud/server/pull/7997. It addes the possibility to add public key or an link on a key to an vCard.

### DIFF
--- a/js/services/vCardProperties.js
+++ b/js/services/vCardProperties.js
@@ -100,6 +100,19 @@ angular.module('contactsApp')
 				{id: 'OTHER', name: t('contacts', 'Other')}
 			]
 		},
+		key: {
+			multiple: true,
+			readableName: t('contacts', 'Public key'),
+			template: 'key',
+			icon: 'icon-key',
+			defaultValue: {
+				value:'',
+				meta:{type:['PGP']}
+			},
+			options: [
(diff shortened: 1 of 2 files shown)


skjnldsv:
Hello there! :)
Sorry for the late feedback! Could you add some verification in this? Since we're suppose to accept a valid base64, It could be great to ensure that the text is always encoded.

Also, the key indicator should probably have a `data:application/pgp-keys;base64` attached for better handling by other apps :)
https://tools.ietf.org/html/rfc6350#section-6.8.1

tacruc:
@skjnldsv thanks for the feedback.

Which files would be the right place to add the base64 check? 
Should it just be added in templates/detailItems/key.html?

And for the key indication `data:application/pgp-keys;base64` while implementing, but somehow failed to encode it in the given strukture. 
So I lookt for alternatives and found this one in https://en.wikipedia.org/wiki/VCard

I but I see your point and will have a look into this, if it is clear what is going to happen to the dependency PR on the server side.

skjnldsv:
@tacruc sorry for the very late delay.
Because we planned to upgrade this app to vue we couldn't merge your pr.

We'll try to add this feature into 3.1.0 :)

PR #488, opened 2018-02-20
Removing default name for new contact and leaving it blank

rutujasurve94:
Removing the check that puts the default name 'New Contact' in case the contact field is new

### DIFF
--- a/js/services/contact_service.js
+++ b/js/services/contact_service.js
@@ -223,9 +223,6 @@ angular.module('contactsApp')
 		newContact.uid(newUid);
 		newContact.setUrl(addressBook, newUid);
 		newContact.addressBookId = addressBook.displayName;
-		if (_.isUndefined(newContact.fullName()) || newContact.fullName() === '') {
-			newContact.fullName(newContact.displayName());
-		}
 
 		return DavClient.createCard(
 			addressBook,


jancborchardt:
@rutujasurve94 welcome to the Contacts app! :tada: :)

@skjnldsv @Henni @juliushaertl this needs a Javascript review :)

Henni:
I believe this leads to issues as the full name has to be defined. (Possibly fixed by #490)

Henni:
Removing these lines is mostly fine, but we still have to catch imported contacts which do not have a FN field. So getting rid of the check for `''` is fine, but we still have to check for undefined.

But to achieve the goal of this PR, the contact_model.js has to be modified as well.
@rutujasurve94 take a look at `js/models/contact_model.js` (line 439). This is where the fn property is initialized for new contacts. This should be replaced with an empty string, so that the property still exists, but the input field is blank.

rutujasurve94:
@Henni thanks a lot for the review, I'll modify the code soon and update it here

rutujasurve94:
@Henni @jancborchardt I have modified the code in the file contact_model.js to substitute 'New Contact' with an empty string.

js/models/contact_model.js
```diff
@@ -436,7 +436,7 @@ angular.module('contactsApp')
 		} else {
 			angular.extend(this.props, {
 				version: [{value: '3.0'}],
-				fn: [{value: t('contacts', 'New contact')}]
+				fn: [{value: t('contacts', '')}]
```
Henni: The `t()` function translates a string to the local language. We don't need to translate `''`. So just replace `t('contacts', '')` with `''`.

js/services/contact_service.js
```diff
@@ -223,9 +223,6 @@ angular.module('contactsApp')
 		newContact.uid(newUid);
 		newContact.setUrl(addressBook, newUid);
 		newContact.addressBookId = addressBook.displayName;
-		if (_.isUndefined(newContact.fullName()) || newContact.fullName() === '') {
```
Henni: We still need to check if the fullName of newContact is undefined. So please add the first half of this check back in.

rutujasurve94:
@Henni @jancborchardt I have modified the code according to the review

jancborchardt:
@nextcloud/contacts need further review here :)

js/models/contact_model.js
```diff
@@ -436,7 +436,7 @@ angular.module('contactsApp')
 		} else {
 			angular.extend(this.props, {
 				version: [{value: '3.0'}],
-				fn: [{value: t('contacts', 'New contact')}]
+				fn: [{value: ''}]
```
skjnldsv: Server is still refusing the vcard

> Validation error in vCard: The FN property must appear in the VCARD component exactly 1 time

I'll suggest using a space here

PR #497, opened 2018-02-26
Removed unnecessary calc function

skjnldsv:


Henni:
Do you want to continue over [here](https://github.com/nextcloud/contacts/pull/497/files#diff-0e65f3f3b823d0d3772b8ff75cfb0041R238) or should we leave it at that?

skjnldsv:
> Do you want to continue over here or should we leave it at that?

What do you mean?

Henni:
line 238 also contains a calc function that could be moved over to scss.

PR #500, opened 2018-02-28
Three dot menu for Download and Delete

rutujasurve94:
Patch to introduce a three dot menu icon for download and delete (pipovermenu using icon-more class)

### DIFF
--- a/templates/contactDetails.html
+++ b/templates/contactDetails.html
@@ -34,11 +34,22 @@
 				</div>
 			</div>
 			<div id="details-actions">
-				<button id="contact-failed-save" ng-click="ctrl.updateContact()" class="icon-checkmark-white" ng-if="ctrl.contact.failedProps.length>0" title="{{ctrl.t.save}}"></button>
-				<a href="{{ctrl.contact.data.url}}" id="contact-export-link"
-				   class="icon-download-white" title="{{ctrl.t.download}}"
-				   download="{{ ctrl.contact.readableFilename() }}"></a>
-				<button ng-click="ctrl.deleteContact()" ng-if="!ctrl.addressBook.readOnly" class="icon-delete-white" title="{{ctrl.t.delete}}"></button>
+				<div>
+					<span class="icon-more"></span>
+					<div class="popovermenu">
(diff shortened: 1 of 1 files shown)


rutujasurve94:
@jancborchardt @Henni @skjnldsv Please review this pull request, I've been following https://docs.nextcloud.com/server/13/developer_manual/design/popovermenu.html for the pipovermenu

skjnldsv:
The popover won't open if you don't create the javascript behind it :)
Otherwise, the structure looks good :+1: 

@rutujasurve94 since you're a member of the contacts now, could you use the official repository to create your branches? it makes it easier for us to review and help :)
Also, you can set your own labels on the right when creating a new issue/pr ;)

rutujasurve94:
@skjnldsv Could you guide regarding where exactly in contactDetails_controller.js should I introduce the script for enabling the popovermenu action? This would add a function that in turn calls something like ctrl.deleteContact and downloadContact right? Could you point to a similar action in the contacts codebase which I could refer for the same?

skjnldsv:
Of course, we use a similar function in the addressbook list:
https://github.com/nextcloud/contacts/blob/master/js/components/addressBook/addressBook_controller.js#L57
https://github.com/nextcloud/contacts/blob/master/templates/addressBook.html#L17

skjnldsv:
A little update on this? :)

rutujasurve94:
Hi! @skjnldsv @jancborchardt, I went through the addressBook.html and addressBook_controller.js codes. Should I simply add the following snippet to addressBook.html file : 
	<span class="action">
		<span
			class="icon-more popovermenu"
			ng-click="ctrl.toggleMenu($id)">
		</span>
	</span>
Since the openMenu, closeMenus and toggleMenu methods are already defined in the controller file, we don't need to modify that, right?
If this approach is correct, I proceed with it and raise a patch as well.

skjnldsv:
@rutujasurve94 yes, but you also need to make sure the ctrl.toggleMenu is accessible in the scope location you're using (in your case, contactsDetails controller) :)

Issue #445, opened 2017-11-24
Number of Contacts shows just 999+ instead of real number


MarkusH1975:
### Steps to reproduce
1. add more the 999 contacts, e.g. 1255

### Expected behaviour
Should show e.g. 1255 instead of 999+

### Actual behaviour
Shows 999+

**Nextcloud version:** (see Nextcloud admin page)
V12.0.3
**Contacts version:** (see Nextcloud apps page)
V2.0.1

![contacts](https://user-images.githubusercontent.com/30717540/33207094-aa844f74-d10c-11e7-8541-152d105d0742.png)

irgendwie:
This was/is a design choice - do you have a need for the exact number of contacts?

mburnicki:
If the real number of contacts is shown anyway, why not show it? This would really help verifying that everything works as expected.

The current situation with PIM apps is generally pretty unsatisfying, IMO. I've got a huge number (~3000) of contacts in different address books that evolved over several years. Originally this was hosted with owncloud, and you could at least edit all properties with the contacts app. 

Some time ago I migrated to nextcloud, using several different clients, each of which has some limitations:

- This contacts app (v2.0.1) is unable sort the contacts list reliably. I'd prefer sorting contacts from all addressbooks by "last name, first name", but even if I select sorting by last name I can't see that this is done correctly. Also, some contacts that are known to be available in the database are not shown at all. Don't know why, and didn't have time to try to find out, yet. Anyway, if you don't see a properly sorted list it's hard to find out. If the true number of contacts was shown I could at least find out if all contacts have been loaded, or some are missing. Loading the contacts from nextcloud is *very* slow. 

- The cardbook addon for Thunderbird has made some good progress in the recent past. It loads the contacts much faster, but also some contacts are missing in the displayed list. Interestingly, the contacts missing here are different ones than in the contact app.

Eventually there are some entries in the nextcloud DB that have invalid characters or something due to bugs in earlier versions of some client or owncloud server version, but with a huge number of contacts it's hard to find out which ones are missing, and why, and fix it.

- The KDE PIM software shows all contacts, even those that are skipped by nextcloud/contact and cardbook. So you can at least verify that a specific contact is in the cloud DB. However, doing changes in this software removes some information when the contact is saved back in the DB.

- Android contacts editor seems to load all contacts, too, and it basically seems to have the least number of bugs. It has at least a statistics menu wherecan quickly check how many contacts are in each address book.

[...]

PR #509, opened 2018-03-04
Fix issue 445 to increase range of number of contacts shown.

sleepypioneer:
(#445 ) Worked with @suntala as Team Popcorn for RGSoC 2018 application.

We increased the counter logic to only show a static number from above 9999 contacts of 9999+ instead of the original 999+ We changed the padding in the css so that there was none on the left hand side, which causes a number of this size to go outside of the element.

We also changed the test file counterFormatter_filter.js, to reflect the above change in the script.js file.

We are aware that above 9999+ you no longer see the real number and would propose to implement a hover over function to show the full number. @jancborchardt @skjnldsv please let us know what you think of this solution and if we should go ahead with the hover over feature.

### DIFF
--- a/js/tests/filters/counterFormatter_filter.js
+++ b/js/tests/filters/counterFormatter_filter.js
@@ -9,13 +9,13 @@ describe('counterFormatter filter', function() {
 		$filter = _$filter_;
 	}));
 
-	it('should return the same number or 999+ if greater than 999', function() {
+	it('should return the same number or 9999+ if greater than 9999', function() {
 		var counterFormatter = $filter('counterFormatter');
 		expect(counterFormatter(Number.NaN)).to.be.Nan;
 		expect(counterFormatter(15)).to.equal(15);
 		expect(counterFormatter(0)).to.equal('');
 		expect(counterFormatter(-5)).to.equal(-5);
-		expect(counterFormatter(999)).to.equal(999);
-		expect(counterFormatter(1000)).to.equal('999+');
+		expect(counterFormatter(9999)).to.equal(9999);
+		expect(counterFormatter(10000)).to.equal('9999+');
 	});
 });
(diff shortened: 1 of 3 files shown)


js/tests/filters/counterFormatter_filter.js
```diff
@@ -15,7 +15,7 @@ describe('counterFormatter filter', function() {
 		expect(counterFormatter(15)).to.equal(15);
 		expect(counterFormatter(0)).to.equal('');
 		expect(counterFormatter(-5)).to.equal(-5);
-		expect(counterFormatter(999)).to.equal(999);
-		expect(counterFormatter(1000)).to.equal('999+');
+		expect(counterFormatter(9999)).to.equal(9999);
+		expect(counterFormatter(10000)).to.equal('9999+');
```
Henni: This code just changes the tests. Please also take a look at [this file](https://github.com/nextcloud/contacts/blob/master/js/filters/counterFormatter_filter.js#L7) which contains the actual logic.

Henni:
I would also be fine with completely removing this restriction and always displaying the complete number of contacts in a group (even after 9999).
@sleepypioneer @suntala would this work or would this lead to styling issues if we can't predict the maximum length of the field?

js/tests/filters/counterFormatter_filter.js
sleepypioneer: Many thanks @Henni just realised and changed that also now. :)

skjnldsv:
Welcome aboard! :)
I'm okay with completely removing the count formatter as well :)

sleepypioneer:
@Henni @skjnldsv we would like to merge this PR.

We suggest displaying the full number with hovering, as there would have to be a upper limit due to styling restrictions. What do you think?

Henni:
If you want to have a go at the hover effect, I would suggest using a tooltip. Nextcloud uses the tooltip library from bootstrap (https://getbootstrap.com/docs/3.3/javascript/#tooltips).

sleepypioneer:
@Henni thank you for the tip, we will have a look into that and get back to you with a new PR.

PR #510, opened 2018-03-05
MD5 invalid uids

### LINKED ISSUE #487: Broken contact links in web interface
Opened [this issue in the server repo](https://github.com/nextcloud/server/issues/8404) but I was told it belongs here. :innocent: 

I recently upgraded to NC 13 and it seems some links in the Contacts app are not working in the web interface.

When I hover the contacts in the contact list (no click) I see 2 different types of links:

1. https://domain.com/apps/contacts/#/Group/e5c9a57f-d84c-4b8c-a59b-ee742e1b42a1
2. https://domain.com/apps/contacts/#/Group/https://User@old.domain.com/remote.php/carddav/addressbooks/User/contacts/2EAF6525-17ADC861-38D6BB1D.vcf

The second link type is broken (i.e.: if I click it, I get redirected to the Group's default page, which is the first contact in the group).

This did not happen before upgrade. Is it a known issue? Is there any way I can fix that?

Please, tell me if you need any further information.

Notes
------

- The contact information is intact and correct (i.e.: no information lost). Only the link is broken.
- Contacts that have the second-type link are older (i.e.: were created maybe long time ago). Contacts that have the first-type link are all more recent contacts.
- The second link contains "old.domain.com" too. The domain change is, however, something I did some time ago, before even upgrading to NC 12. And I didn't have this problem with NC 12.

Some extra information
------------------------

- Server:
  - OS: Ubuntu 16.04.3 LTS (Linux nextcloud 4.4.0-104-generic #127-Ubuntu SMP Mon Dec 11 12:16:42 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux)
  - Web server: Apache
  - PHP version: PHP 7.0.25-0ubuntu0.16.04.1
  - Nextcloud version: 13.0.0
  - Upgraded from: Nextcloud 12 (fully upgraded when 13.0.0 was released)

skjnldsv:
Fix #487

If not saved, vcard with fixed uid is not sent to server. But for better handling we use our own md5 if not valid uid. It should not cause any issues.

vcard test 
[2EAF6525-17ADC861-38D6BB1D.vcf.txt](https://github.com/nextcloud/contacts/files/1780604/2EAF6525-17ADC861-38D6BB1D.vcf.txt)

### DIFF
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -1,5 +1,5 @@
 angular.module('contactsApp')
-.factory('Contact', function($filter, MimeService) {
+.factory('Contact', function($filter, MimeService, uuid4) {
 	return function Contact(addressBook, vCard) {
 		angular.extend(this, {
 
@@ -28,7 +28,9 @@ angular.module('contactsApp')
 					return model.setProperty('uid', { value: value });
 				} else {
 					// getter
-					return model.getProperty('uid').value;
+					var uid = model.getProperty('uid').value;
+					/* global md5 */
+					return uuid4.validate(uid) ? uid : md5(uid);
 				}
 			},
 


Henni:
Haven't tested it yet, but I am fine with going this way for now. It isn't a great solution, but I see that the alternative solution would take a lot more work to implement.
So to fix #487 in a timely fashion: 👍

Henni:
Works as expected with the given vcf.

During testing I realized that there would be an alternative way to solve this.
Every contact has a href attribute which we could use instead of the uid for identification in the url.

@skjnldsv what do you think of this? is it worth changing in your opinion?

skjnldsv:
I'm not convinced. Feels like too much random. With this we're good to go and we're sure we won't encounter any further issues :)

PR #511, opened 2018-03-05
Fixed lastName sorting

### LINKED ISSUE #505: Contact sorting still doesn't work as expected
At the moment we have the option to sort the contacts by "Display name", "First name" or "Last name". But the "Last name" sorting still doesn't work.

Her are simple data to test with:

BEGIN:VCARD
VERSION:3.0
FN:Auser Zname 
UID:4d73a9f3-a63c-453c-b042-5620c6373115
N:Zname;Auser;;;
REV:20180228T151648Z
END:VCARD
BEGIN:VCARD
VERSION:3.0
FN:Buser Yname 
UID:8fc04e91-a1e0-44a5-a17b-90c15a2ff692
N:Yname;Buser;;;
REV:20180228T151718Z
END:VCARD
BEGIN:VCARD
VERSION:3.0
FN:Cuser Xname 
UID:5db8def9-31fc-4504-9f4c-9241d67f6791
N:Xname;Cuser;;;
REV:20180228T151751Z
END:VCARD

First I imported these data in to the default address book:

contacts1

contacts2

Afterwards I switched to "Last Name" Sorting.

contacts3

This seems to be ok. But after switching to a different app and coming back to the contacts, I see the following:

contacts4

For "Auser Zname" the entry is switched to "Zname, Auser" and moved in the list. But only the selected contact entry is affected, which was the first one in the list.

The only way to get the list sorted by "Last name" is to click every single entry.

contacts5

I think we need a solution here. Otherwise the contact app isn't really usable in a business environment.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55622541-contact-sorting-still-doesn-t-work-as-expected?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
Fix #505

### DIFF
--- a/js/dav/dav.js
+++ b/js/dav/dav.js
@@ -1799,7 +1799,7 @@ var listVCards = _co2['default'].wrap(regeneratorRuntime.mark(function callee$0$
       case 0:
         debug('Doing REPORT on address book ' + addressBook.url + ' which belongs to\n        ' + addressBook.account.credentials.username);
 
-        vCardListFields = ['EMAIL', 'UID', 'CATEGORIES', 'FN', 'TEL', 'NICKNAME'].map(function (value) {
+        vCardListFields = ['EMAIL', 'UID', 'CATEGORIES', 'FN', 'TEL', 'NICKNAME', 'N'].map(function (value) {
           return {
             name: 'prop',
             namespace: ns.CARDDAV,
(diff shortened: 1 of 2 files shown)


RainerEmrich:
I can't approve, but this simple change indeed fixes the sorting.

jancborchardt:
@RainerEmrich just wanted to say how much we appreciate people testing pull requests – it helps a lot and is always welcome! :rocket:

PR #530, opened 2018-03-31
Create "related" property and show info icon to explain it

### LINKED ISSUE #426: Relation property not being synced to android
Hi,

it seems the "relation" property of contacts is not properly synced to android.

### Steps to reproduce
Case A:
1. Create dummy account with relation entry in nextcloud . 
2. Sync to android using DAVDroid.

Case B:
1. Create dummy account with relation entry on android.
2. Sync to nextcloud using DAVDroid.

### Expected behaviour
Case A:
The dummy account should have an entry for relation on android.

Case B:
The dummy account should have an entry for relation in nextcloud..

### Actual behaviour
Case A:
The dummy account has no relation entry on android.

Case B:
The dummy account has no relation entry in nextcloud.

### General server configuration

**Operating system:** Linux Spielzeug 4.4.0-97-generic #120-Ubuntu SMP Tue Sep 19 17:28:18 UTC 2017 x86_64

**Web server:** Apache/2.4.18 (Ubuntu) (apache2handler)

**Database:** mysql 5.7.20

**PHP version:** 7.0.22-0ubuntu0.16.04.1

**Nextcloud version:** 12.0.3 - 12.0.3.3

**Contacts Version:** 2.0.1

xh3n1:
Created the "related" field and info icons to explain this and "relationship" field, to avoid confusion. 

Fix #426

### DIFF
--- a/templates/propertyGroup.html
+++ b/templates/propertyGroup.html
@@ -1,6 +1,7 @@
 <h3 class="propertyGroup__title">
 	<i ng-class="ctrl.getIconClass()"></i>
 	{{ctrl.getReadableName()}}
+	<i ng-class="ctrl.getInfoClass()" title="{{ctrl.getInfoText()}}"></i>
 </h3>
 <detailsItem ng-repeat="property in ctrl.properties" name="ctrl.name" data="property" model="ctrl.contact" index="$index"
 			 class="details-item-{{ctrl.name}}" ng-class="{ 'failed': ctrl.contact.failedProps.indexOf(ctrl.name) !== -1 }"></detailsItem>
--- a/js/components/propertyGroup/propertyGroup_controller.js
+++ b/js/components/propertyGroup/propertyGroup_controller.js
@@ -12,6 +12,17 @@ angular.module('contactsApp')
 		return ctrl.meta.icon || 'icon-contacts-dark';
 	};
 
+	this.getInfoClass = function() {
+		if (ctrl.meta.hasOwnProperty('info')) {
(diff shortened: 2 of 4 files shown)


templates/propertyGroup.html
```diff
@@ -1,6 +1,7 @@
 <h3 class="propertyGroup__title">
 	<i ng-class="ctrl.getIconClass()"></i>
 	{{ctrl.getReadableName()}}
+	<i ng-class="ctrl.getInfoClass()" title="{{ctrl.getInfoText()}}"></i>
```
skjnldsv: We also should only add this element if there is a text, use ng-if and the length of the getInfoText? :)

xh3n1:
@skjnldsv Thanks for the review. I made the changes.

skjnldsv:
@xh3n1 just add the if condition to make sure items without a text don't have this element! :)

AS-dh:
This is great news. Unfortunately the syntax used here, while consistent with https://tools.ietf.org/html/rfc6350#section-6.6.6, for example:
     RELATED;TYPE=CHILD:Theresa 21.05.1987
is different from what Google uses when exporting to the vcard format:
     item1.X-ABRELATEDNAMES:Theresa 21.05.1987
     item1.X-ABLabel:_$!<Child>!$_

Since I suppose many asking for this specific feature are trying to transfer their contacts from Google to Nextcloud and claim their independence :-), it would be very useful to have some sort of compatibility mode for the initial import.

PR #540, opened 2018-04-11
Fix issue 445 to display number of contacts in tooltips.

suntala:
(#445 ) Worked with @sleepypioneer as Team Popcorn, RGSoC 2018 applicants.

We enabled tooltips for group contact numbers above 9999 to display the actual number instead of 9999+. We created a new filter for this (counterTooltipDisplay) as well as a test to go along with it.

As a next step, we propose to implement this feature for all groups in contacts, not just the default groups. @jancborchardt @skjnldsv please let us know what you think of this solution and if we should go ahead with it.


sleepypioneer:
@skjnldsv We just added another PR for this #542 to add Tooltips for non default groups also

PR #542, opened 2018-04-12
Add Tooltips for non default groups

sleepypioneer:
(#445) Worked with @suntala  as Team Popcorn, RGSoC 2018 applicants.

We enabled tooltips for group contact numbers for groups that are not default to complete this issue.


skjnldsv:
Yes, somehow we missed that! :see_no_evil:

Issue #257, opened 2017-07-15
Wrong Etag return from server prevents modifications from being saved 


GnaXi:
### Steps to reproduce
1. Open the Contacts app
2. Perform any kind of change to a contact or add a new contact
3. Reload the page afterwards

### Expected behaviour
The information typed in any contact should be saved

### Actual behaviour
Information typed in the lost 5-10 seconds before refreshing the page is lost

**Operating system**: Debian 9

**Web server:** Apache 2.4.25

**Database:** MariaDB 10.1.23

**PHP version:** 5.6.30

**Nextcloud version:** 12.0.0

**Contacts version:** 1.5.3

**Updated from an older Nextcloud or fresh install: Migrated from ownCloud 10 to Nextcloud 12

```
Login as admin user into your Nextcloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.

No errors have been found.
```

**List of activated apps:**
Enabled:
  - activity: 2.5.2
  - bruteforcesettings: 1.0.2
  - calendar: 1.5.3
  - comments: 1.2.0
  - contacts: 1.5.3
  - dav: 1.3.0
  - federatedfilesharing: 1.2.0
  - files: 1.7.2
  - files_external: 1.3.0
  - files_pdfviewer: 1.1.1
  - files_sharing: 1.4.0
  - files_texteditor: 2.4.1
  - files_trashbin: 1.2.0
  - files_versions: 1.5.0
  - files_videoplayer: 1.1.0
  - firstrunwizard: 2.1
  - gallery: 17.0.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.0.0
  - news: 11.0.5
  - nextcloud_announcements: 1.1
  - notifications: 2.0.0
  - oauth2: 1.0.5
  - password_policy: 1.2.2
  - provisioning_api: 1.2.0
  - registration: 0.2.5
  - serverinfo: 1.2.0
  - sharebymail: 1.2.0
  - survey_client: 1.0.0
  - systemtags: 1.2.0
  - theming: 1.3.0
  - twofactor_backupcodes: 1.1.1
  - twofactor_totp: 1.3.0
  - updatenotification: 1.2.0
  - workflowengine: 1.2.0
```

**Nextcloud configuration:**
{
    "system": {
        "updatechecker": false,
        "instanceid": "oc45dii8hdi1",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
[...]

PR #549, opened 2018-04-25
Fix "412 precondition failed" when response has no ETag, but OC-ETag …

phillipp:
A customer of our webhosting service www.lima-city.de (I'll shemlessly plug it here) complained about the issue described in #257 and #135. We had a look at the issue and it seems like @xenithorb was correct: there is another issue underlying. It looks like Apache modified the ETag header. Luckily we have the OC-ETag header that from this patch forward the JS is going to use. Bang, problem solved.

Please review. I'm not sure if there are other solutions that adress this already.

### DIFF
--- a/js/services/contact_service.js
+++ b/js/services/contact_service.js
@@ -234,7 +234,7 @@ angular.module('contactsApp')
 				filename: newUid + '.vcf'
 			}
 		).then(function(xhr) {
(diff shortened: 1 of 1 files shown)


skjnldsv:
I completely thought I already implemented this fix... 🙈
Thanks a lot for your help! 😊👍

Issue #106, opened 2017-01-26
Allow assignment of images already in Nextcloud as contact photos


mossholderm:
- [ ] If the image is already in Nextcloud, should we really need to upload it again?
- [ ] Provide an editor to cut and resize images like we don on the user settings

jancborchardt:
@icewind1991 @nickvergessen @MorrisJobke who implemented the »Choose from Files« function for the user avatar on the personal settings page? We could reuse that here. :)

MorrisJobke:
> @icewind1991 @nickvergessen @MorrisJobke who implemented the »Choose from Files« function for the user avatar on the personal settings page? We could reuse that here. :)

I think @icewind1991 added this. The Choose from Files is `OC.dialogs.filepicker()` and those parameters: https://github.com/nextcloud/server/blob/a2f6fea4081e7920ed6cd33a96e2b80dc0c51303/core/js/oc-dialogs.js#L170-L178

jancborchardt:
@MorrisJobke thanks for the pointer! :) I’m setting this as starter issue then, would you agree?

skjnldsv:
I edited OP to merge two issues into one, see #33

PR #550, opened 2018-04-26
Introduced 3-dot-menu in top right for Download and Delete

### LINKED ISSUE #354: Introduce 3-dot-menu in top right for Download and Delete
Instead of showing these icons (and possibly more in the future) permanently, and without text, we should show a 3-dot-icon which will open a menu listing those actions. That will also save space on mobile to display the name more properly.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49584209-introduce-3-dot-menu-in-top-right-for-download-and-delete?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sleepypioneer:
Fix #354 
Closes #500

I've added the 3 dot menu for the Contact Details header, please check my solution, particularly the styling as I want to tackle #106 with a similar technique. I removed the titles for these symbols as I gave them text spans to mirror the addressbook popover menu. I gave the update option a tooltip to explain that the contact has not been saved and needs to be updated. Please let me know if this explains the situation ok for you.

I wondered if the 3 dot menu should also have a tool tip or if this is too much?

One thing I have noted is that the bin icon only turns red when it is hovered over and not when it's corresponding text is hovered over - I could add a solution for this if desired? @skjnldsv @jancborchardt

### DIFF
--- a/css/_details.scss
+++ b/css/_details.scss
@@ -68,15 +68,11 @@
 	width: 100%;
 }
 
-.contactdetails__header #details-actions {
-	display: flex;
-	flex-shrink: 0;
+.contactdetails__header #details-actions div.icon-more-white {
+	cursor: pointer;
 	padding: 14px;
 	width: 44px;
-}
-
-.contactdetails__header #details-actions span.icon-more-white {
-	cursor: pointer;
+	height: 44px;
 }
 
 .contactdetails__header #contact-failed-save {
--- a/css/_details.scss
+++ b/css/_details.scss
@@ -71,6 +71,8 @@
 .contactdetails__header #details-actions {
 	display: flex;
 	flex-shrink: 0;
+	padding: 14px;
+	width: 44px;
 }
 
 .contactdetails__header #contact-failed-save {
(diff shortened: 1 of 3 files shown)


skjnldsv:
Hey! :)
That looks really great. Could you set the proper labels for this pr? That helps us identify if this is ready for review o not ;)

For you questions:
1. a tooltip is probably a bit too much 
2. you can use the icon-delete class on the `li` instead of a dedicated `span`. Like the first example of the guidelines :)

Cheers

sleepypioneer:
@skjnldsv 

By changing the label for the pr, do you mean the edit you have done? Or the label section down the right hand side, not sure we did that before... sorry about any confusion!

I added the class to the <a> I think thats what you meant looking at the example and have done this to all of them to keep them inline.

No extra tooltips ;)

Best wishes,

Jessica

skjnldsv:
I meant this: "sleepypioneer added the `3. to review`  label 2 minutes ago" :)

sleepypioneer:
No problems, thank you!

sleepypioneer:
@skjnldsv this is ready for review now :)

css/_details.scss
```diff
@@ -71,6 +71,12 @@
 .contactdetails__header #details-actions {
 	display: flex;
 	flex-shrink: 0;
+	padding: 14px;
+	width: 44px;
+}
+
+.contactdetails__header #details-actions span.icon-more-white {
+	cursor: pointer;
```
skjnldsv: According to the docs https://docs.nextcloud.com/server/13/developer_manual/design/popovermenu.html#technical-details
img
> The required right distance to the border (or padding, whatever you want to use) of the three-dot icon should be 14px (5 for menu margin and 6 for arrow position)

Especially since the click handler is on the icon-more and not in the parent. Which makes the more element too small ;)

|bad|good|
|:--:|:--:|
|capture d ecran_2018-05-03_10-04-49 |capture d ecran_2018-05-03_10-04-25|



You need to change the span to a div (or put a display:block) and set the `padding:14px`, `height:44px` and `width:44px` to the icon-more div. Otherwise it's not centered :)

PR #595, opened 2018-08-15
Vue share addressbook

sleepypioneer:
Hi @suntala and @skjnldsv on this branch I have the changes we did in the meeting for the SCSS files. I also have been working on the share-addressbook. This is almost finished bar one styling fix, so I thought it may be worth a PR, thoughts??


css/settings-section.scss
```diff
@@ -21,6 +21,11 @@
  *
  */
 
+
+@import 'settings/settings-addressbook';
```
skjnldsv: I would prefer all imports on the same file :)

css/settings/settings-addressbook-sharee.scss
```diff
@@ -0,0 +1,46 @@
+.addressbook__sharee {
```
skjnldsv: copyright header

css/settings/settings-addressbook-shares.scss
```diff
@@ -0,0 +1,56 @@
+.addressbook-shares {
```
skjnldsv: copyright header

css/settings/settings-addressbook.scss
```diff
@@ -0,0 +1,31 @@
+// BLOCK
```
skjnldsv: copyright header

css/settings/settings-addressbook.scss
```diff
@@ -0,0 +1,31 @@
...
+		&:active {
+			opacity: .7;
+		}
+	}
+	&__share {
+		margin-left: auto;
+	}
+}
+
+// MODIFIER
```
skjnldsv: Not useful I guess? :)

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -22,50 +22,46 @@
...
+			track-by="match"
+			label="match"
+			open-direction="bottom"
+			class="multiselect-vue"
+			@search-change="asyncFind"
+			@input="shareAddressbook">
+			<template slot="singleLabel" slot-scope="props"><span class="option__desc"><span class="option__title">{{ props.option.matchpattern }}</span></span></template>
+			<template slot="option" slot-scope="props">
+				<div class="option__desc">
+					<span>{{ props.option.matchstart }}</span><span class="shareematch--bold">{{ props.option.matchpattern }}</span><span>{{ props.option.matchend }} {{ props.option.matchtag }}</span>
```
skjnldsv: Each one on its line?

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -22,50 +22,46 @@
...
+				</div>
+			</template>
+			<span slot="noResult">{{ noResult }} </span>
+		</multiselect>
 		<!-- list of user or groups addressbook is shared with -->
 		<ul v-if="addressbook.shares.length > 0" class="addressbook__shares__list">
-			<address-book-sharee v-for="sharee in addressbook.shares" :key="sharee.name" :sharee="sharee" />
+			<address-book-sharee v-for="sharee in addressbook.shares" :key="sharee.displayname + sharee.group" :sharee="sharee" />
```
skjnldsv: You can also use the index :)
`v-for="(sharee, index) in addressbook.shares" :key="index"`

Both are fine as long as thy're unique. But we can't make sure displayNames are uniques, id are though :)
Pick one ;)

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -85,32 +81,94 @@ export default {
...
-			selectedUserOrGroup: []
+			usersOrGroups: []
+		}
+	},
+	computed: {
+		placeholder() {
+			return t('contacts', 'Share with users or groups')
+		},
+		noResult() {
+			return t('contacts', 'Oops! No elements found. Consider changing the search query.')
```
skjnldsv: Maybe too long?
I would go for something simpler I think.

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -85,32 +81,94 @@ export default {
...
 	},
 	methods: {
-		limitText(count) {
-			return `and ${count} other users or groups`
+		/**
+		 * Share addressbook
+		 *
+		 * @param {Object} chosenUserOrGroup
+		 */
+		shareAddressbook(chosenUserOrGroup) {
```
skjnldsv: you can also replace `chosenUserOrGroup` by `{ sharee, group }`
Easier to understand and you don't need the `let xxx = `

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -85,32 +81,94 @@ export default {
...
+		formatMatchResults(matches, query, group) {
+			if (matches.length < 1) {
+				return
+			}
+			let regex = new RegExp(query, 'i')
+			let existingSharees = []
+			for (let j = 0; j < this.addressbook.shares.length; j++) {
+				existingSharees.push(this.addressbook.shares[j].displayname + this.addressbook.shares[j].group)
+			}
+			for (let i = 0; i < matches.length; i++) {
```
skjnldsv: There is a cleaner way to do that.
You can map the existing shares into an array of displaynames
`let existingSharees = this.addressbook.shares.map(share => share.displayname)`
Then you can filter out the shares
`matches.filter(share => existingSharees.indexOf(share) === -1)`
And then you can use map to create your `newMatch` object :)

Though this raises a few questions:
1. the group/user request returns **displayNames**, but we can have same **displayName** into a same instance
2. we should probably retrieve both data from the ocs api. and use ids and displayNames alongside

The api doc is not updated yet, but we can use `/users/details` and `/groups/details` to get the full data of a user.
https://github.com/nextcloud/server/blob/master/apps/provisioning_api/appinfo/routes.php#L37
Though we should then probably limit the number of results: https://github.com/nextcloud/server/blob/master/apps/provisioning_api/lib/Controller/UsersController.php#L156

src/components/Settings/SettingsNewAddressbook.vue
```diff
@@ -72,9 +72,10 @@ export default {
 		 * @returns {Promise}
 		 */
 		addAddressBook() {
-			// let addressBook = this.$refs.addressBook.value
-			// let addressBooks = this.$store.getters.getAddressbooks
-			// let newAddressBooksArray = addressBooks.push(addressBook)
+			let addressBook = this.$refs.addressBook.value
+			let addressBooks = this.$store.getters.getAddressbooks
+			let newAddressBooksArray = addressBooks.push(addressBook)
```
skjnldsv: You cannot do that. You are not allowed to mutate data outside the store :)

src/store/addressbooks.js
```diff
@@ -101,21 +101,23 @@ const mutations = {
...
 	 */
-	shareAddressbook(state, addressbook, sharee) {
-		addressbook = state.addressbooks.find(search => search === addressbook)
+	shareAddressbook(state, { addressbook, sharee, group }) {
+		addressbook = state.addressbooks.find(search => search.id === addressbook.id)
 		let newSharee = {}
-		sharee.displayname = sharee
-		sharee.writable = false
-		addressbook.shares.append(newSharee)
+		newSharee.displayname = sharee
```
skjnldsv: You can do that directly in the object declaration

css/settings-section.scss
sleepypioneer: copy also to the copyright header I will do this when in today as the first thing I do. Sorry forgot about the debouncer I need to think how I transfer these comments to a todo for myself, I loose them on github. I will work through the other points also. The only one is addnewaddressbook that is not my work and I think @suntala has probably updated this on her branch.

skjnldsv: Arf, then forget about the new addressbook :) :tada: 
Sorry :see_no_evil:

src/components/Settings/SettingsAddressbookShare.vue
sleepypioneer: I am not sure how this should work, the chosenUserOrGroup is an object that is the return from the custom options, it looks like this:

let newMatch = {
					match: matches[i],
					matchstart: matchResult[0],
					matchpattern: matches[i].match(regex)[0],
					matchend: matchResult[1],
					matchtag: group ? '(group)' : '(user)',
					matchgroup: group
				}

How do I tell the multiselect to only pass on the match and matchgroup as suggested?

src/store/addressbooks.js
sleepypioneer: 💯

src/components/Settings/SettingsAddressbookShare.vue
skjnldsv: You can still pass the full object and destructure the only desired props :)

``` js
({ match, matchgroup } = {
    match: 0,
    matchstart: 1,
    matchpattern: 2,
    matchend: 3,
    matchtag: '(user)',
    matchgroup: 4
})

match
> 0
matchgroup
> 4
```

sleepypioneer: 💯

sleepypioneer:
@skjnldsv I believe I have covered the changes requested, sorry if I have missed any but I think this is good to go.

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -145,12 +147,12 @@ export default {
 					let matchingUsers = Object.values(response[0].data.ocs.data.users)
 					let matchingGroups = response[1].data.ocs.data.groups
 					try {
-						this.formatMatchResults(matchingUsers, query, false)
+						debounce(this.formatMatchResults(matchingUsers, query, false), 250, true)
```
skjnldsv: I used 500 in the details, but this is a different behavior, so 250 is also fine :)

src/components/SettingsSection.vue
```diff
@@ -23,9 +23,9 @@
 <template>
 	<div>
 		<ul id="address-book-list">
-			<address-book v-for="addressbook in addressbooks" :key="addressbook.id" :addressbook="addressbook" />
+			<address-Book v-for="addressbook in addressbooks" :key="addressbook.id" :addressbook="addressbook" />
```
skjnldsv: why the capital B ?

src/store/addressbooks.js
```diff
@@ -101,21 +101,26 @@ const mutations = {
...
-		sharee.displayname = sharee
-		sharee.writable = false
-		addressbook.shares.append(newSharee)
+	shareAddressbook(state, { addressbook, sharee, id, group }) {
+		addressbook = state.addressbooks.find(search => search.id === addressbook.id)
+		let newSharee = {
+			displayname: sharee,
+			id: id,
+			writeable: false,
+			group: group
```
skjnldsv: Quick tip, you can just put group since it will be the same as group: group.
But don't bother, it's just a quick knowledge bump ;)

src/components/Settings/SettingsAddressbookSharee.vue
```diff
@@ -22,7 +22,8 @@
 
 <template>
 	<li class="addressbook__sharee">
-		<span class="icon icon-group" />
+		<span v-if="sharee.group" class="icon icon-group" />
```
skjnldsv: you could just do `<span :class="sharee.group ? 'icon-group' : 'icon-user'" />`

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -85,32 +86,88 @@ export default {
...
+				]).then(response => {
+					let matchingUsers = Object.values(response[0].data.ocs.data.users)
+					let matchingGroups = response[1].data.ocs.data.groups
+					try {
+						debounce(this.formatMatchResults(matchingUsers, query, false), 250, true)
+					} catch (error) {
+						console.debug(error)
+					}
+					try {
+						debounce(this.formatMatchResults(matchingGroups, query, true), 250, true)
```
skjnldsv: This is wrongly positioned, the debounce should be before the request, or we'll still get a big amount of requests which data won't be up to date. :)

See https://github.com/nextcloud/contacts/blob/vue/src/components/properties/PropertyMultipleText.vue#L123

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -85,32 +86,88 @@ export default {
...
+			let regex = new RegExp(query, 'i')
+			let existingSharees = this.addressbook.shares.map(share => share.id + share.group)
+			matches = matches.filter(share => existingSharees.indexOf(share.id + group) === -1)
+			for (let i = 0; i < matches.length; i++) {
+				let matchResult = matches[i].displayname.split(regex)
+				let newMatch = {
+					sharee: matches[i].displayname,
+					id: matches[i].id,
+					matchstart: matchResult[0],
+					matchpattern: matches[i].displayname.match(regex)[0],
```
skjnldsv: match will return undefined if no match. so undefined[0] will throw an error :)

src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -85,32 +86,88 @@ export default {
...
+		 * @param {Boolean} group
+		 */
+		formatMatchResults(matches, query, group) {
+			if (matches.length < 1) {
+				return
+			}
+			let regex = new RegExp(query, 'i')
+			let existingSharees = this.addressbook.shares.map(share => share.id + share.group)
+			matches = matches.filter(share => existingSharees.indexOf(share.id + group) === -1)
+			for (let i = 0; i < matches.length; i++) {
```
skjnldsv: You can also directly do that in a map :)

``` js
            this.usersOrGroups.concat(matches.map(match => {
                let matchResult = match.displayname.split(regex)
                return {
                    sharee: match.displayname,
                    id: match.id,
                    matchstart: matchResult[0],
                    matchpattern: match.displayname.match(regex)[0],
                    matchend: matchResult[1],
                    matchtag: group ? '(group)' : '(user)',
                    group
                }
            }))
```

sleepypioneer:
Nice notes! I will get on them, thank you :)

src/components/SettingsSection.vue
sleepypioneer: mistake!

src/components/Settings/SettingsAddressbookShare.vue
sleepypioneer: so I have done a check that match has a length and if not just returned out of it

src/components/Settings/SettingsAddressbookShare.vue
sleepypioneer: I've taken 500 for the wait, correct?

PR #622, opened 2018-09-10
Vue rename addressbook

### LINKED ISSUE #614: [Vue] Popover menu for Addressbook settings
- [x] Copy link
- [x] Download
- [x] Rename
- [x] Enabled
- [x] Delete
    ``` js
        menu.push({itemObject})
    }
    ```

<bountysource-plugin>

---
</bountysource-plugin>

sleepypioneer:
Partial fix #614 moved the rename input into the popover menu

@skjnldsv I have moved the rename input into the popover menu as discussed, the function works however let me know if I missed any checks or styling :)


src/components/core/popoverMenu/popoverItem.vue
```diff
@@ -31,7 +31,17 @@
 		</a>
 		<!-- If item.input is set instead, an put will be used -->
 		<span v-else-if="item.input" class="menuitem">
-			<input :id="key" :type="item.input" :class="item.input"
+			<!-- does not show if input is checkbox -->
+			<span v-if="item.input != 'checkbox'" :class="item.icon" />
```
skjnldsv: We need radio as well ;)
But I did not checked how we can do that with radios actually.
Because the inputs then needs to be all in the same form and I don't know how we could manage a form across multiple popover items :thinking: 
Any ideas?

src/components/core/popoverMenu/popoverItem.vue
```diff
@@ -31,7 +31,17 @@
...
+			<!-- only shows if input is text -->
+			<form v-if="item.input == 'text'"
```
skjnldsv: always use `===`

src/components/core/popoverMenu/popoverItem.vue
```diff
@@ -31,7 +31,17 @@
...
+				:class="item.input" @submit.prevent="item.action">
+				<input :type="item.input" :value="item.value" :placeholder="item.placeholder">
+				<input v-if="item.input == 'text'" type="submit" value=""
+					class="icon-confirm">
+			</form>
+			<input v-else
```
skjnldsv: Put the first on the same line please ;)

src/components/core/popoverMenu/popoverItem.vue
sleepypioneer: hmm good point, it seems something that becomes more clear when we have a use case, for example with the text. Is there something from the docs/ original app I could try recreate to work this out?

src/components/core/popoverMenu/popoverItem.vue
sleepypioneer: noted

src/components/core/popoverMenu/popoverItem.vue
skjnldsv: You're right! Let's support only checkbox and text for now!! :+1: :1st_place_medal: :D

src/components/core/popoverMenu/popoverItem.vue
sleepypioneer: ? sorry the first? I think I can remove this v-if actually is that your point, or the class I think the linted wanted it on a separate line

skjnldsv: First attribute, sorry :)

sleepypioneer: Arh you mean for `<input v-else` got you!

sleepypioneer:
@skjnldsv alright changes made!

src/components/core/popoverMenu/popoverItem.vue
```diff
@@ -31,8 +31,16 @@
...
 		<!-- If item.input is set instead, an put will be used -->
 		<span v-else-if="item.input" class="menuitem">
-			<input :id="key" :type="item.input" :class="item.input"
-				v-model="item.model" @change="item.action">
+			<!-- does not show if input is checkbox -->
+			<span v-if="item.input != 'checkbox'" :class="item.icon" />
+			<!-- only shows if input is text -->
+			<form v-if="item.input === 'text'"
+				:class="item.input" @submit.prevent="item.action">
+				<input :type="item.input" :value="item.value" :placeholder="item.placeholder">
```
skjnldsv: Should we add a `required` prop to ensure a proper check as well? :)

skjnldsv:
:tada:

PR #623, opened 2018-09-10
Check for Read only addressbooks

### LINKED ISSUE #614: [Vue] Popover menu for Addressbook settings

sleepypioneer:
Partial fix #614 WIP I've added the checks for the read only however the addressbooks dont have a readonly attribute so I have hard coded it in for now.

### DIFF
--- a/src/components/ContactDetails.vue
+++ b/src/components/ContactDetails.vue
@@ -52,17 +52,17 @@
 				<!-- fullname, org, title -->
 				<div id="contact-header-infos">
 					<h2>
-						<input id="contact-fullname" v-model="contact.fullName" :disabled="!contact.addressbook.enabled"
+						<input id="contact-fullname" v-model="contact.fullName" :disabled="!contact.addressbook.readOnly"
 							:placeholder="t('contacts', 'Name')" type="text" autocomplete="off"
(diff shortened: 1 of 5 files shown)


skjnldsv:
> however the addressbooks dont have a readonly attribute

You can add one then ;)

sleepypioneer:
True, was over thinking it - trying to get it up before I left, XD

skjnldsv:
Since we're at it, could you also fix my error here: https://github.com/nextcloud/contacts/blob/c9155e4d44f012de649ad04a13901fba7a69c58a/src/components/ContentList/ContentListItem.vue#L13

We're supposed to use `readOnly`, not `enabled` :see_no_evil:

skjnldsv:
And bonus: https://github.com/nextcloud/contacts/blob/c9155e4d44f012de649ad04a13901fba7a69c58a/src/views/Contacts.vue#L111

:heart:

sleepypioneer:
Alrighty then! Made those changes @skjnldsv

PR #628, opened 2018-09-17
Vue settings popovermenu

### LINKED ISSUE #614: [Vue] Popover menu for Addressbook settings

sleepypioneer:
Partial fix #614 I have added the loading status for each element in the popover menu and also added a condition that watches to see if the popover menu is closed and sets Editing Name to false. These are currently in a Timeout that is emulating the wev dav request.

I would suggest after renaming that the popover menu is shut, what do you think?

Now what is needed is the proper webdav checks.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3873,7 +3873,7 @@
 		},
 		"buffer": {
 			"version": "4.9.1",
-			"resolved": "http://registry.npmjs.org/buffer/-/buffer-4.9.1.tgz",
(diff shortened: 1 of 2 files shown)


skjnldsv:
> I would suggest after renaming that the popover menu is shut, what do you think?

Agreed! :+1:

sleepypioneer:
Please make final review @skjnldsv and merge thank you :)

Issue #630, opened 2018-09-18
[Vue] Add hint on the property if available


skjnldsv:
See https://github.com/nextcloud/contacts/blob/16befd1d9a63e55fb399c506dfb40445659dc992/templates/propertyGroup.html#L4

https://github.com/nextcloud/contacts/blob/vue/src/models/rfcProps.js#L186

![capture d ecran_2018-09-18_18-00-26](https://user-images.githubusercontent.com/14975046/45700446-dd644c00-bb6c-11e8-9a0c-9152b5dc9e01.png)

PR #634, opened 2018-09-21
Vue property hint #630

suntala:
Add a tooltip displaying the 'info' property as a hint when it is available. #630


skjnldsv:
The tooltip should be on the info icon only, not the whole component ;)

skjnldsv:
capture d ecran_2018-09-25_17-20-26
I added a small fix, looks perfect :)

PR #639, opened 2018-09-25
Vue cdav copy addressbook

sleepypioneer:
Please check my code, not feeling so confident with this


src/components/Settings/SettingsAddressbook.vue
```diff
@@ -222,24 +222,27 @@ export default {
...
 			this.copyLoading = true
+			event.stopPropagation()
+
 			// copy link for addressbook to clipboard
 			this.$copyText(this.addressbook.url).then(e => {
+				event.preventDefault()
 				this.copySuccess = true
 				this.copied = true
+				// timeout sets the text back to copy to show text was copied
+				setTimeout(() => {
```
skjnldsv: Hum,timeout are bad feedback when they're not relevant. Here the string is copied right away but we get the feeling we need to wait :)

sleepypioneer: Ok I just thought otherwise it really doesn't show as its pretty quick!

skjnldsv:
@sleepypioneer I guess so yes :)
I did not noticed before, sorry :see_no_evil:

sleepypioneer:
done, I also added notifications for copy and delete for if they were successful, otherwise I think it is unclear something happened.

src/components/Settings/SettingsAddressbook.vue
```diff
@@ -171,19 +170,17 @@ export default {
...
 		},
 		downloadAddressbook() {
-			// change to loading status
-			this.downloadLoading = true
-			setTimeout(() => {
-				// stop loading status regardless of outcome
-				this.downloadLoading = false
-			}, 1500)
+			// Notify download started
+			OC.Notification.showTemporary(t('contacts', 'Downloading Addressbook'))
```
skjnldsv: Not needed, if the download starts, the user already knows it ;)

sleepypioneer: ok will remove it :) anything else before merging?

skjnldsv: :see_no_evil: :grimacing: :rocket: GO

PR #641, opened 2018-09-25
Vue avatars in list

### LINKED ISSUE #604: [Vue] Avatars in list
- [x] Avatar in list with `?photo` 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62764297-vue-avatars-in-list?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sleepypioneer:
Fix #604 Proposed code for adding avatars in list without needing to check contact.photo using contact.url?photo.

I condensed the two divs and set them to return values based on if an url was returned or not.

I removed the filter as it was simpler to return `this.contact.displayName.charAt(0)`or '' for the initial value.


src/components/ContentList/ContentListItem.vue
```diff
@@ -48,6 +45,16 @@ export default {
 			} catch (e) {
 				return 'grey'
 			}
+		},
+		avatarUrl() {
+			return 'url(' + this.contact.url + '?photo' + ')'
```
skjnldsv: Small tip, with latest javascript you can do
```js
`url(${this.contact.url}?photo)`
```

sleepypioneer: Cool very nice, anything else here to change?

skjnldsv: I'll rebase your branch in a second :)

skjnldsv:
hum, I get no letters :/
capture d ecran_2018-09-25_18-35-06

src/components/ContentList/ContentListItem.vue
```diff
@@ -48,6 +45,16 @@ export default {
 			} catch (e) {
 				return 'grey'
 			}
+		},
+		avatarUrl() {
+			return `url(${this.contact.url}?photo)`
+		},
+		initial() {
+			if (this.avatarUrl) {
```
skjnldsv: A contact always have an url. You cannot assume the contact have a picture or not.
If a contact does NOT have an url, that means it's a local contact (present on the store only and not created on the server **yet**) that have not yet been pushed to the server. So in that case, yes we can assume it doesn't have an url. Though I'll suggest checking the `contact.dav` property to check if this is a local contact or not.

skjnldsv:
capture d ecran_2018-09-25_18-55-08

PR #706, opened 2018-11-10
Sharing design and cdav request

skjnldsv:
Last big step towards the beta! :tada:

### DIFF
--- a/src/components/Settings/SettingsAddressbookShare.vue
+++ b/src/components/Settings/SettingsAddressbookShare.vue
@@ -27,37 +27,26 @@
 			:options="usersOrGroups"
 			:searchable="true"
 			:internal-search="false"
-			:options-limit="250"
-			:limit="3"
 			:max-height="600"
 			:show-no-results="true"
 			:placeholder="placeholder"
 			:class="{ 'showContent': inputGiven, 'icon-loading': isLoading }"
+			:user-select="true"
 			open-direction="bottom"
-			@search-change="asyncFind"
-			@input="shareAddressbook">
-			<template slot="singleLabel" slot-scope="props">
-				<span class="option__desc">
-					<span class="option__title">{{ props.option.matchpattern }}</span>
-				</span>
-			</template>
-			<template slot="option" slot-scope="props">
-				<div class="option__desc">
-					<span>{{ props.option.matchstart }}</span><span class="addressbook-shares__shareematch--bold">{{ props.option.matchpattern }}</span><span>{{ props.option.matchend }} {{ props.option.matchtag }}</span>
-				</div>
-			</template>
-			<span slot="noResult">{{ noResult }} </span>
-		</multiselect>
+			track-by="user"
+			label="user"
+			@search-change="findSharee"
+			@input="shareAddressbook" />
 		<!-- list of user or groups addressbook is shared with -->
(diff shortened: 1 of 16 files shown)


src/components/Settings/SettingsAddressbookShare.vue
```diff
@@ -99,75 +88,38 @@ export default {
...
-					api.get(OC.linkToOCS('cloud', 2) + 'users/details?search=' + query),
-					api.get(OC.linkToOCS('cloud', 2) + 'groups/details?search=' + query)
-				]).then(response => {
-					let matchingUsers = Object.values(response[0].data.ocs.data.users)
-					let matchingGroups = response[1].data.ocs.data.groups
-					try {
-						this.formatMatchResults(matchingUsers, query, false)
-					} catch (error) {
-						console.debug(error)
+				const results = await client.principalPropertySearchByDisplayname(query)
```
georgehrke: You have to filter the results to make sure you only get `calendarUserType` of either `INDIVIDUAL` or `GROUP`.
(Yes, all the results you see are either `INDIVIDUAL` or `GROUP`, but when using resource booking backends (which you probably are not), there will also be results of the `calendarUserType` `RESOURCE` and `ROOM`.)

skjnldsv: Oooh! Nice! Sorry I missed that :hugs:

georgehrke:
Quickly skimming the code looks good, will check in depth and test tonight.

src/store/addressbooks.js
```diff
@@ -59,7 +59,28 @@ export function mapDavCollectionToAddressbook(addressbook) {
...
+export function mapDavShareeToSharee(sharee) {
+	const id = sharee.href.split('/').slice(-1)[0]
+	const name = sharee['common-name']
+		? sharee['common-name']
+		: id
+	return {
+		displayName: name,
+		id: id,
+		writeable: sharee.access[0].endsWith('read-write'),
+		isGroup: sharee.href.indexOf('groups') > -1,
```
georgehrke: `sharee.href.indexOf('principal:principals/groups/') === 0`
Otherwise it will also be true if the username contains `groups` for whatever reason.

src/views/Contacts.vue
```diff
@@ -56,7 +56,7 @@ import ImportScreen from 'Components/ImportScreen'
 import Contact from 'Models/contact'
 import rfcProps from 'Models/rfcProps.js'
```
georgehrke: You should remove the `.js` here as well :)

skjnldsv:
@georgehrke Done :)

skjnldsv:
> Somehow the contacts app doesn't load for me :/
#content is just empty, no error during build-js-production, no error message in the console confused

Strange, try just to do a `make`

ChristophWurst:
> Somehow the contacts app doesn't load for me :/
> 
> #content is just empty, no error during `build-js-production`, no error message in the console confused

Same here

bildschirmfoto von 2018-11-12 10-59-08

skjnldsv:
Fixed :)

ChristophWurst:
- [ ] Bug? Created some contacts to test the app

bildschirmfoto von 2018-11-12 12-18-49

- [ ] Bug?

bildschirmfoto von 2018-11-12 12-18-07

What shall I actually test here? :see_no_evil:

PR #719, opened 2018-11-13
actually display Copied label for 2 seconds

georgehrke:
kapture 2018-11-13 at 21 36 31


skjnldsv:
Never stop being awesome! :+1:

PR #725, opened 2018-11-15
window.firstDay is zero based, vue2-timepickers first of of week is not

georgehrke:
`window.firstDay` is zero based, like javascript's `Date.prototype.getDay()`
vue2-timepickers's first of of week is not for whatever reason:

38365a29-7d47-4110-946d-c7f23d77675c

https://www.npmjs.com/package/vue2-datepicker


skjnldsv:
> vue2-timepickers's first of of week is not for whatever reason

Well, Javascript months starts at 0 but Datepicker starts at 1, so I'd say: 1 - 1! :grin:

PR #898, opened 2019-01-30
Add property-defined actions and populate FN if unchanged or empty 

### LINKED ISSUE #821: Displayed name field not automatically populated since vue.js
## Expected behavior
If you create a new contact and choose the "Detailed Name" property, the entered data should automatically populate the Displayed Name field as in previous versions.

## Current behavior
If you create a new contact and choose the "Detailed Name" property, the entered data is not used to populate the Displayed Name field anymore as in previous versions. The Displayed Name field stays empty and need to be manually entered too.

20190105_nc_contacts_detailed_name_not_populating_displayed_name

### Steps to reproduce
1. Open the Contacts app
2. Choose **New contact**
3. Choose **Add new property** and select **Detailed name**
4. Enter data into the **Detailed name** fields **Prefix**, **First name** and **Last name**
5. Check out the **Displaye name** field at the top of the record
-> The field is not automatically populated

#### Server Configuration
OS: Linux 3.16.50
Web server: Apache2 2.4.37
Database: MariaDB 10.2.19
PHP version: 7.2.13
Nextcloud version: 15.0.0
Contacts app version: 3.0.1

#### Client Configuration
Browser: Mozilla Firefox 64.0
Operating system: Windows 10

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68295669-displayed-name-field-not-automatically-populated-since-vue-js?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
Fix #821

| Action | Auto-fill |
|:---------:|:------:|
|peek 30-01-2019 12-27|peek 30-01-2019 12-34|

### DIFF
--- a/css/Properties/Properties.scss
+++ b/css/Properties/Properties.scss
@@ -26,8 +26,8 @@ $property-value-max-width: 250px;
 .property {
 	@include generate-grid-span(1);
 	position: relative;
-	padding-right: 44px; // delete button
-	// we need this to keep the alignment of the ext and delete button
+	padding-right: 44px; // actions menu / button
+	// we need this to keep the alignment of the ext and delete/action button
 	// The flex grow will never go over those values. Therefore we can set
 	// the max width and keep the right alignment
 	max-width: $property-label-max-width + $property-value-max-width + 44px;
@@ -37,8 +37,8 @@ $property-value-max-width: 250px;
 	&--last {
 		margin-bottom: $grid-height-unit;
 	}
-	// no delete icon on addressbook selector
-	&--addressbooks &__delete {
+	// no delete/action icon on addressbook selector
+	&--addressbooks &__actions {
 		display: none !important;
 	}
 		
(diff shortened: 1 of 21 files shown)


jancborchardt:
Nice! Some details:
- The "Detailed name" header is duplicated? The second one (which is also misaligned) seems not necessary? Then the 3-dot menu can just be on the row of the heading.
- The "Delete" action should always be the very last in a menu.
- When you initially use the detailed name to fill, it auto-fills the main name field (second gif). Couldn’t that also happen on change? The "Copy to full name" action seems like unnecessary work for users?

skjnldsv:
Hey thanks :)
1. let's see that in another issue then. I could indeed need some help on this. This design have been like that for ages :see_no_evil: 
2. :+1: 
3. Well, this was a feature that was requested to be removed a lot on the previous versions. Basically in the vcard you can set the display Name and the detailed name. Those can and need to be completely independant. I can rename my wife `Wife 🥰` but still wants to keep the detailed name informations.
  This is why we have an automated step (that use the detailed name to fill the display name in some cases only:
  1. the displayName is not set
  2. the displayName is empty
  3. the displayName is still set to the default `New Contact`
  BUT if you manually want to have the same you can use the copy action. (or not use a detailed name property at all, this is allowed)

PR #945, opened 2019-02-22
Fix spelling mistake in error message

raoulbhatia:
Wrong: `The contact doesn\'t exists anymore on the server.`
Should: `The contact doesn\'t exist anymore on the server.`

This will require adaption of the translation source string.


skjnldsv:
Thanks :)

PR #1018, opened 2019-03-26
Use virtual scroller for big contacts list display performances

skjnldsv:
@fwolfst could you try this build? It should drastically improve the rendering and usability of the contacts list with lots of entries :)
[contacts.tar.gz](https://github.com/nextcloud/contacts/files/3009159/contacts.tar.gz)

### DIFF
--- a/css/ContactsList.scss
+++ b/css/ContactsList.scss
@@ -33,3 +33,14 @@
 	transform: rotate(180deg);
 	margin-top: 44px; // under the show navigation button
 }
+
+
+// Virtual scroller overrides
+.vue-recycle-scroller{
+	position: sticky !important;
+}
+
+.vue-recycle-scroller__item-view {
+	// same as app-content-list-item
+	height: 68px;
+}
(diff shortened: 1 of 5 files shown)


fwolfst:
Unfortunately I dont have a staging copy up yet and if I throw it in the production setup it wants me to run updates. I guess there is some trick to overcome it (fake the version number?). Or does it indeed need to run an update step due to the changed js-package dependencies? If so, I might not be able to do a quick test within the next dozens of hours - i unfortunately do not have an automatic staging setup-setup.

skjnldsv:
> Unfortunately I dont have a staging copy up yet and if I throw it in the production setup it wants me to run updates. I guess there is some trick to overcome it (fake the version number?). Or does it indeed need to run an update step due to the changed js-package dependencies? If so, I might not be able to do a quick test within the next dozens of hours - i unfortunately do not have an automatic staging setup-setup.

I indeed increased he version number to trigger a cache clear (harmless, but this is usually better to make sure cache issues are out of the way)
If this is a production instance and you're not the only one on it, I'd advise against testing it then.
If you're alone, it won't change anything nor trigger database changes. You can edit the appinfo/info.xml file and set it back to 3.0.5 if you don't want to trigger an update :)

PR #1024, opened 2019-03-29
Allow to pick avatar from files + use modal

### LINKED ISSUE #106: Allow assignment of images already in Nextcloud as contact photos

skjnldsv:
Fix #106
Capture d’écran_2019-03-29_10-55-09
Capture d’écran_2019-03-29_10-55-27

@jancborchardt

### DIFF
--- a/css/ContactDetailsAvatar.scss
+++ b/css/ContactDetailsAvatar.scss
@@ -84,9 +84,16 @@
 	}
 }
 
-.contact-header-modal__photo {
+.contact-header-modal {
 	.modal-container {
-		display: flex;
+		display: flex !important;
+		align-items: center;
+		justify-content: center;
+		.contact-header-modal__photo {
+			// animate zooming/resize
+			transition: height 100ms ease,
+				width 100ms ease;
+		}
 	}
 }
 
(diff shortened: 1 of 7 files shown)


jancborchardt:
@skjnldsv this one was hiding somewhere :D sorry for late reply:

- Loooks soooo great! Much better than the icons without labels we had before. You rock!
- The popovers should be centered (for the one in the viewer, should be to the right as far as possible with some whitespace, just like the user dropdown is). :) The left-alignment looks strange. ;)

skjnldsv:
@jancborchardt I think I remember, but did we not wanted to actually set the popover to center as default? :thinking:

jancborchardt:
@skjnldsv yes, you’re right! It definitely also seems that one of my most given feedbacks on popovers were to center them. :D So seems a good call to make that default.

skjnldsv:
Let's do that on the components then :)

PR #1054, opened 2019-04-09
UI: Contact List: fix delete-icon position

rponline:
| Before | After |
|---|---|
| image | image |

### DIFF
--- a/css/ContactsListItem.scss
+++ b/css/ContactsListItem.scss
@@ -24,6 +24,10 @@
 	transition: left 200ms;
 }
 
+.app-content-list-item-text {
+    flex-grow: 1;
+}
+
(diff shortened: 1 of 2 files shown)


skjnldsv:
Sorry, this design is not meant to be changed on contacts. :)
This is part of our design standards: https://docs.nextcloud.com/server/latest/developer_manual/design/list.html

rponline:
Hmm, ok, still feels wrong from a visual perspective, but thank you for pointing out the design standards!

skjnldsv:
@rponline you can still suggest applying the fix to nextcloud/server :)
But it should not require a html layout change. Css tricks only :grin:

PR #1055, opened 2019-04-09
Avoid qrcode getting cropped

rponline:
increase image width for better scanning

remove height to avoid white padding over and under qrcode
(modal box uses the height for content size, but css resize gets ignored)

| Before | After |
|---|---|
| Screen Shot 2019-04-09 at 15 29 27 | Screen Shot 2019-04-09 at 15 28 46 |


skjnldsv:
Thanks for all you've been doing lately @rponline :hugs:

PR #1264, opened 2019-09-14
Add TZ & LANG

### LINKED ISSUE #1048: Support Timezone and Language
Currently, the following two properties, specified by both vCard 3 and vCard 4 are not exposed:

**Language** 'LANG' property
Which sets the language(s) used by the contact.

**Timezone** 'TZ' property
Which sets the usual timezone of a contact.

This is particularly useful for the many Users who have international friends or colleagues to ensure contact is timely.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72558684-support-timezone-and-language?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
Fix #1048


skjnldsv:
Please review :D

skjnldsv:
ping !

georgehrke:
My request to ask for TZ and LANG was a bit selfish, because i would like to use it for the calendar app. Whenever selecting a contact as an attendee, the invitation email should automatically use the contacts language and also show the time and the attendees' local timezone.

As i already mentioned at Nc conf, i would just offer the same list of languages we do in the server.
If we allow the user to enter a language, they will most likely enter stuff like "Deutsch" or "Français", which will be a tremendous pain for us to parse and map to the list of languages that we ship.

Offering that list of languages will provide a big advantage and i personally don't see the need to offer the user to enter every silly dialect that's spoken in whatever village with 5 people living there.

skjnldsv:
Okay :)
Shall I auto match the first prefix of a language tag?

If it's set to fr-ca-xx-xxx-x-x-x, I match  with fr?
But if you change it would just set 'fr' !

georgehrke:
Sounds like a good plan to me :)

skjnldsv:
@tcitworld we should maybe update our server list with transifex data?
https://github.com/transifex/transifex/blob/master/transifex/languages/fixtures/all_languages.json

tcitworld:
Server has locales list (we need the differences for number and time formatting for instance), these one contain only languages.

skjnldsv:
I don't get your message @tcitworl :p

skjnldsv:
Done @georgehrke :)
I need to create a model for each rfc object so I can document and be a little more clean, but that will do for now!

associated matchings:
``` php
fr			=> fr
fr-fr		=> fr-fr
fr_fr		=> fr-fr
fr-fr-*		=> fr
```

tcitworld:
> I don't get your message @tcitworld :p

My bad, didn't get a good look at the Transifex file.

PR #1307, opened 2019-10-17
Make the work profile the default one when adding Contacts

GretaD:


lib/Controller/PageController.php
```diff
@@ -38,16 +39,20 @@ class PageController extends Controller {
 
 	/** @var IFactory */
 	private $languageFactory;
+	/** @var IConfig */
+	private  $config;
 
 	public function __construct(string $AppName,
 								IRequest $request,
+								IConfig $config,
```
nickvergessen: > 1) OCA\Contacts\Controller\PageControllerTest::testIndex
> 
> TypeError: Argument 3 passed to OCA\Contacts\Controller\PageController::__construct() must implement interface OCP\IConfig, instance of Mock_IFactory_a554ddad given, called in /home/travis/build/nextcloud/core/apps/contacts/tests/unit/Controller/PageControllerTest.php on line 58

You need to adjust the test aswell

GretaD: right! :)

src/models/rfcProps.js
```diff
@@ -27,6 +27,7 @@ import ActionCopyNtoFN from '../components/Actions/ActionCopyNtoFN'
 import ActionToggleYear from '../components/Actions/ActionToggleYear'
 import zones from './zones'
 
+const defaultProfileState = loadState('contacts', 'defaultProfile')
```
skjnldsv: Could you add a little comment here to explain what we're doing, so that anyone reading this will understand what it's exactly :)

nickvergessen:
Phone is still defaulting to `HOME`
https://github.com/nextcloud/contacts/blob/cd2e1edef4632fc40542a730e84f9ac26f995d50/src/models/rfcProps.js#L185-L188

src/models/rfcProps.js
```diff
@@ -182,7 +184,7 @@ const properties = {
 		default: true,
 		defaultValue: {
 			value: '',
-			type: ['HOME', 'VOICE']
+			type: [defaultProfileState]
```
nickvergessen: Not the expert, but maybe:
```suggestion
			type: [defaultProfileState, 'VOICE']
```

GretaD: right, my bad :)

skjnldsv: This is tricky, but yes!
Though only `WORK,VOICE` and `HOME,VOICE` are really commonly used here. :)
Should be good for now!

PR #1393, opened 2020-01-04
Use correct syntax for vCard version when saving

### LINKED ISSUE #1312: contact photo not synced due to changed PHOTO value
## Expected behavior
An uploaded contact photo should be stored so that it can correctly been synced e.g. with DAVx5.

## Current behavior
Newly uploaded contact photos are not synced anymore, due to a different syntax used for the PHOTO value. Unfortunately I don't know if DAVx5 is lacking support of this syntax or if it is a problem within Nextcloud.

(1) - Working syntax: `PHOTO;ENCODING=b;TYPE=JPEG:/9j/4AAQSkZJRgA ...`

(2) - Failing syntax: `PHOTO;ENCODING=b;TYPE=jpeg;VALUE=URI:/9j/4AAQSkZJRgAB ...`

## Steps to reproduce
1. Open the contacts app
2. select a contact and upload a photo.
3. check the used CardDAV syntax by exporting a vcf file.
4. force a contact sync in DAVx5
**-> the photo is not synced (2)**

### Cross-check
1. Open Thunderbird (v68.2.0) with activated CardBook (v42.9) add-in
2. select a contact and upload a photo.
3. check the used CardDAV syntax by exporting a vcf file from within the Nextcloud contacts app
4. force a contact sync in DAVx5
**-> the photo is synced (1)**

#### Server Configuration
OS: Linux 3.16.74
Web server: Apache2 2.4.41
Database: MariaDB 10.2.24
PHP version: 7.3.10
Nextcloud version: 17.0.0
Contacts app version: 3.1.6

#### Client Configuration
Browser: Mozilla Firefox 70.0
Operating system: Windows 10

hanzi:
This changes the API used for serialising the vCard object into a string.

The previous method called `ical.stringify` on the underlying jCal object directly, which would always use vCard 4.x syntax regardless of the actual version used in the object.

If the vCard was still in 3.0 format, that would lead to invalid syntax being used. While clients seem to tolerate this most of the time, it breaks DAVx5 photo import.

Here's how the new call is being implemented in ical.js:
https://github.com/mozilla-comm/ical.js/blob/master/lib/ical/component.js#L487

fixes #1312

### DIFF
--- a/src/services/updateDesignSet.js
+++ b/src/services/updateDesignSet.js
@@ -49,6 +49,24 @@ const setLabelAsSingleValue = () => {
 }
 
 /**
+ * Prevents ical.js from adding 'VALUE=PHONE-NUMBER' in vCard 3.0.
+ * While not wrong according to the RFC, there's a bug in sabreio/vobject (used
+ * by Nextcloud Server) that prevents saving vCards with this parameters.
+ *
+ * @link https://github.com/nextcloud/contacts/pull/1393#issuecomment-570945735
+ *
+ * @returns {Boolean} Whether or not the design set has been altered.
+ */
+const removePhoneNumberValueType = () => {
+	if (ICAL.design.vcard3.property.tel) {
+		delete ICAL.design.vcard3.property.tel
+		return true
+	}
+
+	return false
+}
+
+/**
  * Some clients group properties by naming them something like 'ITEM1.URL'.
  * These should be treated the same as their original (i.e. 'URL' in this
  * example), so we iterate through the vCard to find these properties and
@@ -82,6 +100,7 @@ export default function(vCard) {
 	let madeChanges = false
 
 	madeChanges |= setLabelAsSingleValue()
+	madeChanges |= removePhoneNumberValueType()
 	madeChanges |= addGroupedProperties(vCard)
 
 	return madeChanges
--- a/src/store/contacts.js
+++ b/src/store/contacts.js
@@ -330,7 +330,7 @@ const actions = {
 		rev.fromUnixTime(Date.now() / 1000)
 		contact.rev = rev
 
-		const vData = ICAL.stringify(contact.vCard.jCal)
+		const vData = contact.vCard.toString()
 
 		// if no dav key, contact does not exists on server
 		if (!contact.dav) {


hanzi:
Ok, this is annoying. While testing something else, I noticed an issue with this fix:

When saving a vCard in 3.0 format, ical.js adds `VALUE=PHONE-NUMBER` to `TEL` properties, at least sometimes. So a phone number would look like this:

```
TEL;TYPE=HOME;VALUE=PHONE-NUMBER:+123456789
```

According to the RFC this is correct, but our DAV backend doesn't recognise this value type and throws an error instead. I filed a PR for that (sabre-io/vobject#486) but even if this was merged, tagged and included in Nextcloud as soon as possible, older Nextcloud versions would still be broken. From what I understand, the Contacts app could be updated independently from Nextcloud, so that's a no-no.

@skjnldsv What would be your preferred way to deal with this?

The issue is with this line here: https://github.com/mozilla-comm/ical.js/blob/master/lib/ical/design.js#L830

The RFC doesn't actually require the value to be explicitely set, so if we patch the vCard 3 design set simliarly to what I've done in #1394 (something like `delete ICAL.design.vcard3.property.tel` in this case) would work around this problem.

An alternative would be to build a dedicated check that fixes these cases by explicitely making the value type `URI`. That doesn't actually conform to the vCard 3.0, but that's what we've been generating so far anyway and it seems to work with all relevant clients.

skjnldsv:
Wow, well done!!
I was always wondering why the heck the lib was behaving this way!!! :hugs: 

I couldn't figure it out and ran out ot time for contacts with other projects :)


> According to the RFC this is correct, but our DAV backend doesn't recognise this value type and throws an error instead. I filed a PR for that ([sabre-io/vobject#486](https://github.com/sabre-io/vobject/pull/486)) but even if this was merged, tagged and included in Nextcloud as soon as possible, older Nextcloud versions would still be broken

We can backport it then.
Let's add this and make our backend properly compliant :)
Alternatively, we could also indeed patch the lib.
I'm not sure `VALUE=PHONE-NUMBER` adds anything to our vcards. Especially considering no other vcard manager does it (apple, android, outlook, windows...)

So let's still add this to our backend for compatibility, 
But patch the lib like you suggested.
(But I would create a dedicated js file for all those patches, not in the contacts model. This would greatly help understanding the code) :+1:

skjnldsv:
And thanks a lot @hanzi! This is awesome work!! :D

hanzi:
> We can backport it then.
> Let's add this and make our backend properly compliant :)

I agree, that would be preferable.

But since sabre-io/vobject is a dependency of nextcloud/server we cannot rely on the backend having been fixed already: The user might use a fixed version of Contacts with an older version of Nextcloud Server. So this workaround is still necessary.

Anyway, I included the design set fix. Had no further issues with phone numbers while testing.

Issue #1376, opened 2019-12-20
Missing N field in new contacts


sciurius:
### Describe the bug

When a new contact is created, you can fill in the name in the title bar. This becomes the FN field.

Many external applications however rely on the content of the N field, even though it is formally optional. To get an N field, the user must explicitly engage the "Detailed Name" property and then fill the fields.

I'd propose to add a setting to automatically add the "Detailed Name" property upon contact creation. Filling it in takes as little time as filling in the name in the title bar. When N is filled, infer FN from N.

Alternatively, but error prone, infer N from FN.

As stated, many external applications rely on the content of the N field so I decided to label this a bug and not an enhancement. Feel free to disagree.

skjnldsv:
Hello!

Thanks for your suggestion :)
We actually had this suggestion many times before, but this is unfortunately a very tricky one.
You have to support lots of language to properly parse a name and split it into the appropriates values.

We implemented something similar into the old version, but this was too simple and very unpredictable.

Though, we have the option to copy from N to FN.
If you check the N data menu (three dot icon) there is an option to concatenate the data into the FN field.

Showing the detailed name by default would be one way to get the user to maybe fill it more often. @jancborchardt what do you think?

sciurius:
Even though RFC6350 allows N to be empty, it states that 

> The property SHOULD be present in the vCard object when the name of the object the vCard represents follows the X.520 model.

(Which AFAIK, Nextcloud Contacts seems to do)

I do not know any other contact app that does not primarily use the detailed name info to enter new contacts. So I strongly suggest to present the detailed name fields by default, or make it selectable via a setting.

skjnldsv:
> I do not know any other contact app that does not primarily use the detailed name info to enter new contacts

Android :)

jancborchardt:
Considering the issues that are always coming up with this and since we don’t have the resources to throw at this that Google has, I’d say we should simply split it into "First name" and "Last name".
Google Contacts (web version) and iOS Contacts do this as well.

So from our current header:
![heading](https://user-images.githubusercontent.com/925062/71572301-ed5d3080-2b10-11ea-8e8f-40cfb6a0b520.png)

To this:
![First name last name](https://user-images.githubusercontent.com/925062/71572300-ecc49a00-2b10-11ea-8167-eab83fb19e4b.png)

What do you think @skjnldsv?

sciurius:
If first name and last name are going to be put in the N field, great!

[...]

PR #1412, opened 2020-01-16
short company field only if empty

myrho:
Long company names were chopped off way to much. This fixes this while avoiding spread input fields when org field is empty.

### DIFF
--- a/css/ContactDetails.scss
+++ b/css/ContactDetails.scss
@@ -60,7 +60,7 @@
 					color: #fff !important;
 				}
 			}
-			#contact-org {
+			#contact-org:placeholder-shown {
 				max-width: 20%;
 			}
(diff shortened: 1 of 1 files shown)


skjnldsv:
Hey! :)
Welcome :)
Could you provide screenshots as well so that others can review a bit better?

myrho:
### Before this pull request:

**Before editing:**
image

**After editing:**

image


### After this pull request:

**Before editing:**
image

**After editing:**

image

jancborchardt:
Looks like a good fix, thanks @myrho! :) Note that this might become obsolete though if we do go ahead with the proposal at https://github.com/nextcloud/contacts/issues/1376#issuecomment-569604269

PR #1416, opened 2020-01-17
Fix issue that prevented new contacts from being saved

hanzi:
Whenever I would compile the JavaScript sources myself, I would run into an issue when creating a new contact:

1. I would create a new contact and fill out some of its fields.
2. The contact wouldn't save, the exclamation mark icon would just stay. If I reload the page at this point, the new contact is gone.
3. If I then selected another contact and returned to the new one, all my previous changes were gone. But if I filled out some fields now, the contact finally got saved.

I can't reproduce this bug with the released version -- though it would still happen if I compiled the JavaScript from the `v3.1.7`, `v3.1.6` or `v3.1.5` tags myself. So I'm not sure whether this might be an issue with some dependency or I'm just doing something wrong.

Anyway, the issue as far as I could tell is that the `contact` watcher of the `ContactDetails` component fires _twice_ when a new contact is added: once with the new contact and once with the old one. This eventually leads to `localContact` being overwritten with the old one, so that all save actions actually just save the old contact.

This PR fixes this behaviour by adding a check whether the new contact (inside the watcher) is actually different from the old one. I don't really like this fix because it only attacks the symptom and I still don't have any idea why the watcher is fired twice to begin with.

But at least this way the app is usable again for me.

### DIFF
--- a/src/components/ContactDetails.vue
+++ b/src/components/ContactDetails.vue
@@ -401,8 +401,8 @@ export default {
 	},
 
 	watch: {
-		contact: function() {
-			if (this.contactKey) {
+		contact: function(newContact, oldContact) {
+			if (this.contactKey && newContact !== oldContact) {
 				this.selectContact(this.contactKey)
 			}
 		},


DBJRdev:
We encounter the same issue as hanzi wrote, but for us it only began when updating to Contacts 3.1.8 from the GUI within Nextcloud 17.0.2 - no compiling or building anything ourselves. It actually appeared in three of our deployments, all managed servers and one shared hosting at the provider Hetzner.

hanzi:
@DBJRdev That's odd -- this fix is already included in 3.1.8 and I don't have this issue anymore with Contacts 3.1.8 on Nextcloud 17.0.3.

Could you [open a new issue](https://github.com/nextcloud/contacts/issues/new/choose) with some more details (including the browser log)?

DBJRdev:
@hanzi Did as you requested: #1452

skjnldsv:
Hey!
So this is actually the issue that created #1453 

Because on the first save to server, we do not update the real full contact anymore. 

The `selectContact` method also include a proper update of the `localContact`
All the changes are queued when editing a contact to avoid conflict. The first contact sent init the `contact.dav` property which is the dav library method object. Because we do not update the contact after the first sync anymore, the localContact used never have the `ðav` object, and therefore trigger a new vcard creation every time.

Fix incoming

Issue #1395, opened 2020-01-04
RELATED property incorrectly formatted - sync fails


olantrust:
ISSUE

Create a contact in Nextcloud Contacts add a 'related' field set, and enter a name or other free text:

E.g.
Mary J Blige
Sister: Theresa May

Using DavX, the related field will not sync to Android devices (and other systems) and will be lost.

Create the same field on an Android device, and the related field will correctly display on Nextcloud.

CAUSE:

https://tools.ietf.org/html/rfc6350#page-42
Specifies that the RELATED property can either contain a URI value or a TEXT value.
The default is URI. If a TEXT value is entered then it must be specified.

On Android when a free text value is entered in the RELATED field, it creates a vCard record like this:
RELATED;TYPE=sister;VALUE=TEXT:Theresa May

In Nextcloud contacts when a free text value is entered it creates a vCard record like this:
RELATED;TYPE=SISTER:Teresa May

In accordance with the RFC, since no TEXT value is declared, and since 'Teresa May' is not a valid URI, it is an error and is not synced.

SOLUTION
If free text is entered, specify 'VALUE=TEXT'

olantrust:
Also there should be a picker for creating a URI link to a contact in the addressbook, #669 seems to relate to this.

skjnldsv:
Closing in favour of #669

olantrust:
This does not appear to be the same as #669 , @skjnldsv - could you consider reopening or explaining how that bug covers this one - in case I've missed something?
This current issue is a bug about data loss occurring in the current implementation when syncing from Nextcloud to Android (via DavX). The suggested replacement issue is a proposed enhancement to allow direct links between contacts ...
Many thanks

skjnldsv:
#669 is because we force URL as default. Thus why is your issue appearing.
Despite the text you're entering, it is seen as an url (invalid sure, but parsed as is nonetheless)

Having a proper Picker to automatically fill the url and parse any free-typed text as, well, TEXT, is all in the other issue :)

olantrust:
> #669 is because we force URL as default. Thus why is your issue appearing.
> Despite the text you're entering, it is seen as an url (invalid sure, but parsed as is nonetheless)
> 
> Having a proper Picker to automatically fill the url and parse any free-typed text as, well, TEXT, is all in the other issue :)

Thank you for explaining and addressing

olantrust:
[...]

Issue #1324, opened 2019-11-01
Location field tells me to enter a URL.


RobertZenz:
The new Location field is marked red and tells me to enter a URL, even when it is prefilled with coordinates.

 1. Open a contact.
 2. Add the Location field.
 3. It's red already, even though it was just prefilled with the default.

Additionally, the default value uses a semicolon `;` to separate the coordinates, but after saving and loading it has changed to a comma `,`.

### Server configuration

**Operating system**: centOS 7

**Web server:** lighttpd

**Database:** SQLite

**PHP version:** 7.2

**Nextcloud version:** 16

**Contacts version:** 3.1.6

jancborchardt:
@RobertZenz can you describe exactly what would be expected? Then this would be a good first issue. :)

skjnldsv:
@jancborchardt the dav library understand location as an url, and this is forwarded to the html with an input[type=url]. 
We need this as a text, but weshould also add a plugin where you can autocomplete locations/pick on map

olantrust:
Note, a similar scenario occurs with RELATED field (bug 1395) where the RFC requires either a URL or a *declared* text, but Nextcloud only likes a URL, with no picker.

hanzi:
To be clear: the RFC actually requires some kind of URI scheme here, see [section 6.5.2](https://tools.ietf.org/html/rfc6350#section-6.5.2). So this requirement is not actually wrong.

But I agree that there should be a more helpful and user-friendly widget for editing this property. Does anyone know how this property is actually being used in other clients?

PR #1418, opened 2020-01-18
Disable form validation for text properties

### LINKED ISSUE #1214: Skype login is proclaimed to be wrong
**To Reproduce**
1. add an instant messenger
2. select "Skype" as a type
3. enter a plain login name, e.g. echo.echo (this is the skype echo bot)

**Expected behavior**
I can either enter an email-like contact (MSN-like scheme) or a plain login (old Skype scheme). Both will be accepted as valid.

**Actual behavior**
The field falsely claimed that this is invalid as an URL is expected:
skype-Login

**Screenshots**
see above

I think, all this technical information won't be necessary as this is rather a conceptual but less a technical bug.

hanzi:
fixes #1214 
relates to #1324 
relates to #1395 

Currently, we mark some input fields as `type="url"`, `type="tel"` and `type="email"` depending on the property name and sometimes depending on what ical.js expects those fields to contain.

This triggers form validation in some browsers which will then highlight the field as erroneous if the contents don't match the type.

Reasons against this behaviour:

1. Some affected fields such as `RELATED` or `IMPP` might genuinely need to contain values that do not match the usual URL pattern. (For `RELATED` the RFC even explicitely allows arbitrary text values, for `IMPP` see #1214)
2. Other clients such as Android's Contacts app allow arbitrary values here, which Nextcloud Server happily accepts and which would then be flagged as erroneous in Nextcloud Contacts.
3. Most importantly: While this field is marked red in the UI, we _still_ save the value anyway. So there's a discrepancy between what Contacts actually does and what it displays to the user.

In order to keep specialised keyboards in mobile browsers, I used the `inputmode` attribute instead.

### DIFF
--- a/src/components/Properties/PropertyText.vue
+++ b/src/components/Properties/PropertyText.vue
@@ -58,7 +58,6 @@
 				id="textarea"
 				ref="textarea"
 				v-model.trim="localValue"
-				:type="type"
 				:inputmode="inputmode"
 				:readonly="isReadOnly"
 				class="property__value"
@@ -69,10 +68,10 @@
 			<!-- OR default to input -->
 			<input v-else
 				v-model.trim="localValue"
-				:type="type"
 				:inputmode="inputmode"
 				:readonly="isReadOnly"
 				:class="{'property__value--with-ext': haveExtHandler}"
+				type="text"
 				class="property__value"
 				@input="updateValue">
 
@@ -139,19 +138,15 @@ export default {
 			// length is one & add one space at the end
 			return hasTitle + 1 + isLast + noteHeight
 		},
-		type() {
+		inputmode() {
 			if (this.propName === 'tel') {
 				return 'tel'
 			} else if (this.propName === 'email') {
 				return 'email'
-			}
-			return 'text'
-		},
-		inputmode() {
-			if (this.propName === 'uri') {
+			} else if (this.propType === 'uri') {
 				return 'url'
 			}
-			return ''
+			return false
 		},
 		URLScheme() {
 			if (this.propName === 'tel') {
--- a/src/components/Properties/PropertyText.vue
+++ b/src/components/Properties/PropertyText.vue
@@ -59,6 +59,7 @@
 				ref="textarea"
 				v-model.trim="localValue"
 				:type="type"
+				:inputmode="inputmode"
 				:readonly="isReadOnly"
 				class="property__value"
 				@input="updateValueNoDebounce"
@@ -69,6 +70,7 @@
 			<input v-else
 				v-model.trim="localValue"
 				:type="type"
+				:inputmode="inputmode"
 				:readonly="isReadOnly"
 				:class="{'property__value--with-ext': haveExtHandler}"
 				class="property__value"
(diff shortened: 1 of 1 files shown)


skjnldsv:
Thanks for your awesome work!!

You can also just override types in the rfcProps file. I still think that having email, URL.. is an improvement. But for some properties this is indeed an issue :)

hanzi:
I thought about that as well, but ultimately decided against doing it that way.

I don't think that it's necessarily the app's job to validate user input at this point. Nothing really relies on this data, it's more of a fancy notebook. So if the user wants to enter seemingly invalid data -- why should we stop them?

skjnldsv:
> I don't think that it's necessarily the app's job to validate user input at this point. Nothing really relies on this data, it's more of a fancy notebook. So if the user wants to enter seemingly invalid data -- why should we stop them?

It's still giving proper feedback to the user.
The priority should always to be compliant to the browser's standards as this is what is available and used on all website, we don't create our own check and tooltip implementations for that reason :)


> So if the user wants to enter seemingly invalid data -- why should we stop them?

Because then we get corrupted vcards or the icaljs library reject the changes and we get user reporting the app doesn't work ;)

hanzi:
The thing is that the validation **doesn't** prevent the saving of any invalid cards. Saving still occurs because we're not actually using the `submit` handler or even an HTML form to begin with.

Also, for none of the affected fields were there any issues with either ical.js handling the card or the server accepting it during my tests. I agree that it would be nice to have better feedback if there's an error with ical.js/server but the form validation doesn't do that either. It just adds another layer that might (and does) give false positives.

So at the very least, it's misleading UI: A red outline would usually mean 'cannot submit this' to a user, but we're actually submitting it without asking.

And it's bad UX: Form validation should hint at wrong input, when in fact most of the time this will probably highlight _data already coming from the server_.

skjnldsv:
> > So if the user wants to enter seemingly invalid data -- why should we stop them?


> The thing is that the validation **doesn't** prevent the saving of any invalid cards. Saving still occurs because we're not actually using the `submit` handler or even an HTML form to begin with.

Then let's leave it like that?
You said why stop them, but this is not stopping, this is a warning. @jancborchardt please decide ;)

jancborchardt:
> And it's bad UX: Form validation should hint at wrong input, when in fact most of the time this will probably highlight _data already coming from the server_. (Because as I said, other clients like Android or Outlook will let you fill these forms with whatever you want.)

This seems the strongest argument for changing it: Data you already have might be highlighted, _and_ people are used to put everything in from other platforms. :)

Then again, the type of the fields also triggers different keyboards like a number pad for `type=tel`. This is especially useful on touch devices and mobile.

Is it possible to have the benefit of that without the hassle of unneeded inline validation?

skjnldsv:
> Is it possible to have the benefit of that without the hassle of unneeded inline validation?

I don't think so.

hanzi:
> Then again, the type of the fields also triggers different keyboards like a number pad for type=tel. This is especially useful on touch devices and mobile.
>
> Is it possible to have the benefit of that without the hassle of unneeded inline validation?

Yeah, and it's already part of this PR: Using the [inputmode attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode). Basically the same as using `type` but without the validation.

skjnldsv:
> Yeah, and it's already part of this PR: Using the [inputmode attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode). Basically the same as using `type` but without the validation.

Then let's get this in!

skjnldsv:
> Then let's get this in!

Hum, actually no.
Capture d’écran_2020-01-23_10-45-51

hanzi:
Well, support on desktop browsers isn't really that important for on-screen keyboards, is it?
All relevant mobile browsers do support it.

skjnldsv:
Fair enough.

PR #1430, opened 2020-01-19
Remove delete action from list, fix #1368

### LINKED ISSUE #1368: Too easy to delete contact
Currently it is very easy to delete a contact from the web backend. - just by clicking on the trash can next to the contact name in the contact list. The risk is that contacts are deleted by mistake. 

As it is rather the exception that a contact is being deleted from the contact list, I would prefer if a contact could be deleted only by clicking first on the contact itself and then on the three dots in the upper right corner (and finally confirming by deleting contact).

**Additional context**
80_nextclouddeletecontact2

80_nextclouddeletecontact

There's a link to another issue raised earlier: https://github.com/nextcloud/server/issues/18181

jancborchardt:
Simply removes the delete icons in the contact list as mentioned by the issue at https://github.com/nextcloud/contacts/issues/1368 and illustrated in the screenshot. :)

Deleting a contact is still possibly through the 3-dot menu in the detail view.

Please review @skjnldsv @hanzi @ralxxx

### DIFF
--- a/css/ContactsListItem.scss
+++ b/css/ContactsListItem.scss
@@ -19,38 +19,6 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
-// ensure transition for delete and non-deleted transition state
-.app-content-list-item {
-	transition: left 200ms;
-}
-
-.app-content-list-item.deleted {
-	z-index: 2;
-	// make sure we hide the undeleted element by sliding over it
-	// and force the background even on hover
-	background-color: var(--color-main-background) !important;
-
-	.app-content-list-item-icon__avatar {
-		filter: brightness(.8) grayscale(1);
-	}
-	.app-content-list-item-line-one {
-		text-decoration: line-through;
-	}
-
-	// vuejs animation
-	&.delete-slide-left-enter-active,
-	&.delete-slide-left-leave-active {
-		position: absolute;
-		top: 0;
-		left: 0;
-		width: 100%;
-	}
-
-	&.delete-slide-left-enter,
-	&.delete-slide-left-leave-to {
-		left: 100%;
-	}
-}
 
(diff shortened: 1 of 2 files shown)


hanzi:
Another thing: There's still some related CSS code in `css/ContactsListItem.scss` which should also be removed.

jancborchardt:
> Another thing: There's still some related CSS code in `css/ContactsListItem.scss` which should also be removed.

Thanks for the catch – fixed. :)


> But I'm a bit worried that the delete button might be hidden away a bit too much now. That 3-dot menu is far away from everything else. Not sure if it's that obvious to find.

Agree that the 3-dot-menu is too far away. This is caused by the fact that we go wide into oblivion. We should also restrict the width, just like we do for Mail ( 
https://github.com/nextcloud/mail/pull/2180 ) and Talk.
But this is something for a separate pull request. :)

> The Mail app (which has the same layout structure) has a second instance of the 3-dot menu in the mail list. Maybe that's an option for Contacts as well?

While it has the same layout, the big difference between Mail and Contacts is that Mail is an active app – you are likely in there every day, and working a lot with the entries. Contacts is more of a passive app, where you edit stuff once in a while, and don’t delete as often as you do mails. So I’d say there’s no need for the 3-dot menu in the contact list.

hanzi:
> While it has the same layout, the big difference between Mail and Contacts is that Mail is an active app – you are likely in there every day, and working a lot with the entries. Contacts is more of a passive app, where you edit stuff once in a while, and don’t delete as often as you do mails. So I’d say there’s no need for the 3-dot menu in the contact list.

Valid point, but personally I'd be in favour of having consistent patterns across theses apps. If it looks the same, it should work the same.

But you're right, this is probably better suited as a task of its own.

PR #1438, opened 2020-01-24
Allow importing from files

skjnldsv:
image

- [x] Modal
- [x] Design
- [x] Cancel should kill axios request if any
- [x] Jan's approval & changes
- [x] Cleanup old code


src/components/Settings/SettingsImportContacts.vue
```diff
@@ -23,21 +23,87 @@
...
+					</button>
+					<button
+						:class="{'icon-loading': loading}"
+						:disabled="loading"
+						class="button import-contact__button import-contact__button--files"
+						@click="openPicker">
+						<span class="import-contact__button-icon icon-folder" />
+						{{ t('contacts', 'Import from files') }}
+					</button>
+					<!-- <button id="cancel" class="button import-contact__button import-contact__button--cancel" @click="toggleModal">
```
skjnldsv: Not sure about this one, Jan, any thoughts?
Maybe not as big and a simple `Cancel` link under the buttons on the bottom right would be enough?

jancborchardt:
- If there’s only one address book, we don’t need to mention the name. Just saying "Import contacts" as header is enough (no "-ing" if it’s not in progress)
- The buttons are a bit goofily large, no? Just having them with the icon left next to them is enough.
- One of them should be primary style – your call

skjnldsv:
Thanks @jancborchardt :)

skjnldsv:
All good!
Please review ! :tada:

Issue #1445, opened 2020-01-28
Unclear which format to use for social media profiles


jancborchardt:
In Contacts, for `x-socialprofile` the type `text` is forced, but in the info section it says `The url of the profile. e.g. https://www.facebook.com/Nextclouders/`. So should the type not be "uri"?

And then to make it easier, whenever you add a social property, we should prefill the start of the URL like `https://facebook.com/` and focus it after that.
This works even with federated systems like Mastodon, Diaspora and Nextcloud social as we can simply have a default URL like `https://joindiaspora.com/` and `https://mastodon.social/` and it can be adjusted then.

skjnldsv:
Have you tried changing it to url?
Or is this issue only about the placeholder? :)

call-me-matt:
I would say most people will input the usernames only. For twitter, it is quite common to use @twittername, for example. And when you see advertisements, normally you see a reference with facebook-logo plus the respective username. So I would not insist on putting an URL here. Also, it takes up a lot of space with redundant information to the type field, displayed just in front of it.

skjnldsv:
Fixed with placeholders

PR #1446, opened 2020-01-28
Simplify form

jancborchardt:
- Hide delete icons initially to not overflow interface
- Remove unneeded info icons and tooltips
- Work in progress of placeholders for social profiles 

Need help on displaying placeholders for social profiles, see also https://github.com/nextcloud/contacts/issues/1445


jancborchardt:
I need help for the last part :)

> Need help on displaying placeholders for social profiles, see also #1445

src/models/rfcProps.js
```diff
@@ -261,10 +252,6 @@ const properties = {
 		multiple: true,
 		readableName: t('contacts', 'Related'),
 		icon: 'icon-group',
-		info: t(
-			'contacts',
-			'Specify a relationship between another entity and the entity represented by this vCard.'
```
skjnldsv: It was actually added because lots of people are confused about what some properties really mean. In one case it's easy to understand for the `socialprofile` one, but for the difference between related and relationship, this still needs clarification (gotta love vcard, right?)

jancborchardt: Do other contact apps show both? Maybe they just change the wording – "Relationship" for one, and some other word for whatever "Related" means. Relationship can have suggestions, could Related as well?

skjnldsv: > Do other contact apps show both

Yes, they are different properties.

> Maybe they just change the wording

Sure, what do you suggest we use instead? Please remember we do not invent this and this is coming straight from the rfc :)
 

> could Related as well?

Ideally yes, it should suggest other existing contacts for example

jancborchardt: Fixed this by renaming:
- "Relationship" → "Relationship to you"
- "Related" → "Related contacts"

We can add suggestions in a follow-up.

jancborchardt:
Ready to review now! :)

If someone wants to help with the "displaying placeholders for social profiles, see also #1445" you are welcome, otherwise I think we can also finish this in a follow-up.

skjnldsv:
Screenshots?

src/components/Properties/PropertyTitle.vue
```diff
@@ -23,11 +23,9 @@
 <template>
 	<h3 class="property__title property__row">
 		<div :class="icon" class="property__label property__title--icon" />
-		<span class="property__value property__title--right">
-			<div>{{ readableName }}</div>
-			<!-- display tooltip with hint if available -->
-			<div v-if="info" v-tooltip.auto="info" class="property__title--icon-details icon-details" />
```
skjnldsv, line 29: I still disagree with this.
We had to add this because lots of users are confused and vCards is complicated.
I don't want to remove it and I think it doesn't create any issues to leave it there

jancborchardt: Did you ever see something like this in another popular Contacts app?

skjnldsv: Let's ask the question differently, what would you say to a user that says they doesn't understand the difference between RELATED & RELATIONSHIP ? :)

jancborchardt: To that I would say that I fixed the wording via https://github.com/nextcloud/contacts/pull/1446/commits/715df614b33ada3b6ef8c358cec52ae2b9070e6b
→ to 'Relationship to you' and 'Related contacts' 

;)

skjnldsv: Is it more clear? Doesn't really seems to me :see_no_evil:

jancborchardt: Yes, it’s quite clear. One thing is to put in the "Relationship to you" andthe other thing is to put in "Related contacts". Not sure how much clearer it can be.

skjnldsv: Sure, your call!

jancborchardt:
Fixed your feedback @skjnldsv and rebased as well, ready to review @nextcloud/contacts. :)

jancborchardt:
@skjnldsv not sure about the waiting reports, they don’t seem to come back? Want to merge? :)

skjnldsv:
That should trigger it again :)

src/models/rfcProps.js
```diff
@@ -220,35 +219,27 @@ const properties = {
...
+			{ id: 'LINKEDIN', name: 'LinkedIn', placeholder: 'https://linkedin.com/…' },
+			{ id: 'PINTEREST', name: 'Pinterest', placeholder: 'https://pinterest.com/…' },
+			{ id: 'QZONE', name: 'QZone', placeholder: 'https://qzone.com/…' },
+			{ id: 'TUMBLR', name: 'Tumblr', placeholder: 'https://tumblr.com/…' },
+			{ id: 'TWITTER', name: 'Twitter', placeholder: 'https://twitter.com/…' },
+			{ id: 'WECHAT', name: 'WeChat', placeholder: 'https://wechat.com/…' },
+			{ id: 'YOUTUBE', name: 'YouTube', placeholder: 'https://youtube.com/…' },
+			{ id: 'MASTODON', name: 'Mastodon', placeholder: 'https://mastodon.social/…' },
+			{ id: 'DIASPORA', name: 'Diaspora', placeholder: 'https://joindiaspora.com/…' },
+			{ id: 'OTHER', name: 'Other social media', placeholder: 'https://example.com/…' },
```
call-me-matt, line 236: The placeholders did not show up when I tried it out?
placeholder

And because you asked: I find it easier to read when only the username is stored, not the URL. In my point of view that is redundant information to the type and just fills up the complete textfield.
image

call-me-matt:
What do you think about hiding the date picker in the same way we hide the trashcan?
image

PR #1447, opened 2020-01-28
Remove header color bar as it is not related to the contact

jancborchardt:
Just experimenting with this. Especially as it conflicts with the header bar more and more and it just looks a bit off these days? :)

We should think of another way to focus on the photo and name of the contact, showing them more presently.

### DIFF
--- a/css/ContactDetails.scss
+++ b/css/ContactDetails.scss
@@ -27,7 +27,7 @@
 	header {
 		display: flex;
 		align-items: center;
-		height: 100px;
+		padding: 50px 0 20px;
 		font-weight: bold;
 
 		// ORG-TITLE-NAME
@@ -50,14 +50,11 @@
 				padding: 4px 5px;
 				white-space: nowrap;
 				text-overflow: ellipsis;
-				color: #fff !important;
 				border: none;
 				background: transparent;
-				text-shadow: 0 0 2px var(--color-box-shadow);
 				font-size: inherit;
-				&::placeholder {
-					opacity: .8;
-					color: #fff !important;
+				&#contact-fullname {
+					font-weight: bold;
 				}
 			}
 			#contact-org:placeholder-shown {
@@ -71,10 +68,6 @@
 			display: flex;
 			.header-menu {
 				margin-right: 10px;
-				.action-item__menutoggle {
-					// force white over header colour
-					color: white;
-				}
 			}
 			.header-icon {
 				width: 44px;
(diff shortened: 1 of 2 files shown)


jancborchardt:
What do you think in general @nextcloud/contacts? :)

skjnldsv:
Ah right I forgot about this one :p
Maybe still enphasis the header? Primary color?

Also you can cleanup the `colorAvatar` computed and associated methods in the vue files since we don't need em anymore ;)

skjnldsv:
Also, please rebase ;)

jancborchardt:
> Maybe still enphasis the header? Primary color?

I gave it a bit more whitespace on the top and a tad more on the bottom. I think we need to fix the alignment together with the content below, but something for a separate issue. Since the header directly above is using primary color, I wouldn’t do anything else there.

> Also you can cleanup the `colorAvatar` computed and associated methods in the vue files since we don't need em anymore ;)

Done, thanks for the pointer :)


> Also, please rebase ;)

Done as well :)

jancborchardt:
@nextcloud/contacts please review :)

skjnldsv:
Screenshots? :grin:

call-me-matt:
> Screenshots? grin

grafik

jancborchardt:
> After removing the color bar, I found the contact bubbles quite colorful... A suggestion could be the design from Samsung (or a mix of both)

Let’s keep this separate, as the colored bubbles are used by a lot of different apps. They are in use by Google, Signal, Microsoft etc. and that is for a reason: They help to quickly separate people who didn’t set their avatars, making it much easier to tell people apart in a first-run scenario.

skjnldsv:
There is something that bothers me a little though.
Maybe it's just the change :)
I have a related question though, I remember a discussion about having the avatar as blured background if there was one.
Another question, now that I also think about this: after testing the upcoming grid view in Talk, it looks super similar. Are you sure you don't want to keep the colored background?

jancborchardt:
> Nonetheless, I have a related question: I remember a discussion about having the avatar as blured background if there was one. Is this still something you want?
> Another question, now that I also think about this: after testing the upcoming grid view in Talk, it looks super similar. Are you sure you don't want to keep the colored background?

It just doesn’t work really well with text over it, and takes waaaay too much attention. It works for Talk as that is a specific mode you are in and it is very visual, but is not good for when you want to get info or modify it.

So I’d say: Can we merge it?

Issue #522, opened 2018-03-19
[Feature Request] Groups sorting in browser (_top split from Issue #517_)


WNYmathGuy:
I do a lot with the **Groups** field in my address book entries in NextCloud. In the left panel of my browser, the groups seem to have a type of moving random sort to them. I don't know if the last edited group goes to the bottom, or what, but it's hard to be an alphabetically thinking person looking for a group that's NOT alphabetically sorted.

tdoerschel:
It seems that the groups are sorted by number of entries.
I would also prefer an alphabetical sorting.

tdoerschel:
I tried to find the groups-list in source code but could not find.
Can you please hint to where the sorting takes place?

I have a bit of coding experience but new to JS+vue.
What is a good place to start/learn, especially for nextcloud apps?

skjnldsv:
Hello  @tdoerschel !
You will find the group sorting here: https://github.com/nextcloud/contacts/blob/72e0b33e3bbd23635cf2efd23007fb5492d04e57/src/views/Contacts.vue#L215

Regarding development, you can find the vue example app here: https://github.com/skjnldsv/vueexample
You can have a look at our components repository and documentation too: https://github.com/nextcloud/nextcloud-vue

And finally, feel free to ask any questions by pinging me on the #nextcloud-dev channel on irc(freenode) :)

torstendoerschel:
Hello @skjnldsv,
thanks for the information. Sorry I am not good enough in programming and don't have the time at hand to start. We use your app a lot at work. I put a bounty from my private account now.
Maybe I can convince some colleguages to chip in as well.

tdoerschel:
Hi,
any plans when to the next minor release comes out with this issue fixed?
We have >1600 contacts in >120 groups.
Performance of nextcloud on a shared webspace is excellent, but searching with groups is painful, also due to issue #1145.

PR #1451, opened 2020-01-31
Sort groups alphabetically

ImUrX:
Refers to #522 
I did the solution just giving sort no arguments and overriding toString to each group's name.
Would like to test, but I dont have any idea on how to do so.


skjnldsv:
Hi, and thanks for your pr! :)
So, instead of changing the behaviour entirely, I would suggest adding a setting under the contacts sorting dropdown (in the bottom left settings area)
What do you think?

ImUrX:
Sure, but which one would be the default one?

ImUrX:
i accidentaly deleted package-lock.json :/

ImUrX:
Gonna remake the whole branch so it’s a little more clean

mburnicki:
I think the entries for "All contacts" and "Contacts without group" shhould be at the top of the list, just like it's now, and below should be a list of all groups, sorted by name.

I don't know what determines the sequence of groups if they are *not* sorted by name (as it's now, unfortunately). This is just confusing if you have many groups, and below them groups like "colleagues-local" "colleagues-external", colleagues-formerly", etc.

And, by the way, when editing a contact, also the selection box should where you can assign the groups should provide the list of group names sorted by name, IMO.

PR #1462, opened 2020-02-04
Add ability to clone contact

skjnldsv:
Capture d’écran_2020-02-04_16-26-58

If addressbook is readonly, we should be able to import into another available addressbook

- [x] UI
- [x] Clone methods
- [x] 1 avail. addressbook
- [x] Select if more than 1 avail.


src/store/addressbooks.js
```diff
@@ -512,6 +512,34 @@ const actions = {
...
+	 * @param {Object} data.addressbook the addressbook to move the contact to
+	 * @returns {Contact} the new contact object
+	 */
+	async copyContactToAddressbook(context, { contact, addressbook }) {
+		// init new contact & strip old uid
+		const vData = ICAL.stringify(contact.vCard.jCal).replace(/^UID.+/im, '')
+		const newContact = new Contact(vData, addressbook)
+
+		try {
+			const response = await contact.dav.copy(addressbook.dav)
```
skjnldsv: Maybe use contact.addressbook.dav.createVCard(vData) instead to force a different filename and avoid conflicts

rubo77:
How do I get this feature? I updated my Contacts to 3.3.0 but in the 3-dot -menu I only see download, QR and delete options.

I am on NC 18.0.3.

I have those open problems:
```
MySQL is used as database but does not support 4-byte characters. To be able to handle 4-byte characters (like emojis) without issues in filenames or comments for example it is recommended to enable the 4-byte support in MySQL. For further details read the documentation page about this.
The "X-Download-Options" HTTP header is not set to "noopen". This is a potential security or privacy risk, as it is recommended to adjust this setting accordingly.
```

but I guess, that's not the reason for the missing copy-option?

skjnldsv:
Hey!
The clone functionality is only hear when the contact is in read only :)
Do you have a specific use for it to be copied somewhere else?

We had a specific use case in mind, but I guess we could show the option all the time :thinking:

rubo77:
> The clone functionality is only hear when the contact is in read only

How do you set a contact to read-only?

Yes, the option should be there for all contacts

PR #1568, opened 2020-04-11
Allow to use any name for Group creation

### LINKED ISSUE #1563: Creating new Group with special characters in the name
I was trying to crate a group with name "Clients: Branch in a Company" and the group was simply not crated. I was first confused and thought, that this feature is not working. After short search for solution it came to me that maybe is issue with ":" character. And yes it is.

It would be nice to get an error message/notification when doing this. So that a user knows what is happening and it is not confused.

**Describe alternatives you've considered**
Simply state "Character "x" can not be used in a group name.".

Thank You for Your great work!

tintou:
The Group name is already escaped by the underlying vCard library, there is no need to verify things here.
Fixes #1563

### DIFF
--- a/src/components/Properties/PropertyGroups.vue
+++ b/src/components/Properties/PropertyGroups.vue
@@ -169,14 +169,9 @@ export default {
 		 * @returns {boolean}
 		 */
 		validateGroup(groupName) {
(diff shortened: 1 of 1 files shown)


skjnldsv:
Thanks!

PR #1415, opened 2020-01-17
Show transparent contact photo on a lighter background

hanzi:
fixes #1414 
fixes #1066 

This improves support for transparent contact photos by

1. Removing the still visible first letter in the contact list
2. Placing photos on a much lighter background colour so it doesn't clash with the transparent image

SharedScreenshot

PR #1596, opened 2020-04-22
Show transparent contact photos on a lighter background

jancborchardt:
Just picking up @hanzi’s pull request from https://github.com/nextcloud/contacts/pull/1415 and doing a minor adjustment to the values.

SharedScreenshot

### DIFF
--- a/css/ContactDetailsAvatar.scss
+++ b/css/ContactDetailsAvatar.scss
@@ -55,6 +55,8 @@
 		background-repeat: no-repeat;
 		background-position: center;
 		background-size: cover;
+		// White background for avatars with transparency, also in dark theme
+		background-color: #fff;
 	}
 	&__options {
 		position: absolute;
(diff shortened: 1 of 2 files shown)


src/components/ContactDetails/ContactDetailsAvatar.vue
```diff
@@ -26,7 +26,7 @@
 		<div class="contact-header-avatar__wrapper">
 			<div class="contact-header-avatar__background" @click="toggleModal" />
 			<div v-if="contact.photo"
-				:style="{ 'backgroundImage': `url(${contact.photoUrl})` }"
+				:style="{ 'backgroundImage': `url(${contact.photoUrl})`, 'backgroundColor': '#fff' }"
```
skjnldsv, line 29: Better in the direct scss file rather than inline if it's not computed

jancborchardt: Good point – fixed

jancborchardt:
Adjusted, new review please @nextcloud/contacts :)

PR #1644, opened 2020-05-25
Fix Content-Wrapper CSS

### LINKED ISSUE #1632: UI appears broken
### Describe the bug

The addressbook entry, also the very first on upon creation,  is shown below the fold. It is disconnected form the column layout and the some elements (e.g main name, company, phone) are not editable.

**To Reproduce**
1. Get contacts from git, run make, enable it
2. move to contacts app
3. click on "New contact"

**Expected behavior**

Addressbook entries shall be shown as before

**Actual behavior**

Screenshot_20200516_005720

and after pressing space or page down:

Screenshot_20200516_005833

### Server configuration

**Operating system**: Linux (Arch btw)

**Web server:** Apache

**Database:** PostgreSQL... 

**PHP version:** 7.4

**Nextcloud version:** git - stable19 - 9d7198f31e75498f2df35c156cbcfdfc2f346ad2

**Contacts version:** git - master - 4bec35b0ce9b5a114f34b435335271d2db0fe84a

**Updated from an older Nextcloud or fresh install:** fresh

### Client configuration
**Browser:** FF 76

**Operating system:** Linux

#### Browser log

no error

jotoeri:
For the `#app-content-wrapper` i'm not sure, if this is really the right place for the css. But feels like beeing misplaced everywhere else.^^

Fixes #1632

### DIFF
--- a/css/ContactDetails.scss
+++ b/css/ContactDetails.scss
@@ -23,6 +23,7 @@
 #contact-details {
 	$grid-column-gap: 20px;
 	$grid-column-width: 350px;
+	flex: 1 1 524px;
 	// header
 	header {
 		display: flex;
--- a/css/contacts.scss
+++ b/css/contacts.scss
@@ -37,6 +37,10 @@ $grid-input-height-with-margin: $grid-height-unit - $grid-input-margin * 2;
 	}
 }
 
+#app-content-wrapper {
+	display: flex;
+}
+
 // components
 @import 'SettingsSection';
 @import 'Settings/SettingsAddressbook';


css/ContactDetails.scss
```diff
@@ -23,6 +23,7 @@
 #contact-details {
 	$grid-column-gap: 20px;
 	$grid-column-width: 350px;
+	flex: 1 1 524px;
```
skjnldsv, line 26: Where is this number coming from?

jotoeri: That was the one, that came from server, so what applied before changing to vue-classes. No more specific idea behind, just change it if you have better number... Was more like a quick-fix to get UI back to the same usable state as before.

Issue #1513, opened 2020-03-11
Contact picture not changeable / removable


Bodenseematze:
### Describe the bug

If a contact does not have a picture, the contacts app allows to set one (e.g. by uploading a picture from the PC).
But if a contact already has a picture, there is no way to change or remove it.

......
......

**To Reproduce**
1. Open contacts app
2. Click on a contact without picture
3. Click on the empty place where the picture normally is - a dialog opened with "upload a new picture" and "coose from files"
4. Add a picture to that contact - you'll see the picture of the contact instead the empty place
5. If you now click on that "round circle" with the picture, the picture opens "big" - but there's no dialog popping up - neither at the small "round circle picture" nor at the "big picture".

**Expected behavior**

The dialog to choose another picture or to remove the existing picture should be available.

**Actual behavior**

No dialog is available after a picture has been added.

### Server configuration

**Operating system**: Linux

**Web server:** Apache, Nginx...

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud version:** 17.0.3

**Contacts version:** 3.2.0

**Updated from an older Nextcloud or fresh install:** updated

Tested with Firefox 73.0.1

skjnldsv:
Hey! Have you check the top right corner of the image?

Bodenseematze:
Youd mean the "big picture"?

I rechecked - there's nothing happening - neither at the "big picture" nor on the "circle picture"...

skjnldsv:
![Capture d’écran_2020-03-11_15-23-15](https://user-images.githubusercontent.com/14975046/76427146-3e5c9100-63ac-11ea-857f-658f8fa44fdf.png)

Bodenseematze:
Contact without pic:
![Contacts_NewContact01](https://user-images.githubusercontent.com/13251886/76503520-f9813a80-643d-11ea-8fc1-1ce668195064.jpg)

![Contacts_NewContact02](https://user-images.githubusercontent.com/13251886/76503559-02720c00-643e-11ea-83f0-6142d62016c0.jpg)

Contact with pic:
![Contacts_ContactWithPic01](https://user-images.githubusercontent.com/13251886/76503613-0c940a80-643e-11ea-8bfe-86fbf31277b9.jpg)

![Contacts_ContactWithPic02](https://user-images.githubusercontent.com/13251886/76503625-11f15500-643e-11ea-91a4-f87bf3795b55.jpg)

skjnldsv:
[...]

PR #1645, opened 2020-05-26
Additional link to picture menu

call-me-matt:
It took me a while to find the menu where I can delete contact pictures, so here is a proposal to add an overlay and make it more visible to users. Also: moved and merged the menu from the modal, which currently has different menu entries than the button.

hidden-icon

Related issues: #1513

Implementation is very rudimentary, just to give you an idea. Next steps would be to optimize the style (edit or photo icon instead of text, chose color/transparency) and then implement it well in the scss.

### DIFF
--- a/css/ContactDetailsAvatar.scss
+++ b/css/ContactDetailsAvatar.scss
@@ -57,6 +57,9 @@
 		background-size: cover;
 		// White background for avatars with transparency, also in dark theme
 		background-color: #fff;
+		&:hover + .contact-avatar-options__popovermenubtn {
+			opacity: .25;
+		}
 	}
 	&__options {
 		position: absolute;
@@ -68,7 +71,6 @@
 		display: block;
 		width: 100%;
 		height: 100%;
-		opacity: .5;
 		background-color: rgba(0, 0, 0, .2);
 		&:hover,
 		&:active,
@@ -78,6 +80,24 @@
 		&__popovermenu {
 			// center
 			margin: calc((100% - 44px) / 2);
+		}
+		&__popovermenubtn {
+			height: 45px;
+			width: 45px;
+			border-radius: 50%;
+			z-index: 11;
+			background: rgba(0, 0, 0, .2);
+			opacity: 0;
+			&:hover,
+			&:active,
+			&:focus {
+				opacity: 1;
+			}
+			// bottom right
+			margin: calc((100% + 5px) / 2);
+		}
+		&__popovermenu,
+		&__popovermenubtn {
 			& .action-item__menutoggle {
 				// hide three dot menu, in favour of icon-picture-force-white
 				z-index: -1;
--- a/src/components/ContactDetails/ContactDetailsAvatar.vue
+++ b/src/components/ContactDetails/ContactDetailsAvatar.vue
@@ -24,15 +24,24 @@
 
 <template>
 	<div class="contact-header-avatar">
-		<div class="contact-header-avatar__wrapper">
+		<div v-click-outside="closeMenu" class="contact-header-avatar__wrapper">
 			<div class="contact-header-avatar__background" @click="toggleModal" />
 
 			<div v-if="contact.photo"
 				:style="{ 'backgroundImage': `url(${contact.photoUrl})` }"
-				class="contact-header-avatar__photo"
-				@click="toggleModal" />
+				class="avatar contact-header-avatar__options contact-avatar-options contact-header-avatar__photo"
+				@click="toggleModal">
+				<input id="contact-avatar-upload"
+					ref="uploadInput"
+					type="file"
+					class="hidden"
+					accept="image/*"
+					@change="processFile">
+			</div>
 
-			<div v-click-outside="closeMenu" class="contact-header-avatar__options">
+			<div v-if="!contact.photo"
+				v-click-outside="closeMenu"
+				class="contact-header-avatar__options">
 				<a v-tooltip.bottom="t('contacts', 'Add a new picture')"
 					href="#"
 					class="contact-avatar-options"
@@ -53,17 +62,41 @@
 				size="large"
 				:title="contact.displayName"
 				@close="toggleModal">
+				<!-- attention, this menu exists twice in this file -->
 				<template #actions>
-					<ActionButton v-if="!isReadOnly" icon="icon-upload" @click="selectFileInput">
-						{{ t('contacts', 'Upload new picture') }}
+					<ActionButton
+						v-if="!isReadOnly"
+						icon="icon-upload"
+						@click="selectFileInput">
+						{{ t('contacts', 'Upload a new picture') }}
 					</ActionButton>
-					<ActionButton v-if="!isReadOnly" icon="icon-folder" @click="selectFilePicker">
+					<ActionButton
+						v-if="!isReadOnly"
+						icon="icon-folder"
+						@click="selectFilePicker">
 						{{ t('contacts', 'Choose from files') }}
 					</ActionButton>
-					<ActionLink :href="`${contact.url}?photo`" icon="icon-download" target="_blank">
+					<template v-if="!isReadOnly">
+						<ActionButton
+							v-for="network in supportedSocial"
+							:key="network"
+							:icon="'icon-' + network.toLowerCase()"
+							@click="getSocialAvatar(network)">
+							{{ t('contacts', 'Get from ' + network) }}
+						</ActionButton>
+					</template>
+					<!-- FIXME: the link seems to have a bigger font size than the button caption -->
+					<ActionLink
+						v-if="contact.photo"
+						:href="`${contact.url}?photo`"
+						icon="icon-download"
+						target="_blank">
 						{{ t('contacts', 'Download picture') }}
 					</ActionLink>
-					<ActionButton v-if="!isReadOnly" icon="icon-delete" @click="removePhoto">
+					<ActionButton
+						v-if="!isReadOnly && contact.photo"
+						icon="icon-delete"
+						@click="removePhoto">
 						{{ t('contacts', 'Delete picture') }}
 					</ActionButton>
 				</template>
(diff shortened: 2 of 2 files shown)


jancborchardt:
Actually the button being invisible below there is not by design. :) What should happen is that the action menu shows in the top right of the viewer, next to the x button. Then clicking the image anywhere just opens the image, and from there it can be edited.

call-me-matt:
Why don't we put it where everyone else places this function?
|github|facebook|google|
|---|---|---|
|profile-github|profile-facebook|profile-google|

jancborchardt:
Ok, sure. :) Then let's do it like Facebook or the left example of Google do it: Overlaying the menu button on the bottom right, and having an image filetype icon.

call-me-matt:
I liked the other implementation from google, because it is more decent (only shown on mouse over). It's at the spot where you are looking for it and at the same time not disturbing in regular operation. Are you sure you prefer the bubble?

jancborchardt:
Sorry – you are right, I meant to say it’s good to only show the bubble on hover/focus of the picture. :) (Make sure focus also works for keyboard accessibility.)

call-me-matt:
Like this?
I think I don't need to treat the accessibility, as it is just an additional button, see:

overlay

jancborchardt:
Yes, almost perfect – only that the action menu should open from the overlaid image button, not from this invisible action button below the image. :)

call-me-matt:
Should we merge the buttons? Maybe a css pro can help?

call-me-matt:
I have tested it with Firefox and Chromium. Looking forward to your comments.

skjnldsv:
Nice!!

skjnldsv:
Is thsi ready for reviews or just a draft?

call-me-matt:
> Is thsi ready for reviews or just a draft?

It needs a code review from someone who knows scss, I'd say

src/components/ContactDetails/ContactDetailsAvatar.vue
```diff
@@ -25,12 +25,29 @@
...
+				<input id="contact-avatar-upload"
+					ref="uploadInput"
+					type="file"
+					class="hidden"
+					accept="image/*"
+					@change="processFile">
+			</div>
+
+			<div v-if="contact.photo"
+				class="overlay"
```
skjnldsv, line 43: Please use the BEM css class naming methods :)
```suggestion
				class="contact-header-avatar__overlay"
```

src/components/ContactDetails/ContactDetailsAvatar.vue
```diff
@@ -285,7 +310,6 @@ export default {
 		 */
 		removePhoto() {
 			this.contact.vCard.removeProperty('photo')
-			this.maximizeAvatar = !this.maximizeAvatar
```
skjnldsv, line 288: This allowed closing the modal on remove as you can also remove the avatar when displaying it full view

css/ContactDetailsAvatar.scss
```diff
@@ -57,18 +57,42 @@
...
+	.overlay {
+		position: absolute;
+		width: 35px;
+		height: 35px;
+		z-index: 11;
+		cursor: pointer;
+		opacity: 0;
+		border-radius: 50%;
+		background: rgba(0, 0, 0, .5);
+		margin: calc((100% + 15px) / 2);
```
skjnldsv, line 79: Usually not fond of half percentage as this can lead to blurry popups :)

src/components/ContactDetails/ContactDetailsAvatar.vue
call-me-matt: I merged the two menus (that means no modal menu any more)

css/ContactDetailsAvatar.scss
call-me-matt: I took the code as used in the popovermenu (see line 104). Feel free to adapt.

src/components/ContactDetails/ContactDetailsAvatar.vue
call-me-matt: OK, I can do that.

src/components/ContactDetails/ContactDetailsAvatar.vue
skjnldsv: Well, I would keep the modal menu anyway, no?

call-me-matt: I'd say we don't need the menu twice.

skjnldsv: But you cannot access it when the image is opened? SO it is extra clicks, as you need to close and then open the menu again?

skjnldsv: But I'll let  @jancborchardt decide :)

jancborchardt: It’s not bad to have the menu in both places. Sometimes you need to look at it in detail to decide. :)

skjnldsv:
Please rebase @call-me-matt I'll do another review on this :)

src/components/ContactDetails/ContactDetailsAvatar.vue
call-me-matt: In order to prepare that, I moved the menu into a component. @skjnldsv maybe an intermediate review is possible before I start to include the menu also in the modal?

src/components/ContactDetails/ContactDetailsAvatarMenu.vue
```diff
@@ -0,0 +1,91 @@
...
+		<ActionButton
+			v-if="!isReadOnly"
+			icon="icon-picture"
+			@click="$emit('select-file-picker')">
+			{{ t('contacts', 'Choose from files') }}
+		</ActionButton>
+		<!-- FIXME: the link seems to have a bigger font size than the button caption -->
+		<ActionLink
+			v-if="hasPhoto"
+			:href="`${contact-photo}`"
```
call-me-matt, line 43: needs to be contactPhoto

src/components/ContactDetails/ContactDetailsAvatar.vue
skjnldsv: Hey! You can't do that, the Actions component is handled by the Modal one.
That's why I said you'd have to duplicate code for now and put this into `<template #actions>`

src/components/ContactDetails/ContactDetailsAvatar.vue
```diff
@@ -25,12 +25,29 @@
...
+					class="hidden"
+					accept="image/*"
+					@change="processFile">
+			</div>
+
+			<div v-if="contact.photo"
+				class="contact-header-avatar__overlay"
+				@click.stop.prevent="toggleMenu">
+				<div class="overlay-icon icon-picture-force-white" />
+			</div>
```
skjnldsv, line 46: Why not use the menu directly here?
I don't understand the need to have a button that opens the menu?

You can customize the default menu icon if that was the issue you were facing :)
`default-icon="icon-picture-force-white"`

src/components/ContactDetails/ContactDetailsAvatar.vue
call-me-matt: I thought with programming everything is possible... but ok, if I really have to, I will then just duplicate the code.

src/components/ContactDetails/ContactDetailsAvatar.vue
call-me-matt: It's more that I wanted the button to be visible as soon as you hover over the avatar and to change it's opacity as soon as you hover over to the button itself. I could not figure out how to do that with an action button. Is that possible?

|              | action button | with additional button |
|--------------|---------------|------------------------|
| page view | image | image |
| hover avatar | image | image |
| hover button | image | image |

*(icons by [Vectto](https://www.iconfinder.com/icons/2624896/avengers_captain_captain_america_super_hero_icon), CC BY 3.0)*

skjnldsv: Sure, you should still be able to manage it like that :)
Currently this si weird as there is this two element superposition :thinking:

skjnldsv:
Also, don't forget to squash and use fixups. We're already at 12 commits ;)

src/components/ContactDetails/ContactDetailsAvatar.vue
call-me-matt: I have pushed my attempts, but need some help with the detection whether the menu is opened. Is there a way to trigger that toggleMenu when the action button is clicked (without putting a div around it)?

src/components/ContactDetails/ContactDetailsAvatar.vue
```diff
@@ -73,24 +98,52 @@
 			</Modal>
 
 			<!-- out of the avatar__options because of the overflow hidden -->
-			<Actions :open="opened" class="contact-avatar-options__popovermenu">
-				<ActionButton v-if="!isReadOnly" icon="icon-upload" @click="selectFileInput">
+			<!-- attention, this menu exists twice in this file -->
+			<Actions
+				default-icon="icon-picture-force-white"
+				:open="opened"
```
skjnldsv, line 104:
```suggestion
				:open.sync="opened"
```

call-me-matt: Awesome - that was exactly what was missing. :+1: 
Thank you!

src/components/ContactDetails/ContactDetailsAvatar.vue
call-me-matt: Thank you for the hint, the .sync solved the issue!

call-me-matt:
Design question (@jancborchardt)
Is it intentional to have the folder-icon themed?
grafik

skjnldsv:
> Is it intentional to have the folder-icon themed?

Yes :p
But maybe not for the menus :thinking:

call-me-matt:
I realized the "chose from files" icon is an image in the avatar-menu and a folder on the modal-menu. Which one to take?

skjnldsv:
> I realized the "chose from files" icon is an image in the avatar-menu and a folder on the modal-menu. Which one to take?

Let's leave it like that for now? @jancborchardt

PR #1687, opened 2020-07-04
Enh/add virtual group

skjnldsv:
Peek 04-07-2020 10-07
outdated gif
____
- [x] Virtual group & `Add group` button
- [x] Base UI
- [x] Dav plugin
- [x] Picker
  - [x] Cleanup & fix UserBubble component
  - [x] Standardisation
  - [ ] QA
- [x] Progress
  - [x] Modal
  - [x] Animation
  - [x] Finish state
  - [ ] QA
- [x] Update local contacts
- [ ] QA

### Bugs
- [ ] Do NOT provide read-only contacts
- [ ] Do NOT provide contacts that already belong to the group


jancborchardt:
Looks nice from first view, small points:
- The modal on top of modal looks strange – can’t the second modal replace the first? (Or switch content by e.g. sliding in from the right, keeping the same shape – that would look great.)
- When done, the text is still "Adding 6 contacts to Group 1". Instead the feedback should be "Added 6 contacts to Group 1" and the icon can be switched to icon-checkmark.
- Also, if there are no errors on adding, that modal can automatically be closed after ~3 seconds of showing the "Added …" feedback view.

jancborchardt:
(Btw I’m also getting this error of "This contact was broken and received a fix …" on a completely new installations with a contact I just created with just a name. Anything I can do to fix that, or is there a pull request?)

skjnldsv:
> (Btw I’m also getting this error of "This contact was broken and received a fix …" on a completely new installations with a contact I just created with just a name. Anything I can do to fix that, or is there a pull request?)

Not related to this. No pr yet

skjnldsv:
@jancborchardt let me merge and fix in followup so it's easier for me

PR #1688, opened 2020-07-07
Import vcf from files

skjnldsv:


jancborchardt:
Shouldn’t the default action on .vcf files be to just import them? To the majority of people it makes little sense to edit them in a text editor. :) Then people don’t have to know that there’s an import action in the menu and find it – they can just click the file.

PR #1691, opened 2020-07-09
Move routes to info.xml

MorrisJobke:
Initially I wanted to move this to lazy loading code and then realized that putting it into info.xml does already this.


MorrisJobke:
Beside removing code it also reduces the amount of PHP files from 6 to 4 in contacts :P

MorrisJobke:
> Is this really the recommended way?

No, but it's an easy way to lazy load it. If we make it better, then we can go this way.

skjnldsv:
Let get this in then! :rocket: 
Thanks for the fix!

PR #1580, opened 2020-04-14
get contact pictures from social networks

call-me-matt:
Fix #1546
Implements partly https://github.com/nextcloud/contacts/issues/20

PR #1692, opened 2020-07-10
Enh/social sync

skjnldsv:
closes  https://github.com/nextcloud/contacts/pull/1580


lib/Service/Social/InstagramProvider.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+	
+	/**
+	 * Returns the profile-id
+	 *
+	 * @param {string} the value from the contact's x-socialprofile
+	 *
+	 * @return string
+	 */
+	public function cleanupId(string $candidate):?string {
```
skjnldsv, line 44: @call-me-matt on those functions, please directly use the strict type hint.
`basename` always returns a string :)

```suggestion
	public function cleanupId(string $candidate): string {
```

skjnldsv: Same goes for all the other social providers files :)

call-me-matt: For Instagram that is true, but Mastodon or Twitter providers, for example, could return NULL. Thought it would be easier to handle upcoming errors this way. Do you think it's not a good idea?

skjnldsv: The goal to have strict typing s to make sure we know what to expect :)
If there is indeed a potential null, let's leave it like that for the Mastodon or Twitter provider :rocket:

call-me-matt:
I enhanced the facebook provider with a function to determine the numeric id if only the user name is given. It's not very reliable, but better than nothing, I'd say.

skjnldsv:
@call-me-matt please run `composer run cs:fix` and let get this merged :)

skjnldsv:
Congrats! :rocket: :tada:

appinfo/app.php
```diff
@@ -0,0 +1,26 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+use OCA\Contacts\AppInfo\Application;
+
+$app = \OC::$server->query(Application::class);
```
MorrisJobke, line 26: `app.php` is not needed. If there is an `Application` then it will be instantiated automatically.

MorrisJobke: Ah - just on 20+. And as this is for 17-20 it is still needed 🙈

skjnldsv: Yep :+1:

Issue #1594, opened 2020-04-21
Design improvements on "Update avatar from social media" feature


jancborchardt:
(Referring to **get contact pictures from social networks #1580** which is still work in progress.)

Some feedback about the design and user experience:
Currently it seems that every contact has an entry "Update from social media" action in the avatar menu. This is very tedious to go through every person individually and check for updates.

What would be expected:
- Either a button "Update avatars from social media" in the bottom left settings which updates everyone’s avatars. Modal with progress feedback can be shown, just like on import.
- Or ideally it would be entirely automatic. Then we would have an opt-in checkbox in the bottom left settings saying "Automatically update avatars from social media". Works the same as the button, except that it regularly checks in the background and you don’t need to click anymore.

jancborchardt:
cc @call-me-matt if you like to look into that later as well. :)

skjnldsv:
> * Or ideally it would be entirely automatic. Then we would have an opt-in checkbox in the bottom left settings saying "Automatically update avatars from social media". Works the same as the button, except that it regularly checks in the background and you don’t need to click anymore.

With background job :+1: 
We would have to make the user chose which social he prefers though. VCard supports setting a pref option (see #202), but we still don't support it :)

So how do you feel about NOT having automated sync until we can ser pref properties? Or would you see another way like take the first one if there is more than two? or something else?

call-me-matt:
I am very happy you like the idea and I am looking forward to have this feature in the contacts app!

The ultimate target (for me), is to have regular background jobs to keep the contact pictures always up to date. The menu per contact is just a shortcut for manual updates (for example to test the profile id) or to have an instant update after adding a new contact. The menu item only shows up if the contact has entries of supported social networks, by the way.

Originally, I had an overview page per address book in mind, where you could see the status - and update all contacts at once, too. But after reflection with @skjnldsv I aggree that this was a little too much and a background job is perfect. If possible, though, It would be nice to have a notification (activatable in the settings?) when new avatars are synced.

Currently the search order for profile pictures in different social networks is static, the order could be configurable in the settings. The first found picture is then taken. The pref option is very interesting, but I personally would not wait for that.

Another question: should we allow to exclude certain contacts (with social profile) from background updates? (In my opinion, an "all or none" option is sufficient.)

call-me-matt:
Could it be an admin setting?
[...]

PR #1722, opened 2020-07-31
allow for regular background updates of avatars from social networks

call-me-matt:
implements #1594


README.md
```diff
@@ -1,5 +1,5 @@
 # Nextcloud Contacts
-![Downloads](https://img.shields.io/github/downloads/nextcloud/contacts/total.svg?style=flat-square)
+![Downloads](https://img.shields.io/github/downloads/nextcloud/contacts/total?style=flat-square)
```
skjnldsv, line 2: why? :thinking: :grin:

lib/Controller/PageController.php
```diff
@@ -69,13 +71,19 @@ public function __construct(string $appName,
 	 * Default routing
 	 */
 	public function index(): TemplateResponse {
+		$userId = \OC_User::getUser();
```
skjnldsv, line 74: Please inject it, do not use this :)

https://github.com/nextcloud/server/blob/5bf3d1bb384da56adbf205752be8f840aac3b0c5/lib/public/IUserSession.php#L79

lib/Cron/SocialUpdateRegistration.php
```diff
@@ -0,0 +1,91 @@
...
+		parent::setInterval(7 * 24 * 60 * 60);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function run($arguments) {
+
+		// check if admin allows for social updates:
+		$isAdminEnabled = $this->config->getAppValue($this->appName, 'allowSocialSync', 'yes');
```
skjnldsv, line 75: I would go for a different variable name.
Something like `$isSyncEnabledByAdmin`

lib/Cron/SocialUpdateRegistration.php
```diff
@@ -0,0 +1,91 @@
...
+		if (!($isAdminEnabled === 'yes')) {
+			return;
+		}
+
+		$this->userManager->callForSeenUsers(function (IUser $user) {
+
+			// check that user opted-in:
+			$isUserEnabled = $this->config->getUserValue($user->getUID(), $this->appName, 'enableSocialSync', 'no');
```
skjnldsv, line 83: same :)

src/components/AdminSettings.vue
```diff
@@ -1,3 +1,25 @@
+<!--
+  - @copyright Copyright (c) Matthias Heinisch <nextcloud@matthiasheinisch.de>
+  -
+  - @author Matthias Heinisch <nextcloud@matthiasheinisch.de>
```
skjnldsv, line 4: :+1:

lib/Service/SocialApiService.php
```diff
@@ -182,4 +215,82 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+	 *
+	 * Updates social profile data for all contacts of an addressbook
+	 * // TODO: how to exclude certain contacts?
+	 *
+	 * @param {String} network the social network to use (fallback: take first match)
+	 * @param {String} user the address book owner
+	 *
+	 * @returns {JSONResponse} JSONResponse with the list of changed and failed contacts
+	 */
+	public function updateAddressbooks(string $network, string $user) : JSONResponse {
```
skjnldsv, line 231: We usually use `userId` when we use the id, $user is usually refering to the IUser class :)

README.md
call-me-matt: No clue where that came from. I think my git has a personality.

src/components/AdminSettings.vue
call-me-matt: so you know who to blame :man_juggling: :policeman:

lib/Service/SocialApiService.php
```diff
@@ -124,6 +141,22 @@ protected function getAddressBook(string $addressbookId) : ?IAddressBook {
...
+	 * @NoAdminRequired
+	 *
+	 * Retrieves and initiates all addressbooks from a user
+	 *
+	 * @param {string} user the user to query
+	 * @param {IManager} the contact manager to load
+	 */
+	protected function registerAddressbooks($user, IManager $manager) {
+		$cm = new ContactsManager($this->davBackend, $this->l10n);
+		$cm->setupContactsProvider($manager, $user, $this->urlGen);
```
call-me-matt, line 154: Is there a way to limit this to address books owned by a user (no system address book, no shared address books from other users)? Or, alternatively, is there a way to find out about an owner of an IAddressBook?

lib/Service/SocialApiService.php
```diff
@@ -182,4 +215,82 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+		// TODO: keep only owned address books (not shared ones)
+
+		foreach ($addressBooks as $addressBook) {
+			if (is_null($addressBook)) {
+				break;
+			}
+
+			// get contacts in that addressbook
+			$contacts = $addressBook->search('', ['UID'], ['types' => true]);
+			// TODO: filter for contacts with social profiles
```
call-me-matt, line 265: Could be optimized with `$addressBook->search('', ['X-SOCIALPROFILE'], ['types' => true]);`, if `X-SOCIALPROFILE` was [indexed](https://github.com/nextcloud/server/blob/master/apps/dav/lib/CardDAV/CardDavBackend.php#L78-L80)

lib/Service/SocialApiService.php
```diff
@@ -182,4 +215,82 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+		$this->registerAddressbooks($userId, $this->manager);
+
+		$addressBooks = $this->manager->getUserAddressBooks();
+		// TODO: filter out system addr books
+		// TODO: filter out deactivated books
+		// TODO: keep only owned address books (not shared ones)
+
+		foreach ($addressBooks as $addressBook) {
+			if (is_null($addressBook)) {
+				break;
```
call-me-matt, line 257: should be `continue`

lib/Service/SocialApiService.php
```diff
@@ -182,4 +215,82 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+			}
+
+			// get contacts in that addressbook
+			$contacts = $addressBook->search('', ['UID'], ['types' => true]);
+			// TODO: filter for contacts with social profiles
+			if (is_null($contacts)) {
+				break;
```
call-me-matt, line 264: should be `continue`

lib/Cron/SocialUpdateRegistration.php
call-me-matt: Now Codacy complains:

> Avoid excessively long variable names like $isSocialEnabledByAdmin. Keep variable name length under 20.

Shall I go back to the previous variable names?

lib/Service/SocialApiService.php
skjnldsv: cc @georgehrke what do you think? Does it make sense to add filters to the IAddressBook interface?

lib/Cron/SocialUpdateRegistration.php
skjnldsv: xD, Ignore codacy, I don't even know why I enabled it in the first place :p

lib/Controller/SocialApiController.php
```diff
@@ -66,6 +72,38 @@ public function setAppConfig($key, $allow) {
...
+	 *
+	 * update appconfig (user setting)
+	 *
+	 * @param {String} key the identifier to change
+	 * @param {String} allow the value to set
+	 *
+	 * @returns {JSONResponse} an empty JSONResponse with respective http status code
+	 */
+	public function setUserConfig($key, $allow) {
+		$userId = $this->userSession->getUser()->getUid();
```
skjnldsv, line 86: Might be better to get the user and check against it being null before getting the UID :)

lib/Cron/SocialUpdateRegistration.php
call-me-matt: I think I found a compromise :wink:

lib/Service/SocialApiService.php
skjnldsv: > Could be optimized with `$addressBook->search('', ['X-SOCIALPROFILE'], ['types' => true]);`, if `X-SOCIALPROFILE` was [indexed](https://github.com/nextcloud/server/blob/master/apps/dav/lib/CardDAV/CardDavBackend.php?rgh-link-date=2020-07-31T19%3A06%3A55Z#L78-L80)

Mind to add this? :)

skjnldsv: if it's not indexed (for older version) does it still works? Is the searched value just ignored?
If so it's ok I guess :)

call-me-matt: Right, I have included this and added a fallback for older server versions. But concerning the indexer, I could not find out how to add the `X-SOCIALPROFILE`. Simply adding it to the linked array did not have any effect. Do I need to trigger the indexer? Maybe someone from the server-team can help?

call-me-matt: Pull request created: https://github.com/nextcloud/server/pull/22085

call-me-matt: Turns out it was not cron, but me refreshing the profiles and triggering a re-index by that. We'll need a background job to re-index when a user activates this function to make sure all contacts are found. @skjnldsv or is there a better way / place to do that?

MorrisJobke:
Bildschirmfoto 2020-09-15 um 12 36 26

😢

call-me-matt:
Ups, I'm sorry. That was an accident.

PR #1744, opened 2020-08-11
trigger re-index and optimize query by social field

call-me-matt:
trigger re-index of contacts to optimize query in background updates

### DIFF
--- a/lib/Migration/BuildSocialSearchIndex.php
+++ b/lib/Migration/BuildSocialSearchIndex.php
@@ -1,0 +1,91 @@
+<?php
+/**
+ * @copyright 2020 Matthias Heinisch <nextcloud@matthiasheinisch.de>
+ *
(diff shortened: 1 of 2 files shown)


call-me-matt:
This is a follow-up of https://github.com/nextcloud/server/pull/22085

skjnldsv:
Im' wondering if this should really be a cron job. Shouldn't it be a migration instead? @rullzer @nickvergessen @georgehrke  ?

nickvergessen:
It's a repair step, exactly what it should be, but it's not registered in info.xml

skjnldsv:
> Im' wondering if this should really be a cron job. Shouldn't it be a migration instead? @rullzer @nickvergessen @georgehrke ?

AH, I misread the code!
My bad :)

call-me-matt:
Thank you for your feedback, I will move the code to the server then (and pack it into background-chunks)

PR #1745, opened 2020-08-14
Update social avatars in chunks

call-me-matt:
When there are a lot of contacts, the background update of avatars from social networks shall be split into chunks

### DIFF
--- a/lib/Cron/SocialUpdate.php
+++ b/lib/Cron/SocialUpdate.php
@@ -25,18 +25,39 @@ namespace OCA\Contacts\Cron;
 
 use OCA\Contacts\Service\SocialApiService;
 
+use OCP\AppFramework\Http;
+use OCP\BackgroundJob\IJobList;
+
 class SocialUpdate extends \OC\BackgroundJob\QueuedJob {
 	/** @var SocialUpdateService */
 	private $social;
+	/** @var IJobList */
+	private $jobList;
 
-	public function __construct(SocialApiService $social) {
+	public function __construct(SocialApiService $social,
+								IJobList $jobList) {
 		$this->social = $social;
+		$this->jobList = $jobList;
 	}
 
 	protected function run($arguments) {
 		$userId = $arguments['userId'];
+		$offsetBook = $arguments['offsetBook'];
+		$offsetContact = $arguments['offsetContact'];
 
 		// update contacts with first available social media profile
-		$this->social->updateAddressbooks('any', $userId);
+		$result = $this->social->updateAddressbooks('any', $userId, $offsetBook, $offsetContact);
+
+		if ($result->getStatus() == Http::STATUS_PARTIAL_CONTENT) {
+			// not finished; schedule a follow-up
+			$report = $result->getData();
+			$stoppedAtBook = $report[0]['stoppedAt']['addressBook'];
+			$stoppedAtContact = $report[0]['stoppedAt']['contact'];
+			$this->jobList->add(self::class, [
+				'userId' => $userId,
+				'offsetBook' => $stoppedAtBook,
+				'offsetContact' => $stoppedAtContact
+			]);
+		}
 	}
 }
--- a/lib/Cron/SocialUpdate.php
+++ b/lib/Cron/SocialUpdate.php
@@ -42,8 +42,8 @@ class SocialUpdate extends \OC\BackgroundJob\QueuedJob {
 
 	protected function run($arguments) {
 		$userId = $arguments['userId'];
-		$offsetBook = $arguments['offsetBook'];
-		$offsetContact = $arguments['offsetContact'];
+		$offsetBook = $arguments['offsetBook'] ?? null;
+		$offsetContact = $arguments['offsetContact'] ?? null;
 
 		// update contacts with first available social media profile
 		$result = $this->social->updateAddressbooks('any', $userId, $offsetBook, $offsetContact);
@@ -53,6 +53,16 @@ class SocialUpdate extends \OC\BackgroundJob\QueuedJob {
 			$report = $result->getData();
 			$stoppedAtBook = $report[0]['stoppedAt']['addressBook'];
 			$stoppedAtContact = $report[0]['stoppedAt']['contact'];
+
+			// make sure the offset contact/address book are still existing
+			if ($this->social->existsAddressBook($stoppedAtBook, $userId) == false) {
+				$stoppedAtBook = null;
+			}
+			if ($this->social->existsContact($stoppedAtContact, $stoppedAtBook, $userId) == false) {
+				$stoppedAtContact = null;
+			}
+			// TODO: can we check the userId still exists?
+
 			$this->jobList->add(self::class, [
 				'userId' => $userId,
 				'offsetBook' => $stoppedAtBook,
--- a/lib/Cron/SocialUpdate.php
+++ b/lib/Cron/SocialUpdate.php
@@ -48,7 +48,7 @@ class SocialUpdate extends \OC\BackgroundJob\QueuedJob {
 		// update contacts with first available social media profile
 		$result = $this->social->updateAddressbooks('any', $userId, $offsetBook, $offsetContact);
 
-		if ($result->getStatus() == Http::STATUS_PARTIAL_CONTENT) {
+		if ($result->getStatus() === Http::STATUS_PARTIAL_CONTENT) {
 			// not finished; schedule a follow-up
 			$report = $result->getData();
 			$stoppedAtBook = $report[0]['stoppedAt']['addressBook'];
--- a/lib/Service/SocialApiService.php
+++ b/lib/Service/SocialApiService.php
@@ -131,9 +131,9 @@ class SocialApiService {
 	 *
 	 * @returns {IAddressBook} the corresponding addressbook or null
 	 */
-	protected function getAddressBook(string $addressbookId, $manager=false) : ?IAddressBook {
+	protected function getAddressBook(string $addressbookId, IManager $manager = null) : ?IAddressBook {
 		$addressBook = null;
-		if ($manager == false) {
+		if ($manager === null) {
 			$manager = $this->manager;
 		}
 		$addressBooks = $manager->getUserAddressBooks();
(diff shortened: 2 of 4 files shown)


call-me-matt:
Unfortunately addressBook->search does not return the contacts in same order (checking with the logger). Or am I doing something wrong?
So a new method is needed here. Options I see:
i. a counter for addressbook and contact if they are always returned in the same order [failed]
ii. passing the array of treated contacts with the call of the update function and skipping contacts that are part of the list [might be a huge parameter list]
iii. introducing a new database column for the cards with a timestamp of last social check [might introduce incompatibilities with other functions if messing with the table]
iv. introducing a new database table to keep track of the sync status

call-me-matt:
I removed the optimization, so we can merge this.

skjnldsv (approved):

skjnldsv:
@rullzer @ChristophWurst @nickvergessen can I have a backgroundjob expert to do a quick review? :pray:

lib/Service/SocialApiService.php
```diff
@@ -254,6 +259,33 @@ protected function registerUpdateResult(array $report, string $entry, string $st
 		return $report;
 	}
 
+	/**
+	 * @NoAdminRequired
```
nickvergessen, line 263: This only works on controllers. Remove it from all other files please

lib/Cron/SocialUpdate.php
```diff
@@ -25,18 +25,39 @@
...
+	public function __construct(SocialApiService $social,
+								IJobList $jobList) {
 		$this->social = $social;
+		$this->jobList = $jobList;
 	}
 
 	protected function run($arguments) {
 		$userId = $arguments['userId'];
+		$offsetBook = $arguments['offsetBook'];
+		$offsetContact = $arguments['offsetContact'];
```
nickvergessen, line 46: This could be undefined from an old job? In that case add `?? null;`

lib/Service/SocialApiService.php
```diff
@@ -290,25 +324,55 @@ public function updateAddressbooks(string $network, string $userId) : JSONRespon
...
+				if (($this->timeFactory->getTime() - $startTime) > 15) {
+					$response['stoppedAt'] = [
+						'addressBook' => $addressBook->getURI(),
+						'contact' => $contact['UID'],
+					];
+					return new JSONResponse([$response], Http::STATUS_PARTIAL_CONTENT);
+				}
+
+				// delay to prevent rate limiting issues
+				sleep($delay);
```
nickvergessen, line 375: This feels weird. So you sleep here but execute at most 15seconds?

nickvergessen (changes requested):

lib/Cron/SocialUpdate.php
call-me-matt: Sorry, I am not sure I understand. I could verify if the book/contact are actually existing and if they don't, reset them. Is that what you mean?

lib/Service/SocialApiService.php
call-me-matt: Well, this delay is to be compliant to the terms of service of social networks, so I don't want to remove it. And the 15seconds of maximum execution time are to prevent that the update runs forever on huge installations.
How would you propose to tackle this?

lib/Cron/SocialUpdate.php
nickvergessen: Yeah that is what I'm suggesting.
`$a['b'] ?? null` is short for `isset($a['b']) ? $a['b'] : null`

call-me-matt: I have added this plus some more sanity checks, as the contact or address book that are remembered may be deleted between the execution of two chunks.

lib/Cron/SocialUpdate.php
```diff
@@ -25,18 +25,49 @@
...
+		$result = $this->social->updateAddressbooks('any', $userId, $offsetBook, $offsetContact);
+
+		if ($result->getStatus() == Http::STATUS_PARTIAL_CONTENT) {
+			// not finished; schedule a follow-up
+			$report = $result->getData();
+			$stoppedAtBook = $report[0]['stoppedAt']['addressBook'];
+			$stoppedAtContact = $report[0]['stoppedAt']['contact'];
+
+			// make sure the offset contact/address book are still existing
+			if ($this->social->existsAddressBook($stoppedAtBook, $userId) == false) {
```
nickvergessen, line 58:
```suggestion
			if ($this->social->existsAddressBook($stoppedAtBook, $userId)) {
```

lib/Cron/SocialUpdate.php
```diff
@@ -25,18 +25,49 @@
...
+				$stoppedAtBook = null;
+			}
+			if ($this->social->existsContact($stoppedAtContact, $stoppedAtBook, $userId) == false) {
```
nickvergessen, line 61:
```suggestion
			if ($this->social->existsContact($stoppedAtContact, $stoppedAtBook, $userId)) {
```

lib/Cron/SocialUpdate.php
```diff
@@ -25,18 +25,49 @@
...
 	protected function run($arguments) {
 		$userId = $arguments['userId'];
+		$offsetBook = $arguments['offsetBook'] ?? null;
+		$offsetContact = $arguments['offsetContact'] ?? null;
 
 		// update contacts with first available social media profile
-		$this->social->updateAddressbooks('any', $userId);
+		$result = $this->social->updateAddressbooks('any', $userId, $offsetBook, $offsetContact);
+
+		if ($result->getStatus() == Http::STATUS_PARTIAL_CONTENT) {
```
nickvergessen, line 51:
```suggestion
		if ($result->getStatus() === Http::STATUS_PARTIAL_CONTENT) {
```

lib/Service/SocialApiService.php
```diff
@@ -123,17 +124,19 @@ protected function addPhoto(array &$contact, string $imageType, string $photo) {
...
 	 *
 	 * @returns {IAddressBook} the corresponding addressbook or null
 	 */
-	protected function getAddressBook(string $addressbookId) : ?IAddressBook {
+	protected function getAddressBook(string $addressbookId, $manager=false) : ?IAddressBook {
 		$addressBook = null;
-		$addressBooks = $this->manager->getUserAddressBooks();
+		if ($manager == false) {
+			$manager = $this->manager;
+		}
```
nickvergessen, line 138:
```suggestion
	protected function getAddressBook(string $addressbookId, IManager $manager = null) : ?IAddressBook {
		$addressBook = null;
		if ($manager === null) {
			$manager = $this->manager;
		}
```

lib/Service/SocialApiService.php
```diff
@@ -217,8 +216,51 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
 	/**
-	 * @NoAdminRequired
+	 * checks an addressbook is existing
+	 *
+	 * @param {string} searchBookId the UID of the addressbook to verify
+	 * @param {string} userId the user that should have access
+	 *
+	 * @returns {bool} true if the addressbook exists
+	 */
+	public function existsAddressBook($searchBookId, $userId) {
```
nickvergessen, line 226:
```suggestion
	public function existsAddressBook(string $searchBookId, string $userId): bool {
```

lib/Service/SocialApiService.php
```diff
@@ -217,8 +216,51 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+	/**
+	 * checks a contact exists in an addressbook
+	 *
+	 * @param {string} searchContactId the UID of the contact to verify
+	 * @param {string} searchBookId the UID of the addressbook to look in
+	 * @param {string} userId the user that should have access
 	 *
+	 * @returns {bool} true if the contact exists
+	 */
+	public function existsContact($searchContactId, $searchBookId, $userId) {
```
nickvergessen, line 246:
```suggestion
	public function existsContact(string $searchContactId, string $searchBookId, string $userId): bool {
```

lib/Service/SocialApiService.php
```diff
@@ -217,8 +216,51 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+		}
+		return true;
+	}
+
+	/**
+	 * checks a contact exists in an addressbook
+	 *
+	 * @param {string} searchContactId the UID of the contact to verify
+	 * @param {string} searchBookId the UID of the addressbook to look in
+	 * @param {string} userId the user that should have access
```
nickvergessen, line 242:
```suggestion
	 * @param string searchContactId the UID of the contact to verify
	 * @param string searchBookId the UID of the addressbook to look in
	 * @param string userId the user that should have access
```

lib/Service/SocialApiService.php
```diff
@@ -217,8 +216,51 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
 	 *
+	 * @returns {bool} true if the contact exists
```
nickvergessen, line 244:
```suggestion
	 * @returns bool true if the contact exists
```

We don't use `{` in php for those, but it seems to be an overall problem in this app.

lib/Service/SocialApiService.php
```diff
@@ -217,8 +216,51 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+	 */
+	public function existsAddressBook($searchBookId, $userId) {
+		$manager = $this->manager;
+		$coma = new ContactsManager($this->davBackend, $this->l10n);
+		$coma->setupContactsProvider($manager, $userId, $this->urlGen);
+		$addressBooks = $manager->getUserAddressBooks();
+		if ($this->getAddressBook($searchBookId, $manager) == null) {
+			return false;
+		}
+		return true;
```
nickvergessen, line 234:
```suggestion
		return $this->getAddressBook($searchBookId, $manager) !== null;
```

lib/Service/SocialApiService.php
```diff
@@ -217,8 +216,51 @@ public function updateContact(string $addressbookId, string $contactId, string $
...
+		$addressBook = $this->getAddressBook($searchBookId, $manager);
+		if ($addressBook == null) {
+			return false;
+		}
+
+		$check = $addressBook->search($searchContactId, ['UID'], ['types' => true]);
+		if (empty($check)) {
+			return false;
+		}
+		return true;
```
nickvergessen, line 260:
```suggestion
		return !empty($check);
```

nickvergessen (approved):
Looks good, but I didn't execute it

lib/Cron/SocialUpdate.php
call-me-matt: Thank you for your review and good comments!
I suppose this should now start with `if(!$this` ? Or was the negation done on purpose?
(same for the next suggested change)

nickvergessen:
Ready to go I guess?

PR #1755, opened 2020-08-21
Implement masonry

### LINKED ISSUE #654: Layout sometimes broken
broken

As you can see the phone numbers seem to be spread of two columns. Which is odd and looks out of place.

If I zoom in a bit all is good

good

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64036836-layout-sometimes-broken?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
Fix #654 

Capture d’écran_2020-08-21_17-21-42

Peek 21-08-2020 17-04


jotoeri:
Hm, your gif looks good, i'm just not enough into this code so i could judge anything there... 🤷‍♂️

skjnldsv:
> Hm, your gif looks good, i'm just not enough into this code so i could judge anything there...

You can try and see if it behave properly? :pray: :bow:

jotoeri:
Hm, the column sorting is a bit strange... Changing from two to three columns, some pads jump from third to first column and correspondingly, when changing between two contacts, the pads can be sorted completely different, if just one small field is missing... That's quite confusing on usage...
contacts

Sidebar hiding and reappearing on resize is also a bit strange in timing and overlays, but thats different thing i think?

jancborchardt:
Yeah, I was thinking similarly – e.g. "Add property" should always be the last entry and "Address book" second to last (and not show at all if there’s only one address book).

The question is, what is "last" in this layout. I would tend to say the most bottom right one? It is definitely strange when "Address book" or "Add new property" jump to the first column before some of the actual data.

skjnldsv:
> Yeah, I was thinking similarly – e.g. "Add property" should always be the last entry and "Address book" second to last (and not show at all if there’s only one address book).
> 
> The question is, what is "last" in this layout. I would tend to say the most bottom right one? It is definitely strange when "Address book" or "Add new property" jump to the first column before some of the actual data.

I'll see if I can do that!
Masonry is auto balancing between columns, so if I can't, we'll at least be sure they are in every column's end. I'd say it's still an enhancement than splitting same properties into multiple columns?

skjnldsv:
> Sidebar hiding and reappearing on resize is also a bit strange in timing and overlays, but thats different thing i think?

Not sure what you mean by this? :)

jotoeri:
> Not sure what you mean by this? :)

- When making the window smaller, first the sidebar disappears, which is good, i think. But when enlarging again, also first the sidebar appears, while the contacts-details are still in one column, which looks a bit strange imo. (Ok, I just realised that depends on how fast one changes the window. Masonry just seems to be slower than the sidebar to react...)
contacts2

- There is some intermediate window-size, where the contacts-sidebar is hidden by the category-sidebar, when not specifcally closing the category-sidebar. -> Not connected to this PR...
contacts3

jotoeri:
> Not sure what you mean by this? :)
- There is some intermediate window-size, where the contacts-sidebar is hidden by the category-sidebar, when not specifcally closing the category-sidebar. -> Not connected to this PR...
contacts3

skjnldsv:
> * Masonry just seems to be slower than the sidebar to react

Yes, the masonry script is on a debounce to ease the calc, otherwise it's too much fro the browser :)

rullzer (approved):
Lets do this

ChristophWurst (approved):

ma12-co (approved):

PR #1788, opened 2020-09-06
Properly sort groups

### LINKED ISSUE #1598: Enhancement for alphabetic sorting of groups if there are numbers
As a follow-up to https://github.com/nextcloud/contacts/pull/1461 by @ImUrX, when group names start with numbers, the sorting is unexpected:
Contact groups number sorting

Expected would be by value, not by first character:
- 5
- 20
- 50
- 150

Seems a compareFunction is needed for that: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

skjnldsv:
Fix #1598
| Before | After |
|:---------:|:------:|
|dev skjnldsv com_apps_contacts_1contacts|dev skjnldsv com_apps_contacts_ (1)|

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3665,7 +3665,7 @@
 		},
 		"cdav-library": {
 			"version": "git+https://github.com/nextcloud/cdav-library.git#a9642c7c83db37120f8c74306c7e761f3e78524d",
-			"from": "git+https://github.com/nextcloud/cdav-library.git#a9642c7c83db37120f8c74306c7e761f3e78524d",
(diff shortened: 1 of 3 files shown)


call-me-matt (approved):

call-me-matt:
While checking, I stumbled upon this:
Is that a known issue or a bug with my deployment?
grafik

skjnldsv:
> While checking, I stumbled upon this:
> Is that a known issue or a bug with my deployment?
> grafik
> 

Yes :)
Fixed in vue components

PR #1828, opened 2020-10-07
Fix placeholder display

### LINKED ISSUE #1812: email address not shown in minimal contact in Contacts 3.4.0
### Describe the bug
Thunderbird CardBook added a contact to my NC for a mail address that was unknown. When I displayed that contact in contacts web UI, I didn't see an email address – which is odd because CardBook would only add the contact (to its "collected new contacts" group) if id had an address.

So there are at least cases that a valid email address isn't shown in contact details. It is shown in the contacts list however, so it's parsed correctly.

I was able to reproduce the bug by importing the attached file, which I exported through the contacts web app, redacted some entries using a text editor.

**To Reproduce**
1. validate the attached file, should be valid vCard 3.0
2. import the attached file in Contacts web app
3. reload Contacts by clicking on it in NC headline menu
4. select test contact to display its content

**Expected behavior**
Email address should be visible in the contacts list AND in the details.

**Actual behavior**
Email address is visible in the contacts list but NOT in the details.

**vCard**
```
BEGIN:VCARD
VERSION:3.0
PRODID:-//vi/CardBook V51.7//EN
UID:2ab1f9ac-cf5c-718d-f5e8-61ac0e6843cc
FN:test
EMAIL:test@domain.org
REV:2020-09-30T18:18:00Z
END:VCARD
```

**Screenshot**
NoEmail

### Configuration
Arch Linux (current), NGINX 1.18, PHP-FPM 7.4, MariaDB 10.5
Nextcloud 19.0.3.1, Contacts 3.4.0, NC: updated, Signing: OK
Browser: Firefox 81.0 on Arch Linux (current), CardDAV-clients: TB CardBook (yet not relevant here!)

skjnldsv:
fix #1812

### DIFF
--- a/src/components/Properties/PropertyText.vue
+++ b/src/components/Properties/PropertyText.vue
@@ -72,7 +72,7 @@
 				:class="{'property__value--with-ext': haveExtHandler}"
 				type="text"
 				class="property__value"
-				:placeholder="localType.placeholder"
+				:placeholder="placeholder"
 				@input="updateValue">
 
 			<!-- external link -->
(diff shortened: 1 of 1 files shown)


call-me-matt (approved):
don't understand the code 100% - but I can confirm the issue is not present anymore with this.

skjnldsv:
> don't understand the code 100% - but I can confirm the issue is not present anymore with this.

We make sure localType is defined now, and if not we check if the property model have a placeholder entry :)

Issue #1107, opened 2019-05-23
Remove quotes from TYPE for better external platforms compatibility


ghost:
This is a resubmit as I ended up doing more investigation following a switch to using Google Contacts due to this problem - during the switch, I had to manually edit the VCF file to get it to work with Gmail.

The VCF file produced by Contacts doesn't use fields recognised by other applications and appears to be non-standard. As it currently stands, it appears Contacts is incompatible with all other major platforms.

Nextcloud 16.0.1, Contacts 3.1.1

**Example:**
1. New contact created in the web app, with a phone number and email. The type (e.g. home, office) is left on the defaults.
![image](https://user-images.githubusercontent.com/665111/58234425-648a4500-7d36-11e9-8cc4-5e9a05da090c.png)

2. The contact is downloaded through the web app (.VCF file).

3. The file is opened in Outlook, revealing a missing phone number.
![image](https://user-images.githubusercontent.com/665111/58234618-debac980-7d36-11e9-9353-f5d419235490.png)

4. The file is imported to gmail, again revealing a missing phone number.
![image](https://user-images.githubusercontent.com/665111/58234695-0ad64a80-7d37-11e9-8534-2f4833c71393.png)
![image](https://user-images.githubusercontent.com/665111/58234718-1b86c080-7d37-11e9-8465-96e4c071b494.png)

5. Sync to iOS also reveals a missing phone number.
![image](https://user-images.githubusercontent.com/665111/58234836-64d71000-7d37-11e9-8dd5-c271179c12e3.png)

6. Opening the VCF in ATOM reveals the field names. You can see email has the type HOME but phone has (quotes included as in the file) "HOME,VOICE".
![image](https://user-images.githubusercontent.com/665111/58234883-8932ec80-7d37-11e9-8d4f-c327d0a1a1e5.png)

7. Removing ,VOICE and the quotation marks to leave the type as HOME results in Gmail and Outlook now seeing the phone number.
![image](https://user-images.githubusercontent.com/665111/58235069-f8a8dc00-7d37-11e9-9dfc-ace69382aeac.png)

skjnldsv:
Right, we should remove the quotations for compatibility. :)

Thanks for reporting!

giddyhup:
This is also happening with the popular productivity/email program eM Client.

skjnldsv:
@giddyhup Maybe, but we really need https://github.com/mozilla-comm/ical.js/issues/343

giddyhup:
[...]

PR #1843, opened 2020-10-10
more robust social ids for social sync

call-me-matt:
- resolve facebook-id from usernames
- accept mastodon userids without leading 'at'
- ignore apple markers as in https://github.com/nextcloud/contacts/issues/1107#issuecomment-706518724

### DIFF
--- a/lib/Service/Social/TumblrProvider.php
+++ b/lib/Service/Social/TumblrProvider.php
@@ -35,6 +35,7 @@ class TumblrProvider implements ISocialProvider {
 	 * @return string
 	 */
 	public function cleanupId(string $candidate):?string {
+		$candidate = preg_replace('/^' . preg_quote('x-apple:', '/') . '/', '', $candidate);
 		$subdomain = '/(?:http[s]*\:\/\/)*(.*?)\.(?=[^\/]*\..{2,5})/i'; // subdomain
 		if (preg_match($subdomain, $candidate, $matches)) {
 			$candidate = $matches[1];
--- a/lib/Service/Social/FacebookProvider.php
+++ b/lib/Service/Social/FacebookProvider.php
@@ -78,17 +78,11 @@ class FacebookProvider implements ISocialProvider {
 			if ($result->getStatusCode() !== 200) {
 				return $profileName;
 			}
-			$htmlResult = new \DOMDocument();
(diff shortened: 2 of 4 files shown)


call-me-matt:
iOS adds an `x-apple` to the value for social profiles. Is this legal?

`X-SOCIALPROFILE;type=CUSTOM;x-user=nextclouders:x-apple:nextclouders`. 

Nextcloud interprets it like this:
apple-marker

If the format from apple is acceptable, this PR filters out the additional `x-apple`.
Alternatively, the contacts app could be adapted to strip the value (which would display things nicer in the web app)

lib/Service/Social/TumblrProvider.php
```diff
@@ -35,6 +35,7 @@ public function __construct() {
 	 * @return string
 	 */
 	public function cleanupId(string $candidate):?string {
+		$candidate = preg_replace('/^' . preg_quote('x-apple:', '/') . '/', '', $candidate);
```
skjnldsv, line 38: Maybe this should go into the level above so $candidate is already sanitized?

skjnldsv: https://github.com/nextcloud/contacts/blob/709615173f3b38dbea2331a530c1edd5bb68f880/lib/Service/Social/CompositeSocialProvider.php#L82

call-me-matt: Sure, it can be there, too. But apple only attaches this to custom fields. For certain networks (for example Facebook and Twitter) it adds the domain of the service. So because it is not valid for all networks, I only added it where necessary. I'll let you decide where it's best.

skjnldsv: Well, I guess it's easier to have it once and for all rather than on every provider?
Maybe I'm missing something :thinking: :wink:

call-me-matt: just to be clear: we do not need it for every provider. apple only adds this "x-apple" thingy to the networks it does not know. All pre-defined networks don't have that tag. So I thought its more a temporary thing and it can be removed time after time from the known networks.

call-me-matt:
big news: I found a better way to get the profile id from facebook for any profile! It works very nicely (until fb changes their site...)

skjnldsv (approved):

Issue #1662, opened 2020-06-11
Crop photo for the contact picture


Spartachetto:
Nextcloud version: 18.0.3
Contacts verion: 3.3.0

I tried to assign a photo to a contact and I had some issues.
You can read part of them here:
https://github.com/nextcloud/nextcloud-vue/issues/631#issuecomment-642947863

This issue is about another thing. When I selected the right photo on my Nextcloud instance, I was expecting to have the possibility to crop the photo I.e. to select a part of the photo with an easy interactive tool. I was accustomed by other websites that I could select a part of the image "zooming " on the person of interest. I am aware that with that operation you lose quality (=you have a worse resolution). Yet it is very practical, otherwise you have to download the photo and create a new file with a image modification program.

call-me-matt:
just a remark with the same context: I find it also confusing that the contact list on the left shows the top part of the picture, while the selected contact's avatar is showing the center part of the image; so you have the same picture side by side being cropped differently.

calvinrw:
I agree with both points completely. On other platforms, a user will typically be asked to zoom and crop a photo before using it as an avatar.

I think the best solution to drop all instances of automatic cropping, beyond what is necessary to generate appropriately-sized thumbnails, in favor of allowing a user to manually edit a photo before assigning it as an avatar.

![Screenshot_2020-06-14_19-59-48](https://user-images.githubusercontent.com/687129/84608825-cc4b8a80-ae79-11ea-85d1-6bccde6fd2dc.png)

Spartachetto:
Nextcloud version: 18.0.3

I found that this functionality is available when you pick a photo for your user profile.
Maybe you could reuse this for contacts...

PR #1846, opened 2020-10-12
Fix avatar display on read-only contacts and use global Avatar component

skjnldsv:
| Before | After |
|:---------:|:------:|
|Kazam_screenshot_00004|Kazam_screenshot_00003|

PR #1854, opened 2020-10-15
harmonize avatars

call-me-matt:
refers to https://github.com/nextcloud/contacts/issues/1662#issuecomment-643137544

I find it confusing that the contact list on the left shows the top part of the picture, while the selected contact's avatar is showing the center part of the image; so you have the same picture side by side being cropped differently:

before

This small change harmonizes both thumbnails by taking the top part of the avatar.

Signed-off-by: call-me-matt <nextcloud@matthiasheinisch.de>

### DIFF
--- a/src/components/ContactDetails/ContactDetailsAvatar.vue
+++ b/src/components/ContactDetails/ContactDetailsAvatar.vue
@@ -475,7 +475,7 @@ export default {
 		// White background for avatars with transparency, also in dark theme
 		background-color: #fff;
 		background-repeat: no-repeat;
-		background-position: center;
+		background-position: top;
 		background-size: cover;
 	}
 


skjnldsv:
Wait for https://github.com/nextcloud/contacts/pull/1846
This also fixes some avatar things :warning: 

But we align to center, we do not crop to top anywhere else on nextcloud :)

call-me-matt:
sounds great!
and you are right, not cropping but aligning here - and chose the top part because that looks better for avatars when you see the face and not the belly ;)

Another issue I came across in the same context: usually profile pictures are small - but if they aren't the modal crops them. Shouldn't we show the complete picture?
modal2

skjnldsv:
> Another issue I came across in the same context: usually profile pictures are small - but if they aren't the modal crops/hides them. Shouldn't we always show the complete picture?

yes, this one is a bug I also noticed :)

call-me-matt:
> Wait for #1846

As far as I can see, it is not treating this issue?

skjnldsv:
> This also fixes some avatar things

I didn't said this was fixing the same thing, but the other one is fixing things that I want to get into a new release asap :)

skjnldsv:
Closing, let's fix in components :)

call-me-matt:
Where exactly can I address this?
Latest version of contacts makes this even worse: cutting the heads in the details and modal, plus squeezing pictures without respecting aspect ratio in the list. Looks very awkward now for contact pictures that are not squared.

skjnldsv:
The modal one needs to be addressed here though :+1: 
I closed this because I fixed the avatars in the vue lib ;)

jancborchardt:
Sorry for the late reply – just because this also came up in the design review of the Contacts app: We should have a cropper in the "Upload new picture" process which is fixed to cropping a square, right? Other contacts apps like e.g. on Android have that as well, and it prevents this issue.

Issue #1868, opened 2020-10-20
Download contact pictures from Xing


ntimo:
**Is your feature request related to a problem? Please describe.**
When setting Xing as a social media profile for contacts, the profile picture is not automatically downloaded and updated.

**Describe the solution you'd like**
Nextcloud contacts should update profile pictures using Xing.

call-me-matt:
I would love to implement this! In order to do so, I need a "recipe" (or public API) to fetch the profile picture from there. I found those for facebook, instagram, twitter, tumblr, mastodon, and diaspora. But xing and linkedin seem to lock their network from external access.

call-me-matt:
ah well, turns out the profile pictures are public. yay!

call-me-matt:
@skjnldsv should Xing be part of the selection-list?
`src/models/rfcProps.js:		{ id: 'XING', name: 'Xing', placeholder: 'https://www.xing.com/profile/…' },`

skjnldsv:
> @skjnldsv should Xing be part of the selection-list?
> `src/models/rfcProps.js:		{ id: 'XING', name: 'Xing', placeholder: 'https://www.xing.com/profile/…' },`

I guess? Hard to tell, those properties are most likely not super standard 😬

call-me-matt:
> I guess?

I'll let you decide. Also if that's part of the above or a separate PR...

ntimo:
Thanks for adding this so fast :)

skjnldsv:
Yes, thanks @call-me-matt :rocket: :hugs:

PR #1871, opened 2020-10-21
add xing to list of social networks

call-me-matt:
shall we add Xing to the list to select from?
see https://github.com/nextcloud/contacts/issues/1868#issuecomment-713151207


skjnldsv (approved):
Yep!

Issue #1820, opened 2020-10-01
Picture uploded in Web GUI are not synchronized anymore in Android with DAVx5 3.3.1-ose with contacts 3.4.0


theyoungstone:
### Describe the bug
I have just updated contacts to 3.4.0 version.
In the Web GUI, I upload an new picture for a contact.
When I synchronize the contract on my Android 10 smarphone with DAVx5 3.3.1-ose, the old picture is deleted but the new one does not appear. So the contact has not picture anymore.

**To Reproduce**
1. Go into a contact card into NextCloud Web GUI with a browser
2. Click on "Add a new picture"
3. Choose an picture
4. On Android device, synchronize contacts with DAVx5
5. Open the contact card on the Android device
6. There is no picture anymore

**Expected behavior**
We should see the new picture uploaded into the Web GUI

**Actual behavior**
The is no picture as if it was deleted

### Server configuration

**Operating system**: Linux Debian 10 up to date

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud version:** 19.0.3

**Contacts version:** 3.4.0

**Updated from an older Nextcloud or fresh install:** Updated

**Signing status:**
```
Login as admin user into your Nextcloud and access 
http://example.com/index.php/settings/integrity/failed 
No errors have been found.

**List of activated apps:**
Enabled:
  - accessibility: 1.5.0
  - activity: 2.12.0
  - bruteforcesettings: 2.0.1
  - calendar: 2.0.4
  - cloud_federation_api: 1.2.0
  - comments: 1.9.0
  - contacts: 3.4.0
  - contactsinteraction: 1.0.0
  - dav: 1.15.0
  - deck: 1.0.5
  - federatedfilesharing: 1.9.0
  - federation: 1.9.0
  - files: 1.14.0
  - files_pdfviewer: 1.8.0
  - files_rightclick: 0.16.0
  - files_sharing: 1.11.0
  - files_trashbin: 1.9.0
  - files_versions: 1.12.0
  - files_videoplayer: 1.8.0
  - firstrunwizard: 2.8.0
  - gpxedit: 0.0.13
  - gpxmotion: 0.0.11
  - gpxpod: 4.2.2
  - groupfolders: 7.0.0
  - logreader: 2.4.0
  - lookup_server_connector: 1.7.0
  - mail: 1.4.1
  - nextcloud_announcements: 1.8.0
  - notes: 3.6.4
  - notifications: 2.7.0
[...]

PR #1918, opened 2020-11-07
vcard should only have 1 photo field when adding/removing photos

eleith:
when adding photo fields through the social api service, an extra photo
field is created for vcard version 3.0 due to setting a field as empty
instead of unsetting it all together

when removing photo fields, this empty photo field is left untouched

thus, syncing photos to address book clients can be buggy as most
address book clients do not properly iterate through all photo fields to
display the contact photo

closes: https://github.com/nextcloud/contacts/issues/1820


skjnldsv (approved):
That makes total sense!
And with tests too? :heart_eyes: Yes please! :hugs:

Issue #1891, opened 2020-10-27
Incorrect image from Facebook


Thovi98:
### Steps to reproduce
1. Setup Facebook url
2. Get image from Facebook

### Expected behaviour
Have the image of the profil of the url

### Actual behaviour
The new image is the the default avatar of Facebook (so not exactly an image). I have tried with different profiles, it’s always the same. And I tried with Instagram which works normaly. 

When I try to get from Facebook for first time
![2020-10-27 18_57_15-](https://user-images.githubusercontent.com/51749973/97343782-380e1d00-1888-11eb-9919-743e7e14f776.jpg)

When I try to update from Facebook an already updated Facebook image
![2020-10-27 18_57_42-](https://user-images.githubusercontent.com/51749973/97343785-38a6b380-1888-11eb-831d-7a04c2c306e8.jpg)

## Server configuration detail

**Operating system:** Linux 5.8.0-0.bpo.2-amd64 #1 SMP Debian 5.8.10-1~bpo10+1 (2020-09-26) x86_64

**Webserver:** nginx/1.14.2 (fpm-fcgi)

**Database:** mysql 10.3.25

**PHP version:** 

7.3.23-4+0~20201018.71+debian10~1.gbpfc8934
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, sodium, session, standard, cgi-fcgi, mysqlnd, PDO, xml, apcu, bcmath, bz2, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, gmp, iconv, igbinary, imagick, imap, intl, json, ldap, exif, mysqli, pdlib, pdo_mysql, apc, posix, readline, redis, shmop, SimpleXML, smbclient, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, zip, Phar, libsmbclient, Zend OPcache

**Nextcloud version:** 20.0.1 - 20.0.1.1

**Updated from an older Nextcloud/ownCloud or fresh install:** Updated from 19.04

**Where did you install Nextcloud from:** Yunohost (https://github.com/yunohost-apps/nextcloud_ynh)

<details><summary>Signing status</summary>

Integrity checker has been disabled. Integrity cannot be verified.
</details>

<details><summary>List of activated apps</summary>

```
Enabled:
 - accessibility: 1.6.0
 - audioplayer: 2.12.0
 - calendar: 2.1.2
[...]

PR #1920, opened 2020-11-08
Fix/facebook lowres

call-me-matt:
as facebook depreciated their API, here is a try to retrieve the public available thumb. Quality is bad, reliability too, still better than nothing. Moved to last priority when fetching different sources.

fixes partly #1891


skjnldsv (commented):
Like mentioned, please use the meta tag with the incorporated image, that is reliable as this is what is used for sharing context creation :)
It feels less hacky and resolution is somewhat decent (200x200)

It works for people that set their visibility to public.
For others, there is nothing we can do :/

skjnldsv:
Also, labels please! :see_no_evil:

lib/Service/Social/FacebookProvider.php
```diff
@@ -59,25 +59,54 @@ public function supportsContact(array $contact):bool {
...
+	 * @param {string} profile url
+	 *
+	 * @return string|null
+	 */
+	protected function getImageUrl(string $profileUrl):?string {
+		try {
+			$result = $this->httpClient->get($profileUrl);
+			$htmlResult = $result->getBody();
+
+			$avatar = '/meta property="og:image" content="(http[^"]+)/';
```
call-me-matt, line 85: the layout changed and I cannot figure out how to get the picture. but it seems still to be available without login.

ChristophWurst:
I'm closing this due to inactivity. @call-me-matt if this is still a reasonable fix for the facebook images then feel free to reopen and update your branch so it can get a round of review and testing.

PR #1936, opened 2020-11-15
Replace the deprecated OC.Util.relativeModifiedDate by nextcloud/moment

tintou:
As per https://github.com/nextcloud/server/pull/20988 OC.Util.relativeModifiedDate is deprecated.


skjnldsv (approved):
:rock
Nice!! :)

Issue #609, opened 2018-08-31
[Vue] GROUP vcard support


skjnldsv:
- [ ] Replace categories
- [ ] Group edition support
- [ ] Option in settings (for now, to be dropped in nc 15)

olantrust:
Does this mean that user will not be able to choose between using Group for categories and contact groups?

skjnldsv:
@olantrust what do you mean? You mean using the categories as *tags* for contacts?

nalt:
To back this on Bountysource: Will this reliably solve groups in MacOS, see #213?

skjnldsv:
> To back this on Bountysource: Will this reliably solve groups in MacOS, see #213?

yes :)
Those are both the same feature.

zlajo:
Can someone provide a hint where and how to implement this? I found out how the "GROUP"-feature is supposed to work but I don't know where the serialization into the vCard-Format is implemented.

Thanks!

skjnldsv:
@zlajo this requires a major change on how Groups are managed on this app.

zlajo:
@skjnldsv Wouldn't it be sufficient to just update the serialization for categories? How strict is the vCard-standard?

skjnldsv:
> @skjnldsv Wouldn't it be sufficient to just update the serialization for categories? How strict is the vCard-standard?

What's wrong with the current categories?

zlajo:
@skjnldsv That's what I would like to find out. It just would be much easier with some background information. Obviously someone already put some thought into that matter.

skjnldsv:
So, a bit of history then ;)
CATEGORIES was never there to be used for groups, it was designed to be labels.
There was never a real "group" function in the pre-vcard4 era.

So when vcard4 came out they created a dedicated way of having a group as a vcard. This makes it easier as you can now update groups without touching every member. You can just edit the vcard.

Unfortunately not all systems supports vcard 4 groups.
MacOS switched to it not so long ago, and davx supports it but asks you what kind of groups is your carddav server using (group as individual card or group within vcards)

I have no idea on other vcard manager. Before diving into this, we should make sure the most used software here supports it so the integration doesn't break.
Thunderbird, cardbook... I don't know how much are people using.
As well as iOS of course. 

So we went for the easier way and how we used to do it on previous contacts versions.
But moving to this means changing how we read GROUPS in the app and keep CATEGORIES as individual labels per contacts.

zlajo:
[...]

PR #1999, opened 2020-12-30
Improve Compatibily with macOS Contacts

zlajo:
I would like to use nextcloud contacts as a replacement to storing my contacts in iCloud. Unfortunately there are some things that don't work as expected and while some issues require fundamental changes (like [#609](https://github.com/nextcloud/contacts/issues/609)) there are some which I was able to fix with only a few small changes.

One thing I found out was that there already is support for hiding contact groups (as defined by [https://tools.ietf.org/html/rfc6350#section-6.1.4](https://tools.ietf.org/html/rfc6350#section-6.1.4)). macOS Contacts still uses vCard 3.0 and because of that cannot relay on the vCard kind property defined by vCard 4.0. But there is a proprietary attribute which does the same thing (X-ADDRESSBOOKSERVER-KIND). I updated the contacts app to also respect this Apple specific property and added an easy way to add further proprietary "kind"-properties. 

There is also a problem (at least I think so) with the underlying vCard serialization which leads to strange property names. See my comment in [src/store/contacts.js](https://github.com/zlajo/contacts/blob/e87629a46721a54c1ac50d7441ff87673e5a961d/src/store/contacts.js#L30) for a detailed description of the problem and a potential workaround. 

Please let me know if my update works and contact me if further changes/discussions are required.

### DIFF
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -37,6 +37,12 @@ const isEmpty = value => {
 	return (Array.isArray(value) && value.join('') === '') || (!Array.isArray(value) && value === '')
 }
 
+export const ContactKindProperties = ['KIND', 'X-ADDRESSBOOKSERVER-KIND']
+
+export const MinimalContactProperties = [
+	'EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME',
+].concat(ContactKindProperties)
+
 export default class Contact {
 
 	/**
@@ -79,7 +85,6 @@ export default class Contact {
 			const rev = new ICAL.VCardTime(null, null, 'date-time')
 			rev.fromUnixTime(Date.now() / 1000)
 			this.vCard.addPropertyWithValue('rev', rev)
-
 		}
 	}
 
@@ -300,7 +305,13 @@ export default class Contact {
 	 * @memberof Contact
 	 */
 	get kind() {
-		return this.firstIfArray(this.vCard.getFirstPropertyValue('kind'))
+		return this.firstIfArray(
+			ContactKindProperties
+				.map(s => s.toLowerCase())
+				.map(s => this.vCard.getFirstPropertyValue(s))
+				.flat()
+				.filter(k => k)
+		)
 	}
 
 	/**
--- a/src/store/addressbooks.js
+++ b/src/store/addressbooks.js
@@ -25,7 +25,7 @@ import { showError } from '@nextcloud/dialogs'
 import pLimit from 'p-limit'
 import Vue from 'vue'
 
-import Contact from '../models/contact'
+import Contact, { MinimalContactProperties } from '../models/contact'
 
 import client from '../services/cdav'
 import parseVcf from '../services/parseVcf'
@@ -346,8 +346,7 @@ const actions = {
 	 */
 	async getContactsFromAddressBook(context, { addressbook }) {
 		return addressbook.dav
-			.findAllAndFilterBySimpleProperties(['EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N',
-				'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME'])
+			.findAllAndFilterBySimpleProperties(MinimalContactProperties)
 			.then((response) => {
 				// We don't want to lose the url information
 				// so we need to parse one by one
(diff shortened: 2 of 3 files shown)


src/models/contact.js
```diff
@@ -39,6 +39,15 @@ const isEmpty = value => {
...
 export default class Contact {
 
+	static get ContactKindProperties() {
+		return ['KIND', 'X-ADDRESSBOOKSERVER-KIND']
+	}
+
+	static get MinimalContactProperties() {
+		return ['EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME']
+			.concat(Contact.ContactKindProperties)
+	}
```
skjnldsv, line 49: I would prefer this as module export constants please :)

src/store/addressbooks.js
```diff
@@ -346,8 +346,7 @@ const actions = {
 	 */
 	async getContactsFromAddressBook(context, { addressbook }) {
 		return addressbook.dav
-			.findAllAndFilterBySimpleProperties(['EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N',
-				'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME'])
+			.findAllAndFilterBySimpleProperties(Contact.MinimalContactProperties)
```
skjnldsv, line 349: Would be much cleaner! Nice! :+1:

src/store/contacts.js
```diff
@@ -27,6 +27,23 @@ import Vue from 'vue'
...
+ * 'TEL;TYPE="HOME,VOICE":0815 123456', which makes the value appear as a single value
+ * containing a comma instead of two separate values. By forcing all values being escaped by
+ * double quotes the string serialization can be fixed.
+ *
+ * However there is a pull request (https://github.com/mozilla-comm/ical.js/pull/460) waiting
+ * to be merged for ical.js. Until this fix is merged and released the following configuration
+ * changes apply the workaround described above.
+ */
+ICAL.design.vcard3.param.type.multiValueSeparateDQuote = true
+ICAL.design.vcard.param.type.multiValueSeparateDQuote = true
```
skjnldsv, line 45: Nice! I was not aware of such possibilities :)

skjnldsv (changes requested):
See comments :)

src/store/contacts.js
skjnldsv: Ah, yes, it was on the old designSet update workaround! Nice! :+1:

zlajo:
@skjnldsv I have just updated my pull request.

src/models/contact.js
```diff
@@ -37,6 +37,17 @@ const isEmpty = value => {
...
+const ContactKindProperties = ['KIND', 'X-ADDRESSBOOKSERVER-KIND']
+
+const MinimalContactProperties = [
+	'EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME',
+].concat(ContactKindProperties)
+
+export {
+	ContactKindProperties,
+	MinimalContactProperties,
+}
```
skjnldsv, line 49:
```suggestion
export const ContactKindProperties = ['KIND', 'X-ADDRESSBOOKSERVER-KIND']

export const MinimalContactProperties = [
	'EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME',
].concat(ContactKindProperties)
```

skjnldsv (commented):
One last cleanup on the code :+1: 
Looks great! Nice job :)

skjnldsv (approved):
Good for me :)
Please rebase and squash your commits into a single one before merging! :rocket:

jancborchardt:
@zlajo only the Developer Code of Origin test fails, could you sign off your commit? :) You just have to `git commit --amend --signoff` – more info at https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff

jancborchardt:
By the way @zlajo we have a Nextcloud Talk instance for easier communication – we also have a channel for Groupware/Contacts there. If you like, we can invite you? :)

zlajo:
@jancborchardt I just did a rebase on the current master and a force push. Is this what I had to do?

skjnldsv (approved):

PR #2011, opened 2021-01-10
Regression Tests to Ensure Stability of Contact Grouping

zlajo:
@skjnldsv In order to ensure that my updates regarding #609 don't introduce unpredictable behaviour I decided to introduced some regression tests. Since there were no javascript tests I had to introduce some new libraries as well. In order to run the tests you need a clean nextcloud instance and some configuration (tests/config.js; see tests/config.sample.js for an example).

Please let me know what you think and which further tests have to be added to protect critical functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,9 @@
 		"lint": "eslint --ext .js,.vue src",
 		"lint:fix": "eslint --ext .js,.vue src --fix",
 		"stylelint": "stylelint src",
-		"stylelint:fix": "stylelint src --fix"
+		"stylelint:fix": "stylelint src --fix",
+		"test": "mocha --timeout 30000 tests/js/*",
+		"test:lint": "eslint tests/js"
 	},
 	"dependencies": {
 		"@mattkrick/sanitize-svg": "^0.3.1",
@@ -82,6 +84,7 @@
 		"@nextcloud/webpack-vue-config": "^4.0.3",
 		"babel-eslint": "^10.1.0",
 		"babel-loader": "^8.2.2",
+		"chai": "^4.2.0",
 		"css-loader": "^4.3.0",
 		"eslint": "^7.25.0",
 		"eslint-config-standard": "^16.0.2",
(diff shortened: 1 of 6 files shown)


skjnldsv:
Oh wow! :scream:

skjnldsv:
> Please let me know what you think and which further tests have to be added to protect critical functionality.

Well, I always dreamed of having enough time to add testing here,, but well...!
Thank you so much for starting this, I guess that opens a big door for anyone to add more and more tests! Thank you so much! :rocket: :heart_eyes:

package.json
```diff
@@ -30,7 +30,9 @@
 		"lint": "eslint --ext .js,.vue src",
 		"lint:fix": "eslint --ext .js,.vue src --fix",
 		"stylelint": "stylelint src",
-		"stylelint:fix": "stylelint src --fix"
+		"stylelint:fix": "stylelint src --fix",
+		"test": "node_modules/.bin/mocha --timeout 30000 tests/js/*",
```
skjnldsv, line 34: You should be able to just use `mocha` ?
Any reason you used a relative node_modules path ?

skjnldsv (commented):
One comment.
Other than this, we need to add this to the CI: https://github.com/nextcloud/contacts/blob/4bd46d656fb9e21becc23b069180c09fa3dcb112/.github/workflows/node.yml#L11

In a different step than build, so it's more granular I'd say ?

package.json
zlajo: You are right! I've just updated the call to mocha.

zlajo:
> One comment.
> Other than this, we need to add this to the CI:
> 
> https://github.com/nextcloud/contacts/blob/4bd46d656fb9e21becc23b069180c09fa3dcb112/.github/workflows/node.yml#L11
> 
> In a different step than build, so it's more granular I'd say ?

I would also use a different integration step. The tricky part will probably be to start up a clean nextcloud instance. Do you have a template to do that?

skjnldsv:
Hey @zlajo, sorry I was out in vacations for quite some time!
I'm back now, I'll take another look here :)

zlajo:
@skjnldsv Sorry, I've been busy with my job and my family, but I just did the rebase on the current master.

skjnldsv:
No need to be sorry.

So, I gave some thoughts to this. The issue I have is that it's using an entire different testing suite that what we're used to work with at NC.
For regression and acceptance tests, we use cypress (aside from server, but it's the exact same problem)

Implementing selenium and a tech that only you know, will not help us maintain or fix any potential issues in the future.

What do you think @ChristophWurst ?
I'm afraid this will backfire

skjnldsv:
> How about you prepare a skeleton for cypress so that it will be consistent with the rest of your tests

Unfortunately I'm lacking time and ressources to focus on this :(

ChristophWurst:
With https://github.com/nextcloud/nextcloud-cypress it looks like we moved even more onto the cypress ship.

Only twofactor_totp uses Selenium. But that is also for historic reasons and because the app isn't much developed/changed these days.

I would therefore suggest to close this PR. Hope that is okay with you @zlajo.

PR #2040, opened 2021-02-05
Fix display name with empty last name when sorting by last name

NerLOR:
When sorting by last name, contacts without last name result in `, Alice` because last and first name are joined with `', '` regardless of their contents. Easy fix: return only first name if last name is empty.


skjnldsv (approved):
That makes more sense! :)
Thanks for the feedback!

PR #2064, opened 2021-02-17
Fix custom labels not being displayed

### LINKED ISSUE #1842: ITEM1 property not displayed
I saw there have been similar issues (#42 #942) but they're all *closed* now and I'm still experiencing the issue.

### Steps to reproduce
Add a ITEM1 prefix on any line (as done by an iOS14 device for example for addresses and websites)

### Expected behaviour
Property is shown correctly

### Actual behaviour
Properties with an ITEM1 prefix are not shown completely in the contacts app. Data is still present in the underlying vcf card though.

Example:
Item `ADR;type=WORK;type=pref:;;;Rome;RM;;IT` is correctly shown, while `item1.ADR;type=WORK;type=pref:;;;Rome;RM;;IT` isn't completely.

### Server configuration
**Nextcloud version:**
18.0.10

**Contacts version:**
3.4.0

**Updated from an older Nextcloud or fresh install:**
Updated

Y0hy0h:
Should fix #1842.

The `canDisplay` filtering was [introduced almost a year ago](https://github.com/nextcloud/contacts/commit/417448cac5e1ebac691a8fc417a578f45e9bf08e#diff-42f8560182cc1534801c80adee527b8a79c2c55308e110a799c86ccb808cdf39R384). It checks whether the property name is "registered" in [`rfcProps.js`](https://github.com/nextcloud/contacts/blob/master/src/models/rfcProps.js), and if not, it does not pass it along.

The problem is that custom labels are implemented by using two vCard fields with the same, e.g., `group1.X-ABLabel` and `group1.TEL`, where the `.` separates the prefix from the field types. [The sorting function sorts by what is after the `.`](https://github.com/nextcloud/contacts/blob/master/src/components/ContactDetails.vue#L366), since we are not interested in the prefix. By the same logic, we should only check what is after the `.` for whether it is a property we can display.

I don't have the energy to figure out how to test this change. But I'm confident enough that this is the fix that I wanted to propose it in the form of a PR.

### DIFF
--- a/src/components/ContactDetails.vue
+++ b/src/components/ContactDetails.vue
@@ -759,7 +759,9 @@ export default {
 		 * @returns {boolean}
 		 */
 		canDisplay(property) {
-			const propModel = rfcProps.properties[property.name]
+			// Make sure we have some model for the property and check for ITEM.PROP custom label format
+			const propModel = rfcProps.properties[property.name.split('.').pop()]
+
 			const propType = propModel && propModel.force
 				? propModel.force
 				: property.getDefaultType()


Y0hy0h:
Sorry, I had an embarrassing missing paren.

Y0hy0h:
And now DCO is happy.

Y0hy0h:
And now I turned of my auto-formatter. Sorry for the noise.

skjnldsv (changes requested):
Hey, thanks for this!
The fix should go in the canDisplay method.
https://github.com/nextcloud/contacts/blob/5c63289fd7eb7ac5122d1e061ab43a1b89442957/src/components/ContactDetails.vue#L763-L770

Here you drop the normal properties. You need to stay compatible with the non-split properties

skjnldsv (approved):

skjnldsv:
Done, I pushed! @Y0hy0h please double check my changes :)

Thanks a lot for narrowing this!

Y0hy0h:
Yeah, this makes sense! Also great idea to add the comment. 

I'm happy it helped! Thanks for fixing the fix so quickly. 😊

sciurius:
I can confirm that this solved the problem for me. I have many contacts that have ADR entries similar to

    item0.ADR;TYPE=HOME: ...
    item0.X-ABADRLnl

With the fix the address information is displayed again.

Thanks, Johannes!

skjnldsv:
I'll release a new version as soon as possible

PHP-Expert:
Just a short push / comment, that this Topic is still open and can be reproduced when connecting IOS to Nextcloud contacts

ChristophWurst:
@PHP-Expert file a new bug report please

PHP-Expert:
https://github.com/nextcloud/contacts/issues/3645

PR #2066, opened 2021-02-19
Fix communication of enabled/disabled addressbook (#1927)

### LINKED ISSUE #1927: Enabling address books with odd user communication logic
In order to disable an address book, you uncheck **Enabled** in the address book's ... menu. That's fine.

image

However, in order to enable a disabled address book, you need to check **Disabled**, which is illogical.

image

Tested with Contacts 3.4.1 on Nextcloud 20.0.1

harryyoud:
Not super sure how to tackle removing the "Disabled" translation, since as far as I can tell, it isn't used anywhere else

Fix #1927

### DIFF
--- a/src/components/Settings/SettingsAddressbook.vue
+++ b/src/components/Settings/SettingsAddressbook.vue
@@ -71,11 +71,11 @@
 				<ActionCheckbox v-if="!toggleEnabledLoading"
 					:checked="enabled"
 					@change.stop.prevent="toggleAddressbookEnabled">
-					{{ enabled ? t('contacts', 'Enabled') : t('contacts', 'Disabled') }}
+					{{ t('contacts', 'Enabled') }}
 				</ActionCheckbox>
 				<ActionButton v-else
 					icon="icon-loading-small">
-					{{ enabled ? t('contacts', 'Enabled') : t('contacts', 'Disabled') }}
+					{{ t('contacts', 'Enabled') }}
 				</ActionButton>
 			</template>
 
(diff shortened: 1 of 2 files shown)


src/components/Settings/SettingsAddressbook.vue
```diff
@@ -71,11 +71,11 @@
 				<ActionCheckbox v-if="!toggleEnabledLoading"
 					:checked="enabled"
 					@change.stop.prevent="toggleAddressbookEnabled">
-					{{ enabled ? t('contacts', 'Enabled') : t('contacts', 'Disabled') }}
+					{{ t('contacts', 'Enabled') }}
 				</ActionCheckbox>
 				<ActionButton v-else
 					icon="icon-loading-small">
-					{{ enabled ? t('contacts', 'Enabled') : t('contacts', 'Disabled') }}
+					{{ t('contacts', 'Enabled') }}
```
skjnldsv, line 78: Should be Disable/Enable ? No?

harryyoud: I'm not sure what the best UX is, and I'm definitely not a UX person, so I'd appreciate some further thoughts.

If the address book is disabled, I'd personally expect an "Enabled" checkbox to be unticked (`enabled == true`), or a "Disabled" checkbox to be ticked (`disabled == true`).

At present, if it's disabled, there is a "Disabled" checkbox that is unticked, which to me means `disabled == false`.

That's what I decided it would be best sticking with the same wording in both states and rely on the checkbox to convey the current state.

skjnldsv: I think we should move away from the preterit form. Use the action: "Enable/Disable" :)

PR #2077, opened 2021-02-21
Allow sending emails to contact group (#1848)

### LINKED ISSUE #1848: Mailto button for groups
When you have groups in Nextcloud Contacts, you may download an entire group by one click.

What is missing, is a `mailto:` button, that opens the default mailer on the user system.

harryyoud:
Fix #1848

### DIFF
--- a/src/components/AppNavigation/GroupNavigationItem.vue
+++ b/src/components/AppNavigation/GroupNavigationItem.vue
@@ -38,6 +38,11 @@
 				@click="downloadGroup(group)">
 				{{ t('contacts', 'Download') }}
 			</ActionButton>
+			<ActionButton
+				icon="icon-mail"
+				@click="emailGroup(group)">
+				{{ t('contacts', 'Send email') }}
+			</ActionButton>
 		</template>
 
 		<AppNavigationCounter v-if="group.contacts.length > 0" slot="counter">
@@ -126,6 +131,33 @@ export default {
 				download(response.data, filename, 'text/vcard')
 			})
 		},
+
+		/**
+		 * Open mailto: for contacts in a group
+		 *
+		 * @param {Object} group of contacts to be emailed
+		 */
+		emailGroup(group) {
+			const emails = []
+			group.contacts.forEach(key => {
+				// The email property could contain "John Doe <john.doe@example.com>", but vcard spec only
+				// allows addr-spec, not name-addr, so to stay compliant, replace everything outside of <>
+				const email = this.contacts[key].email.replace(/(.*<)([^>]*)(>)/g, '$2').trim()
+				const name = this.contacts[key].fullName.replace(/[,<>]/g, '').trim()
+				if (email === '') {
+					return
+				}
+				if (name === null || name === '') {
+					emails.push(email)
+					return
+				}
+				emails.push(`${name} <${email}>`)
+			})
+			// We could just do mailto:${emails}, but if we want to use name-addr, not addr-spec, then we
+			// have to explicitly set the "To:" header.
+			window.location.href = `mailto:?to=${emails.map(encodeURIComponent).join(',')}`
+		},
+
 	},
 }
 </script>


skjnldsv:
@jancborchardt thoughts?

shadowkrad:
yes I am also interested in this implementation, it is possible to share files to users entered in the personal address book but you have no user groups.

harryyoud:
This seems to work well for me with Thunderbird as my default `mailto:` handler. 

I'm happy to test this with a few other mail clients such as Roundcube, Windows 10 Mail, Outlook, and macOS Mail.App.

I'd appreciate if someone could test any other common mail clients that I haven't thought of.

shadowkrad:
image

mine was not with the mail system but with file sharing, where the users of the user's address book can be selected but the user groups are not

harryyoud:
I've noticed that when you have mailto handled by the browser that NextCloud is open in (e.g. RoundCube as mailto handler), it hijacks the NextCloud tab/window. We could get around this by using _target=blank, but this has an unwanted effect for users using a desktop mail client, creating a new blank tab. See https://stackoverflow.com/q/8549816 for more details.

I think the result we have right now is probably the best one

harryyoud:
> image
> 
> mine was not with the mail system but with file sharing, where the users of the user's address book can be selected but the user groups are not

I'm not sure this is related to this issue/pull request, maybe you should open a new issue?

jancborchardt (approved):
Neat little feature, and works great! :) Thanks @harryyoud

src/views/Contacts.vue
```diff
@@ -662,6 +667,32 @@ export default {
...
+				}
+				if (name === null || name === '') {
+					emails.push(email)
+					return
+				}
+				emails.push(`${name} <${email}>`)
+			})
+			// We could just do mailto:${emails}, but if we want to use name-addr, not addr-spec, then we
+			// have to explicitly set the "To:" header. Plus, we want to BCC the recipients
+			window.location.href = `mailto:?bcc=${emails.map(encodeURIComponent).join(',')}`
```
jancborchardt, line 693: Super tiny enhancement: A space after the comma would be nice here to separate the different recipients better. Doesn’t matter for the actual functionality, just looks a bit nicer while writing. :)

jancborchardt:
@harryyoud btw we have a Nextcloud Talk instance for contributors, where we also have a channel for Contacts / Groupware, for quicker communication. We could add you there if you like? :)

harryyoud:
@jancborchardt yeah, that would be great, thank you

src/views/Contacts.vue
harryyoud: I did try this and I think I found that one of the clients would then add %20 before each of the email addresses. It was a while ago now, so I'll have to double check

jancborchardt: Ahh ok, then nevermind! :)

jancborchardt:
Good to go then from my side @skjnldsv @zlajo @call-me-matt what do you think? :)

skjnldsv (approved):
:elephant:

src/views/Contacts.vue
ChristophWurst, line 693: why bcc and not to?

harryyoud: Email etiquette is usually to BCC group recipients to avoid leaking email addresses to the recipients. I guess if we wanted to offer an option, maybe we could hide it behind a user preference?

ChristophWurst: I'd say this really depends. If I sent to a group of coworkers I would certainly want them to know who else go the email. For something where I send holiday wishes to clients I wouldn't want them to see each other's address.

A user preference doesn't seem right either. This is more a case-by-case decision. One email might go to undisclosed recipients, the other one is for direct recipients.

In the end most email client will allow you to move people from *to* to *bcc*. I'd vote for *to* by default and let people move recipients around if they wish.

jancborchardt: > I'd vote for to by default and let people move recipients around if they wish.

Would say that as well from a user perspective. The question is all about the default, and the less confusing one is if it goes in "to". :)

skjnldsv:
Needs rebase :)
Lots of changes happened, sorry about that

harryyoud:
Rebased, and changed to To: rather than BCC

ChristophWurst (approved):

PR #2085, opened 2021-02-22
use all email addresses for gravatar search

### LINKED ISSUE #2084: gravatar profile is only taken from first mail entry
if there are multiple email addresses, the gravatar is only searched for the first entry. In case the first address does not have a gravatar, no picture is downloaded - even if the second address has one.

call-me-matt:
fixes #2084

### DIFF
--- a/lib/Service/SocialApiService.php
+++ b/lib/Service/SocialApiService.php
@@ -216,7 +216,6 @@ class SocialApiService {
 						break;
 					}
 				} catch (\Exception $e) {
-					return $e;
 				}
 			}
 


skjnldsv (approved):

Mannshoch:
I found this in the Changelog. Is it possible to deactivate Gravatar search? I do not wish to expose my address book to Gravatar.

call-me-matt:
No worries, this feature is deactivated by default. As admin, you can even disallow your users to make use of it (in the admin settings).
And by the way: even if you chose to fetch avatars from gravatar, you only transmit a hash of the email addresses.

Mannshoch:
Ah, good, thank you for the hint. 

Sorry but I do nod allow using Gravatar. Even it is only a Hash. Gravatar knows where this mail is used, so it has a network of website you visited and may could compare it with other. 
If I understand it right, in the Contact app, you upload a hash of all contacts in several address book you have. In this case you allow Gravatar to track mail addresses that not itself opt-in getting tracked by Gravatar. If Gravatar in any case is able to detect the mail address behind the hash. e.g. by intelligent brute force >>{BRUTEFORCE}@{COMMON-MAIL-DOMAIN<< your whole Internet live is known by them.
I'm may be able to decide if I wish be tracked for my mail address (I don't wish) but I'm not able to do it for people in my address book.

PR #2099, opened 2021-03-04
Fix codecov token

skjnldsv:


ChristophWurst (approved):

skjnldsv:
Doesn't seems to work though, I need to find the time to look at this

Issue #2245, opened 2021-06-03
Explanatory Circles text also directly in nav as empty content


jancborchardt:
Right now we show this helpful Circles explanatory text, but only as soon as you add a new circle:
![circles](https://user-images.githubusercontent.com/925062/120658789-085e6d80-c486-11eb-9ac7-2ab23e48c9c1.png)

We could show that first paragraph also below the "Circles" heading in the navigation. If it’s an issue with space, we could only show it when there are no Circles yet.

skjnldsv:
![dev skjnldsv com_apps_contacts_All%20contacts_466823ea-ff34-441e-b274-e42e92c44bb6_contacts](https://user-images.githubusercontent.com/14975046/120755040-be20cf00-c50d-11eb-9198-6b6663513903.png)

PR #2250, opened 2021-06-04
Show explanatory description on empty circles list

skjnldsv:
Fix https://github.com/nextcloud/contacts/issues/2245

image


skjnldsv:
Maybe the icon and title is too much, a simple `<p>` would suffice?

PR #2265, opened 2021-06-07
Fix new circle disabled button

skjnldsv:


mejo- (approved):
Perfect. Just tested and indeed it fixes the described problem!

PR #2283, opened 2021-06-14
Fix renaming a circle

mejo-:
So far, attempts to rename a circle resulted in the circle description
being changed instead.

Also don't set `loadingDescription` to true while updating the circle
name.

### DIFF
--- a/src/components/CircleDetails.vue
+++ b/src/components/CircleDetails.vue
@@ -34,17 +34,19 @@
 			</template>
 
 			<!-- display name -->
-			<input
-				slot="title"
-				v-model="circle.displayName"
-				:readonly="!circle.isOwner"
-				:placeholder="t('contacts', 'Circle name')"
-				type="text"
-				autocomplete="off"
-				autocorrect="off"
-				spellcheck="false"
-				name="displayname"
-				@input="onNameChangeDebounce">
+			<template #title>
+				<input
+					v-model="circle.displayName"
+					:readonly="!circle.isOwner"
+					:placeholder="t('contacts', 'Circle name')"
+					type="text"
+					autocomplete="off"
+					autocorrect="off"
+					spellcheck="false"
+					name="displayname"
+					@input="onNameChangeDebounce">
+				<div v-if="loadingName" class="circle-name__loader icon-loading-small" />
+			</template>
 
 			<!-- org, title -->
 			<template v-if="!circle.isOwner" #subtitle>
@@ -142,6 +144,7 @@ export default {
 	data() {
 		return {
 			loadingDescription: false,
+			loadingName: false,
 		}
 	},
 
@@ -196,11 +199,14 @@ export default {
 			this.onNameChange(event.target.value)
 		}, 500),
 		async onNameChange(name) {
+			this.loadingName = true
 			try {
 				await editCircle(this.circle.id, CircleEdit.Name, name)
 			} catch (error) {
 				console.error('Unable to edit circle name', name, error)
 				showError(t('contacts', 'An error happened during name sync'))
+			} finally {
+				this.loadingName = false
 			}
 		},
 	},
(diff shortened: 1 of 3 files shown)


src/components/CircleDetails.vue
```diff
@@ -192,18 +192,15 @@ export default {
 			}
 		},
 
-		onDisplayNameChangeDebounce: debounce(function(event) {
-			this.onDisplayNameChange(event.target.value)
+		onNameChangeDebounce: debounce(function(event) {
+			this.onNameChange(event.target.value)
 		}, 500),
-		async onDisplayNameChange(description) {
-			this.loadingDescription = true
```
skjnldsv, line 199: Right, probably a copy/paste leftover :)
Would you mind adding a loading indicator next to the title then?

src/services/circles.d.ts
```diff
@@ -24,9 +24,9 @@ interface MemberPairs {
     id: string;
     type: MemberType;
 }
-declare type CircleEditType = 'displayName' | 'description' | 'settings' | 'config';
+declare type CircleEditType = 'name' | 'description' | 'settings' | 'config';
 export declare enum CircleEdit {
-    DisplayName = "displayName",
+    Name = "displayName",
```
skjnldsv, line 29:
```suggestion
    Name = "name",
```

src/services/circles.ts
```diff
@@ -28,9 +28,9 @@ interface MemberPairs {
 	type: MemberType
 }
 
-type CircleEditType = 'displayName' | 'description' | 'settings' | 'config'
+type CircleEditType = 'name' | 'description' | 'settings' | 'config'
 export enum CircleEdit {
-	DisplayName = 'displayName',
+	Name = 'displayName',
```
skjnldsv, line 33:
```suggestion
	Name = 'name',
```

src/services/circles.d.ts
mejo-: That would require the circles route to change first. But as @daita already planned to do it, let's do it now ;)

I just opened https://github.com/nextcloud/circles/pull/619

src/components/CircleDetails.vue
mejo-: Sure, I could do that. But that would have to go to component `DetailsHeader`, as that's the one used to display the circle name field. You don't have a loading indicator there yet, not for contacts title either.

Shall I add a generic loading indicator to `DetailsHeader` and make `CircleDetails` use that one? If so, where would you place it? Left to the title/name input field looks very odd on big screens as the input field takes the full width per default. Another option would be to display it instead of (or as overlay over) the avatar.

Anyway, probably that change should go into a separate PR, no?

mejo-: I certainly meant "right to the title/name input field looks very odd" above :wink:

skjnldsv: I would just use the `title` slot as well

mejo-: Ok, that's the least invasive solution :)

I'm a bit unsure where to place the loading spinner design-wise. Just show it instead of the name field? Or in an overlay over the name field and place it there on the left?

skjnldsv: show it after the input?

mejo-:
>show it after the input?

As written earlier, the problem here is that the input field takes the whole width, which results in the loading indicator being placed at the very right end of your screen. That looks odd on large screens.

mejo-:
> As written earlier, the problem here is that the input field takes the whole width, which results in the loading indicator being placed at the very right end of your screen. That looks odd on large screens.

Here's a screenshot, how it would look like on my screen:

2021-06-15T10:45:56,723495815+02:00

src/components/CircleDetails.vue
mejo-: Ok, after playing around a bit and realizing that I don't know a better solution, I ended up adding the loading indicator after the input field now :)

mejo-:
Alright, PR should be ready to merge now :slightly_smiling_face:

mejo-:
I just rebased on latest master.

PR #2295, opened 2021-06-21
Support population count

skjnldsv:
ref https://github.com/nextcloud/circles/pull/633/


ArtificialOwl:
not looking at the code, but at the result, and it looks good

image

ArtificialOwl (approved):

PR #2298, opened 2021-06-22
Support contacts member type

skjnldsv:
To reproduce:
1. Add a member type contacts from occ
   `occ circles:mem:add ADgqG5w9ixrAUh3 'admin/contacts/9788480a-d95f-45b6-bd8d-b0989563938c'`
   format is `USER/ADDRESSBOOK/UID`
2. Open the circle on contacts
3. See a red error

This fixes the display, but we still cannot add contacts through the UI for now.
Needs to be addressed later (either add a new `SHARE_TYPE_CONTACT`, or find something else)

### DIFF
--- a/src/components/MemberList.vue
+++ b/src/components/MemberList.vue
@@ -151,11 +151,15 @@ export default {
 				.map(type => parseInt(type, 10))
 				// Map populated types to the group entry
 				.map(type => CIRCLES_MEMBER_GROUPING.find(group => group.type === type))
+				// Removed undefined group
+				.filter(group => group !== undefined)
 				// Injecting headings
-				.map(group => [{
-					heading: true,
-					...group,
-				}, ...(this.groupedList[group.type] || [])])
+				.map(group => {
+					return [{
+						heading: true,
+						...group,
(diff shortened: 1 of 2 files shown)


src/components/MemberList.vue
```diff
@@ -151,11 +151,16 @@ export default {
...
 				.map(type => CIRCLES_MEMBER_GROUPING.find(group => group.type === type))
+				// Removed undefined group
+				.filter(group => group !== undefined)
 				// Injecting headings
-				.map(group => [{
-					heading: true,
-					...group,
-				}, ...(this.groupedList[group.type] || [])])
+				.map(group => {
+					console.info(group)
```
ChristophWurst, line 158: leftover or intended to stay in production code?

ArtificialOwl:
Result is fine for me, haven't check the code

image

ArtificialOwl (approved):

src/components/MemberList.vue
skjnldsv: :roll_eyes: Thank you!

Issue #2362, opened 2021-07-20
`npm ci` fails due to wrong cdav-library tarball checksum.


mejo-:
Installing NodeJS dependencies for the contacts app currently fails both locally on my machine and in Gitlab CI due to a integrity checksum missmatch for nextcloud/cdav-library:

```
> git show -q
commit b8dc17def6e82c6e59e28d00ff27980050c49057 (HEAD -> master, origin/master, origin/HEAD)
[...]
> npm -v
7.6.3
> npm ci
[...]
npm WARN tarball tarball data for cdav-library@git+ssh://git@github.com/nextcloud/cdav-library.git#0d826181fdb3958ad7ec214fc00b450edb55a866 (sha512-pa6beP0vY/exZKPOhEpqHTnkVLtksjZaPKFTmhCHvhieDxrNbpqCC2EHHZHy5x1RPFxMbpqeipTLK80m3A+MZw==) seems to be corrupted. Trying again.
npm WARN tarball tarball data for cdav-library@git+ssh://git@github.com/nextcloud/cdav-library.git#0d826181fdb3958ad7ec214fc00b450edb55a866 (sha512-pa6beP0vY/exZKPOhEpqHTnkVLtksjZaPKFTmhCHvhieDxrNbpqCC2EHHZHy5x1RPFxMbpqeipTLK80m3A+MZw==) seems to be corrupted. Trying again.
npm ERR! code EINTEGRITY
npm ERR! sha512-pa6beP0vY/exZKPOhEpqHTnkVLtksjZaPKFTmhCHvhieDxrNbpqCC2EHHZHy5x1RPFxMbpqeipTLK80m3A+MZw== integrity checksum failed when using sha512: wanted sha512-pa6beP0vY/exZKPOhEpqHTnkVLtksjZaPKFTmhCHvhieDxrNbpqCC2EHHZHy5x1RPFxMbpqeipTLK80m3A+MZw== but got sha512-Ndwy4++X1m1jIqoQFP882SQH+RfkDTLbYRXnePavNNlLx5vmR4sHLRWZMIlwphWtxiU+mKhGisPM10V6K6ISXA==. (262717 bytes)

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/user/.npm/_logs/2021-07-20T08_57_59_563Z-debug.log
```

`package-lock.json` references `cdav-library` as follows:

```
               "node_modules/cdav-library": {
                        "version": "0.0.1",
[...]

PR #2363, opened 2021-07-20
Fix sha512 checksum for NPM dependency cdav-library (Fixes: #2362)

mejo-:
The sha512 checksum for cdav-library commit 0d82618 in
`package-lock.json` was wrong.


mejo-:
I was able to track down the issue to the NPM version being installed: with NPM 7.7.0 or newer installed, the existing integrity checksum is calculated, with NPM 7.6.4 or older, the other one. I suspect the change regarding npm publish in NPM 7.7.0 to be the reason for this.

In other words, there's no problem with the package-lock.json as it is right now, but it requires NPM >= 7.7.0. Sorry for the noise.

I's always so much fun (not!) to debug the toolchain only to realize that something backwards-incompatible changed "somewhere" :cry:

Issue #1703, opened 2020-07-15
Different "views" in the left pane


FBachofner:
It would be great if Contacts would have an easier way to show what is in a particular address book (or, possibly, group of books).

Right now, if a user has multiple address books, the _only_ apparent way to show what is in a single (or multiple) address book(s) is to turn off (i.e. uncheck "enabled") ALL the "unwanted" books.

It would be fantastic if "views" for each address book (similar to the "all contacts" view) could be placed in the left pane.

Clicking on such a "view"(or multi-selecting from multiple books somehow) would then show only contacts from selected books in the middle pane/column (i.e. the contacts pane)

skjnldsv:
SO listing the addressbooks as clickable entries out of the settings menu and on the navigation?
@jancborchardt what do you think? I always thought the addressbooks were too hidden in the settings. We could list them as a collapsible entry just above the settings, like we do with trashbin for example?

![Capture d’écran_2020-07-15_08-05-34](https://user-images.githubusercontent.com/14975046/87509508-fc9d6900-c671-11ea-86d5-04d3590cf46e.png)

FBachofner:
@skjnldsv this is an interesting, slightly different approach to what I was thinking.

I think it could work well and _might also more clearly expose to users how to add address books_.  Look for example at recent issue [1635](https://github.com/nextcloud/contacts/issues/1635) to see how various people have trouble with this.

Thanks for your consideration!

jancborchardt:
So, the default case and most often used for regular people will be 1 address bool, that's it. Other products don't even really have the notions of different address books. It's different groups and that's that.

However, we can improve the design for people who do worl with multiple address books. It shouldn't mess with the default experience though. So how about:
- The normal experience stays like it is
- If there are 2+ address books, the display is more like in the Mail app, with subheadings for the address books (like the accounts in Mail) and a unified list too.

mburnicki:
Hi, I'd just like to leave a note here:

Currently I have about 2800 contacts in 45 address books, private ones, business ones, shared with specific groups of other people, etc.

There are about 100 contact groups that share contacts from different address books. I consider groups as a second way to sort contacts.

For example if you have a huge number of items or objects you could group them either by forms (round or cornered) or by colors (red, green, ...) and of course the resulting groups would intersect with each other. In a similar way I use address books and contact groups.

[...]

PR #2372, opened 2021-07-22
Streamline Settings

szaimen:
This PR is best reviewed like this: https://github.com/nextcloud/contacts/pull/2372/files?diff=unified&w=1

| Before | After |
|---|---|
| image | image |

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/css/SettingsSection.scss
+++ b/css/SettingsSection.scss
@@ -22,26 +22,62 @@
  */
 
 // contacts settings section
+.app-contacts #app-settings-content {
+    padding-left: 0px !important;
+}
+
 #app-settings-content {
 	.settings-section {
 		display: flex;
 		align-items: center;
-		margin-top: 5px;
 	}
+}
+
+.new-addressbook-entry {
+	display: flex;
+	align-items: center;
 }
 
 #new-addressbook-form {
 	display: flex;
+	width: calc(100% - 44px);
 	input {
 		margin-top: 0;
 		margin-bottom: 0;
 	}
 }
 
+.settings-addressbook-list {
+	display: flex;
+	li {
+		width: calc(100% - 44px);
+	}
+	&__header {
+		padding-left: 44px;
+		color: var(--color-primary-element);
+	}
+}
+
+.settings-line__icon {
+	width: 44px;
+	height: 44px;
+}
+
+.social-sync__list-entry {
+	label::before {
+		margin-left: 14px !important;
+		margin-right: 14px !important;
+	}
+	em {
+		margin-left: 44px !important;
+	}
+}
+
 // Sort Contacts
 .sort-contacts {
+	display: flex;
 	.multiselect {
-		width: 100%;
+		width: calc(100% - 44px);
 		margin: 0;
 		.multiselect__single {
 			padding-right: 24px !important;
@@ -49,6 +85,9 @@
 			background-position: right 4px center;
 
 			@include icon-color('triangle-s', 'actions', $color-black, 1, true);
+		}
+		&__content-wrapper {
+			bottom: unset !important;
 		}
 	}
 }
@@ -73,6 +112,10 @@
 		width: 100%;
 		margin: 0;
 		margin-top: -1px;
+	}
+	&__button-main {
+		margin-left: 10px;
+		width: calc(100% - 10px) !important;
 	}
 }
 
--- a/src/components/AppNavigation/SettingsSection.vue
+++ b/src/components/AppNavigation/SettingsSection.vue
@@ -23,7 +23,8 @@
 
 <template>
 	<div>
-		<div v-if="allowSocialSync">
+		<SettingsSortContacts class="settings-section" />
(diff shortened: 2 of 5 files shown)


szaimen:
Every time when I try to compile the js it fails with 
`npm WARN tarball tarball data for cdav-library@git+ssh://git@github.com/nextcloud/cdav-library.git#e62c7c9942d6759743dc9a0350921e898440f2d5 (sha512-xgZ7MVeYeiXwQl5SuyXRVxXPM2swxG46HS0m0IFQ2SmzOgBTm3AdzUFJicggJ34e3SIzrFniWaftkhr550eyJw==) seems to be corrupted.`

skjnldsv:
Requires @jancborchardt approval.
But there was different plans for the settings: https://github.com/nextcloud/contacts/issues/1703#issuecomment-703083443

szaimen:
> But there was different plans for the settings: [#1703 (comment)](https://github.com/nextcloud/contacts/issues/1703#issuecomment-703083443)

Yes but until it isn't inside a modal which I am not able to do myself, I thought this was a good small fix which improves the readability of the settings by a lot, imho.

src/components/AppNavigation/SettingsSection.vue
```diff
@@ -23,7 +23,7 @@
 
 <template>
 	<div>
-		<div v-if="allowSocialSync">
+		<div v-if="allowSocialSync" class="socialSyncListEntry">
```
artonge, line 26: Should use [BEM naming style](https://cssguidelin.es/#bem-like-naming)

```suggestion
 		<div v-if="allowSocialSync" class="social-sync__list-entry">
```

css/SettingsSection.scss
```diff
@@ -26,22 +26,57 @@
...
+	align-items: center;
+	.icon-add {
+		width: 44px;
+		height: 44px;
 	}
 }
 
 #new-addressbook-form {
 	display: flex;
+	width: calc(100% - 44px);
```
artonge, line 43: Not a fan of specifying width en elements that do not need to be constrained. Can't the browser guess that width from the parent ?

css/SettingsSection.scss
```diff
@@ -26,22 +26,57 @@
...
-		margin-top: 5px;
+	}
+}
+
+.new-addressbook-entry {
+	display: flex;
+	align-items: center;
+	.icon-add {
+		width: 44px;
+		height: 44px;
```
artonge, line 37: Could be grouped with other icon styles in a shared class like `settings-line__icon` or something.

artonge (changes requested):

src/components/AppNavigation/SettingsSection.vue
szaimen: done

css/SettingsSection.scss
szaimen: done

jancborchardt:
Hmm, so since these icons would not be there when moving the settings to the modal, I would say we don’t need to add them here. (And also e.g. the icon is better inside the "Import contacts" button than outside it. :)

szaimen:
> Hmm, so since these icons would not be there when moving the settings to the modal, I would say we don’t need to add them here.

The question is how long it will take to bring the settings into the Modal since the current settings are really bad from a        usability perspective, imo. So I'd rather merge this and revert it when bringing it into the settings modall than living with it for another release 😅

css/SettingsSection.scss
szaimen: that doesn't work unfortunately.

szaimen:
> (And also e.g. the icon is better inside the "Import contacts" button than outside it. :)

I've reverted that change, see updated screenshot.

nimisha-vijay:
If we are moving to a modal soon then these changes will not be in effect for too long, but I do think the icons look good. Couple of suggestions:
- move the content of the settings menu slightly to the left so that the checkbox and icons in the content are left aligned with the settings icon in the header, and all the text of the header and content is left aligned as well
- "Enter new address book name" can be changed to something shorter like "Add new address book"?

szaimen:
@nimisha-vijay thanks for the feedback! :)
Addressed both points, see updated screenshots

artonge (approved):
OK, code wise

jancborchardt:
Ok – then I just have 2 more pieces of feedback :)
- The sorting setting would be better as the first item in the settings, as it’s the most basic one.
- The address books could do with a header "Address books" to actually understand what they are. (This heading should also have some whitespace above so that section is a bit separated from the two simpler settings above.)

szaimen:
Done. Please have another look at the updated screenshots @jancborchardt and @nimisha-vijay :)

jancborchardt (approved):

szaimen:
@artonge do you mind reviewing the code another time? I made some small changes...

Pytal (approved):

Issue #2338, opened 2021-07-07
blank screen on Contacts app since last update (4.0)


olivier6931:
Hello guys,

To begin, thanks to all for your work!

So... my issue :
Since the last upgrade of the app (to 4.0.0), I got a blank screen on Contacts page on my Nextcloud instance.
Contact sync with davX seems to work fine, but the frontend stay blank, as follow :

![image](https://user-images.githubusercontent.com/67951376/124732674-b5a82380-df13-11eb-8595-98e0bd329fbe.png)

Thanks for your help !

### Server configuration

**Operating system**: Linux Debian 10 + docker (CE) 20.10.7

**Web server:** Apache (based on image "nextcloud:latest") + reverse proxy Nginx

**Database:** MariaDB (based on image "mariadb:latest")

**PHP version:** PHP 7.4.2 (based on image "nextcloud:latest")

**Nextcloud version:** 21.0.3 (based on image "nextcloud:latest")

**Contacts version:**  4.0.0

**Updated from an older Nextcloud or fresh install:** updated (with docker compose + local volumes)

**Signing status:** No errors have been found.

**List of activated apps:**
  - accessibility: 1.7.0
  - bookmarks: 4.2.2
  - bruteforcesettings: 2.2.0
  - calendar: 2.3.0
  - cloud_federation_api: 1.4.0
  - contacts: 4.0.0
  - cookbook: 0.8.4
  - dashboard: 7.1.0
  - dav: 1.17.1
  - deck: 1.4.2
  - federatedfilesharing: 1.11.0
  - files: 1.16.0
  - files_external: 1.12.0
  - files_rightclick: 1.0.0
  - files_trashbin: 1.11.0
  - lookup_server_connector: 1.9.0
  - notes: 4.1.0
  - notifications: 2.9.0
  - oauth2: 1.9.0
  - passwords: 2021.7.21
  - provisioning_api: 1.11.0
  - settings: 1.3.0
  - theming: 1.12.0
  - twofactor_backupcodes: 1.10.0
  - updatenotification: 1.11.0
  - viewer: 1.5.0
  - weather_status: 1.1.0
  - workflowengine: 2.3.0

### Client configuration

**Browser:** Firefox 89 (desktop & mobile) or edge 91
**Operating system:** Windows 10 + Android 11

#### *Web server error log*
no special errors

#### *data/nextcloud.log*
no special errors

#### Browser log
# Firefox
Uncaught TypeError: OC.Share is undefined
[...]

PR #2400, opened 2021-08-18
Use sharing constants from a package, not another app

ChristophWurst:
This makes the app more :sparkles: independent :sparkles: 

Fixes https://github.com/nextcloud/contacts/issues/2338


skjnldsv (approved):
Ouh, nice!!! :rocket:

PR #2407, opened 2021-08-24
Run SVGs through some sanitization

ChristophWurst:
Better safe than sorry. And if they don't pass then we just display no photo. Good enough :)

### DIFF
--- a/src/components/ContactDetails/ContactDetailsAvatar.vue
+++ b/src/components/ContactDetails/ContactDetailsAvatar.vue
@@ -32,8 +32,8 @@
 			@change="processFile">
 
 		<!-- Avatar display -->
-		<div v-if="contact.photo"
-			:style="{ 'backgroundImage': `url(${contact.photoUrl})` }"
+		<div v-if="photoUrl"
+			:style="{ 'backgroundImage': `url(${photoUrl})` }"
 			class="contact-header-avatar__photo"
 			@click="toggleModal" />
 		<Avatar v-else
@@ -136,7 +136,7 @@
 			<div class="contact-header-modal__photo-wrapper"
 				@click.exact.self="toggleModal">
 				<img ref="img"
-					:src="contact.photoUrl"
+					:src="photoUrl"
 					class="contact-header-modal__photo">
 			</div>
 		</Modal>
@@ -178,11 +178,18 @@ export default {
 		},
 	},
 
+	watch: {
+		contact() {
+			this.loadPhotoUrl()
+		},
+	},
+
 	data() {
 		return {
 			maximizeAvatar: false,
 			opened: false,
 			loading: false,
+			photoUrl: false,
 			root: generateRemoteUrl(`dav/files/${getCurrentUser().uid}`),
 		}
 	},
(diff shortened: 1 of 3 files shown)


src/components/ContactDetails/ContactDetailsAvatar.vue
```diff
@@ -354,6 +359,15 @@ export default {
...
 		},
 
+		async loadPhotoUrl() {
+			const photoUrl = await this.contact.getPhotoUrl()
+			if (!photoUrl) {
+				console.warn('contact has an invalid photo')
+				return
+			}
+			this.photoUrl = photoUrl
+		},
```
skjnldsv, line 369: Why not a computed?

src/components/ContactDetails/ContactDetailsAvatar.vue
```diff
@@ -210,6 +211,10 @@ export default {
 		},
 	},
 
+	mounted() {
+		this.loadPhotoUrl()
```
skjnldsv, line 215: will this catch a contact change if the component stays mounted?

src/models/contact.js
```diff
@@ -229,11 +230,13 @@ export default class Contact {
 	 * Return the photo usable url
 	 * We cannot fetch external url because of csp policies
 	 *
-	 * @readonly
 	 * @memberof Contact
 	 */
-	get photoUrl() {
+	async getPhotoUrl() {
```
skjnldsv, line 235: There are ways to have an async getter :)

src/components/ContactDetails/ContactDetailsAvatar.vue
ChristophWurst: Because it's async

src/models/contact.js
ChristophWurst: Yes there are but it feels like that's not what getters are supposed to do. I would rather have this operation in a traditional method

src/components/ContactDetails/ContactDetailsAvatar.vue
ChristophWurst: Good point. Not yet. I suppose this needs a watcher

LukasReschke (approved):
image

src/models/contact.js
skjnldsv: Hum, that's also a fair point.
That would solve the contact change and avoid weird watchers for the `ContactDetailsAvatar.vue`

ChristophWurst: The watcher is weird indeed. But even if I have an async getter then I can't use a computed props because Vue computed props are not async. One needs a custom plugin for that and I'm not sure if that is not creating more problems. After all template rendering is done sync. So there also needs to be some kind of default value for an async prop before it got evaluated. It's … weird.

skjnldsv: What about something like that?

```js
  get photoUrl() {
    return (async () => {
      try {
        ...
        const cleanSvg = await sanitizeSVG(imageSvg)
        ...
      } catch(e) {
        return 0;
      }
    })();
```

ChristophWurst: :ping_pong: it still doesn't make sense to use async code inside sync code. you won't be able to evaluate the result of the promise

ChristophWurst: Talked with @st3iny about this. The code above won't work. We need the async method, an initial value and a watcher for changes.

kesselb (approved):

PR #2417, opened 2021-08-29
 Fixed search / filter at add contacts dialog is case sensitive issue #1939

### LINKED ISSUE #1939: search / filter at "add contacts" dialog is case sensitive
### Describe the bug

As a user I want to find an existing contact in order to add it to a contact group. Currently the search is case sensitive. But IMHO it should not be. 

**To Reproduce**
1. Go to '/apps/contacts/'
2. Click on '...'-Symbol of a contact group
3. Click on 'Add contacts'
4. Search for a name of your choice in lower case only
5. See error: you do not see any result

**Expected behavior**
The search / filter should return results regardless of the upper or lower case or mixed case of the search term.

**Actual behavior**
The search / filter will only return results if the search term matches exactly.

**Screenshots**
That works: 

Screenshot_Contacts_Stef_dsgvo

That does not: 

Screenshot_Contacts_stef

### Server configuration
**Operating system**: Linux 5.4.0-45-generic x86_64

**Web server:** Apache, Nginx...

**Database:** MariaDB, MySQL, PostgreSQL... 

**PHP version:** 7.0, 7.1, 7.2...

**Nextcloud version:** Nextcloud 19.0.4

**Contacts version:** 3.4.1

**Updated from an older Nextcloud or fresh install:** fresh

**Signing status:** 
```
No errors have been found.
```

**List of activated apps:**

**Nextcloud configuration:**

### Client configuration
**Browser:** Firefox 83.0

**Operating system:**

**CardDAV-clients:**

#### Browser log

```
jQuery is deprecated: The global jQuery is deprecated. It will be updated to v2.4 in Nextcloud 20 and v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own. 80 globals.js:61:15
Content Security Policy: Die Einstellungen der Seite haben das Laden einer Ressource auf inline blockiert ("script-src"). utils.js:35:9
jQuery is deprecated: The global jQuery is deprecated. It will be updated to v2.4 in Nextcloud 20 and v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own. 108 globals.js:61:15
```

Zishan-7:
Fix #1939

Issue was that the search was case sensitive. I fixed it by converting the search query to lower case and also converting the data set to lower case so that the case sensitivity can be removed.

### DIFF
--- a/src/components/EntityPicker/EntityPicker.vue
+++ b/src/components/EntityPicker/EntityPicker.vue
@@ -254,7 +254,7 @@ export default {
 			// If internal search is enabled and we have a search query, filter data set
 			if (this.internalSearch && this.searchQuery && this.searchQuery.trim !== '') {
 				return this.dataSet.filter(entity => {
-					return entity.label.indexOf(this.searchQuery) > -1
+					return entity.label.toLowerCase().indexOf(this.searchQuery.toLowerCase()) !== -1 
 				})
 			}
 			return this.dataSet
--- a/src/components/EntityPicker/EntityPicker.vue
+++ b/src/components/EntityPicker/EntityPicker.vue
@@ -254,7 +254,7 @@ export default {
 			// If internal search is enabled and we have a search query, filter data set
 			if (this.internalSearch && this.searchQuery && this.searchQuery.trim !== '') {
 				return this.dataSet.filter(entity => {
-					return entity.label.toLowerCase().indexOf(this.searchQuery.toLowerCase()) !== -1 
+					return entity.label.toLowerCase().indexOf(this.searchQuery.toLowerCase()) > -1
 				})
 			}
 			return this.dataSet


ChristophWurst (approved):
Fine by me :)

skjnldsv (approved):

src/components/EntityPicker/EntityPicker.vue
```diff
@@ -254,7 +254,7 @@ export default {
 			// If internal search is enabled and we have a search query, filter data set
 			if (this.internalSearch && this.searchQuery && this.searchQuery.trim !== '') {
 				return this.dataSet.filter(entity => {
-					return entity.label.indexOf(this.searchQuery) > -1
+					return entity.label.toLowerCase().indexOf(this.searchQuery.toLowerCase()) !== -1
```
skjnldsv, line 257:
```suggestion
					return entity.label.toLowerCase().indexOf(this.searchQuery.toLowerCase()) !== -1
```

skjnldsv (changes requested):

skjnldsv (approved):

PR #2419, opened 2021-08-30
Fix Don't show groups in share placeholder if group sharing is disabled #2242

### LINKED ISSUE #2242: Don't show groups in share placeholder if group sharing is disabled
### Steps to reproduce
1. Check out nextcloud/server#25658
2. Disallow sharing with groups in the sharing settings
3. Open the share section on an address book

### Expected behaviour
No mention of "groups" in the placeholder.

### Actual behaviour
Mention of "groups" in the placeholder.

Zishan-7:
Fix #2242 by checking group sharing enabled or not then updating the value of placeholder based on that value

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -106,7 +106,10 @@ class PageController extends Controller {
 		$isCirclesEnabled = $this->appManager->isEnabledForUser('circles') === true;
 		// if circles is not installed, we use 0.0.0
 		$isCircleVersionCompatible = $this->compareVersion->isCompatible($circleVersion ? $circleVersion : '0.0.0', 22);
+		// Check whether group sharing is enabled or not
+		$isGroupSharingEnabled = $this->config->getAppValue('core', 'shareapi_allow_group_sharing', 'yes') === 'yes';
 
+		$this->initialStateService->provideInitialState(Application::APP_ID, 'isGroupSharingEnabled', $isGroupSharingEnabled);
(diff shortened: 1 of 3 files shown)


lib/Controller/PageController.php
```diff
@@ -106,6 +106,8 @@ public function index(): TemplateResponse {
 		$isCirclesEnabled = $this->appManager->isEnabledForUser('circles') === true;
 		// if circles is not installed, we use 0.0.0
 		$isCircleVersionCompatible = $this->compareVersion->isCompatible($circleVersion ? $circleVersion : '0.0.0', 22);
+		// Check whether group sharing is enabled or not
+		$isGroupSharingEnabled = $this->config->getAppValue('core', 'shareapi_allow_group_sharing', 'yes') === 'yes';
```
skjnldsv, line 110: You also need to provide it through the initial state

```php
$this->initialStateService->provideInitialState(Application::APP_ID, 'isGroupSharingEnabled', $isGroupSharingEnabled);
```

Zishan-7: Thanks for the suggestion, I have added it.

tcitworld (commented):
LGTM. Can you [squash](https://levelup.gitconnected.com/how-to-squash-git-commits-9a095c1bc1fc) your commits?

ChristophWurst (approved):
Looks good!

hamza221:
Fixed lint check, and rebased.

PR #2501, opened 2021-11-07
Allow special chars for xing profile image

danopz:
Hi, I just ran into the issue that the social provider for Xing wasn't able to load the profile image of a contact having some german umlauts (äöü) in their name.
After correctly url-encoding them they contain a percentage sign (%) so I just added that to the regex to enable certain special chars.
I also added a test case with some special chars.

### DIFF
--- a/lib/Service/Social/XingProvider.php
+++ b/lib/Service/Social/XingProvider.php
@@ -85,7 +85,7 @@ class XingProvider implements ISocialProvider {
 			$result = $this->httpClient->get($profileUrl);
 			$htmlResult = $result->getBody();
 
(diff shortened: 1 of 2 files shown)


call-me-matt (approved):
so simple and so effective - thank you!

skjnldsv (approved):

ChristophWurst (approved):

ChristophWurst:
Unit tests fail. Please see the action output :)

danopz:
Should be fixed, sad scoping in PHP :smile:

ChristophWurst (approved):

ChristophWurst:
CI still says no :zipper_mouth_face:

danopz:
Next try
Sry, don't have a full Nextcloud dev setup so I'm not able to run tests locally

Issue #2495, opened 2021-11-05
[Settings] When you share a contact to some other contact, the name its cut short


GretaD:
If the name has more than 5 characters the name is ellipse. Also in different languages, if the can edit is a log phrase, the name is ellipsed in 2 characters.

![contacts](https://user-images.githubusercontent.com/12728974/140521046-0a65dcb5-17dc-420f-9667-953035421c0d.png)

PR #2503, opened 2021-11-10
Fix the width for the sharee name label

### LINKED ISSUE #2473: Display errors / names are missing
### Describe the bug
If contacts are shared with several users individually, the names are no longer displayed. So you don't know with whom something is shared or who has which permissions.

**To Reproduce**
Share contacts with multiple users

**Expected behavior**
Names are still legible

**Actual behavior**
Names can no longer be read

**Screenshots**
image

GretaD:
This PR fixes the width and adds a popover with the full sharee name.
before
shareebefore

after
shareee
fixes #2473  #2495

### DIFF
--- a/src/components/AppNavigation/Settings/SettingsAddressbookShare.vue
+++ b/src/components/AppNavigation/Settings/SettingsAddressbookShare.vue
@@ -140,3 +140,9 @@ export default {
 	},
 }
 </script>
+<style lang="scss" scoped>
+.addressbook-shares__list {
+	margin-left: -8px;
+	width: 282px;
+}
+</style>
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

GretaD:
i found an edge case, will fix asap

st3iny (approved):
Tested and works. Some linter issues still need to be fixed.

Mululu:
Hi there,

so the problem still persists for me

GretaD:
> Hi there,
> 
> so the problem still persists for me

can you please open an issue for this with a screenshot how it looks on your end?

PR #2552, opened 2021-12-09
Remove NC20, limit php versions

miaulalala:


ChristophWurst (approved):

ChristophWurst:
Once CI runs the expected jobs please adjust https://github.com/nextcloud/contacts/settings/branch_protection_rules/424289 accordingly to drop the old Nextcloud 20 jobs and require the newly added ones

appinfo/info.xml
```diff
@@ -39,7 +39,8 @@
     </types>
 
 	<dependencies>
-		<nextcloud min-version="20" max-version="24" />
+		<php min-version="7.3" max-version="8.0" />
```
skjnldsv, line 42: is this something we should do globally? Like contacts doesn't really have a specific need to define it's own php version rather than those server provides?

I have no expertise on the matter, genuinely wondering if we should apply this globally to our apps :)

ChristophWurst: We recently screwed up due to the lack of this. It's not a must-have. It just makes the php support range more explicit.

miaulalala:
> Once CI runs the expected jobs please adjust https://github.com/nextcloud/contacts/settings/branch_protection_rules/424289 accordingly to drop the old Nextcloud 20 jobs and require the newly added ones

I can't access that page

PR #2580, opened 2021-12-23
Add option to clone contact even when not read-only

harryyoud:
Example scenario: created a contact for a business acquaintance, filled in detailed addresses etc. Now in contact with someone else at the same company. Much easier to clone a contact with existing addresses and generic contact details than to copy each address field one-by-one.

Previously, you could clone only from a read-only address book to a read-write address book. This patch allows you to clone from a read-write address book to another read-write address book, including the original address book the contact being cloned belongs to.

Potential drawbacks: sends whole contact over the wire, rather than using a server-side dav copy, in order to get around the issue of not being able to copy an item into the collection it already belongs to.


src/store/addressbooks.js
```diff
@@ -553,17 +553,7 @@ const actions = {
 		const vData = contact.vCard.toString().replace(/^UID.+/im, '')
 		const newContact = new Contact(vData, addressbook)
 
-		try {
```
skjnldsv, line 556: You're not syncing the contact to the server anymore? :thinking:

harryyoud: It syncs just fine, once you've edited the contact. I believe it is added to the local store, and once edited, is then synced.

It behaves as if you have created a new contact, but with the fields pre-filled. This behaviour is intended, since a direct clone of the contact, since in most cases you would want to change a piece of information before saving.

skjnldsv: > once you've edited the contact

This is the issue, this is not the initial goal of that UX.

harryyoud: Okay, so it needs to be reworked to sync the contact to the server before updating the context?

ChristophWurst (changes requested):

ChristophWurst:
Hey @harryyoud!

Long time no see. Do you have time to finish the PR and get it into a mergeable state based on the feedback above?

Cheers

GVodyanov:
Was implemented

PR #2590, opened 2021-12-30
Remove all obsolete make targets

ChristophWurst:
Those were not document nor used anywhere. Goodbye :wave: 

Follow-up to https://github.com/nextcloud/contacts/pull/2589


GretaD (approved):
farewell then

PR #2632, opened 2022-02-04
Force disabling single-quote list for type parameter

skjnldsv:
Fix https://github.com/nextcloud/contacts/issues/1107

___
Debug session with @miaulalala 
- https://github.com/mozilla-comm/ical.js/blob/ba8e2522ffd30ffbe65197a96a487689d6e6e9a1/lib/ical/stringify.js#L246
- https://github.com/mozilla-comm/ical.js/blob/ba8e2522ffd30ffbe65197a96a487689d6e6e9a1/lib/ical/stringify.js#L227
- https://github.com/mozilla-comm/ical.js/blob/ba8e2522ffd30ffbe65197a96a487689d6e6e9a1/lib/ical/stringify.js#L132
- https://github.com/nextcloud/contacts/blob/e939943bcd14411ce486339e9d61a6c0fffb3e71/src/store/contacts.js#L356

Impossible to override the `stringify.propertyValue` as we don't want to do that for all types,

Suggestion: override the toString method and use a regex to force-remove quotes*
`/TYPE="([a-zA-Z-,]+)"/gm`

### DIFF
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -566,4 +566,9 @@ export default class Contact {
 		}
 	}
 
+	toStringStripQuotes() {
+		const regexp = /TYPE="([a-zA-Z-,]+)"/gm
+		const card = this.vCard.toString()
+		return card.replaceAll(regexp, 'TYPE=$1')
+	}
 }
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -567,8 +567,9 @@ export default class Contact {
 	}
 
 	toStringStripQuotes() {
-		const regexp = /TYPE="([a-zA-Z-,]+)"/gm
+		const regexp = /TYPE="([a-zA-Z-,]+)"/gmi
 		const card = this.vCard.toString()
-		return card.replaceAll(regexp, 'TYPE=$1')
+		return card.replace(regexp, 'TYPE=$1')
 	}
+
 }
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -14,10 +14,10 @@ on:
       - stable*
 
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
 
-    name: node
+    name: test
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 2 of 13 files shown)


src/models/contact.js
```diff
@@ -28,6 +28,8 @@ import store from '../store'
 import updateDesignSet from '../services/updateDesignSet'
 import sanitizeSVG from '@mattkrick/sanitize-svg'
 
+window.ICAL = ICAL
```
ChristophWurst, line 31: I hope this is just for debugging?

skjnldsv: ahaha of course :wink:

miaulalala:
This doesn't solve cases where the type string looks like this: `TYPE="HOME","VOICE"` but I don't think that this would be RFC compliant in the first place.

skjnldsv:
Maybe some testing? :see_no_evil:

skjnldsv:
Oh wow, we have zero js testing :see_no_evil: 
If you feel like adding a very basic one to test the `src/models/contact.js` file

rexbron:
This is a big quality of life improvement for iOS users, anything that can be done to move it along?

miaulalala:
@skjnldsv I have no ida how to add a testing framework to the app at all. This needs to be done by somebody who has JS dev experience - so, not me.

miaulalala (approved):

moeb98:
Sure, this is fixed? Installed NC 23.0.5 and Contacts 4.1.1... still effect that categories on iOS 15.5 are shown as "HOME or "CELL, etc...

rexbron:
> Sure, this is fixed? Installed NC 23.0.5 and Contacts 4.1.1... still effect that categories on iOS 15.5 are shown as "HOME or "CELL, etc...

What I've noticed is that iOS and the CardDav server will sometimes exchange the old incorrectly quoted label as a custom label. What has worked for me is reapply the mobile lable in the web interface and then immediately trigger a sync on contacts. Very annoying that this bug polluted the database but alas.

moeb98:
> > Sure, this is fixed? Installed NC 23.0.5 and Contacts 4.1.1... still effect that categories on iOS 15.5 are shown as "HOME or "CELL, etc...
> 
> 
> 
> What I've noticed is that iOS and the CardDav server will sometimes exchange the old incorrectly quoted label as a custom label. What has worked for me is reapply the mobile lable in the web interface and then immediately trigger a sync on contacts. Very annoying that this bug polluted the database but alas. 

Hmmm... didn't work on my end (assume you meant the NC web interface... not iCloud, right?). At the same time, this would imply to go through hundreds / thousands of contact labels (depending on your address book size). 

If this is due to a polluted database, isn't there an occ command (as a workaround at least)?

OmarWazzan:
@moeb98 Were you able to figure anything out for this?

moeb98:
Nope, after waiting for any feedback here for a few days, I honestly lost track due to other priorities and rather switched back to the old iCloud sync using NC for calendar only.

PR #2647, opened 2022-02-18
Fix missing strict types for the social service

ChristophWurst:


call-me-matt:
I guess we have to convert the http status to string [here](https://github.com/nextcloud/contacts/blob/f6603af4109d61eeab9ce586ec83e433df73cfd0/tests/unit/Service/SocialApiServiceTest.php#L112) for unit tests to pass with strict types

skjnldsv (approved):

ChristophWurst:
> I guess we have to convert the http status to string [here](https://github.com/nextcloud/contacts/blob/f6603af4109d61eeab9ce586ec83e433df73cfd0/tests/unit/Service/SocialApiServiceTest.php#L112) for unit tests to pass with strict types
> ... or to integer [here](https://github.com/nextcloud/contacts/blob/f6603af4109d61eeab9ce586ec83e433df73cfd0/lib/Service/SocialApiService.php#L292)?

It should be an int. Can you fix this please?

miaulalala (approved):

lib/Service/SocialApiService.php
```diff
@@ -408,9 +411,9 @@ public function updateAddressbooks(string $userId, string $offsetBook = null, st
 
 				try {
 					$r = $this->updateContact($addressBook->getURI(), $contact['UID'], $network);
-					$response = $this->registerUpdateResult($response, $contact['FN'], $r->getStatus());
+					$response = $this->registerUpdateResult($response, $contact['FN'], int($r->getStatus()));
```
ChristophWurst, line 414:
```suggestion
					$response = $this->registerUpdateResult($response, $contact['FN'], (int) $r->getStatus());
```

call-me-matt: ah yes, sorry I used your pipeline for testing. Didn't have a dev setup.

call-me-matt (approved):

aloxe (approved):

PR #2653, opened 2022-02-21
Drop php 7.3

skjnldsv:
Not sure it's the best way :thinking: 

Any idea?


ChristophWurst:
https://github.com/nextcloud/contacts/blob/8e3aca924efc74553b926f3ebebdefa37b3ba236/appinfo/info.xml#L42 this means 21 installations on 7.3 would get an incompatible update.

I wouldn't drop 7.2 before we drop Nextcloud 23.

PR #2686, opened 2022-03-23
Fix form id duplication

GretaD:
fix form id duplication and attribution indentation


ChristophWurst (approved):
Easier to review as https://github.com/nextcloud/contacts/pull/2686/files?w=1

Issue #2411, opened 2021-08-27
Sharing address book to groups is not working 


elhananjair:
### Describe the bug
......
I am having a problem with sharing address book to groups, after I share the address book and reload the whole page, the group I shared won't listed and the group doesn't get the share too.
......

**To Reproduce**
1. Go to contacts
2. Click on Settings 
3. On address book click on share option and search for a group you want to share
4. After the share reload the page, the shared group will disappear and the group or user won't get the shared address book too 

**Expected behavior**
The group must have been there for the shared group or user

**Actual behavior**
It wont be shared 

**Screenshots**

https://user-images.githubusercontent.com/69236638/131080910-947b5e1b-d8b3-4b4f-a029-eaeea8c1c173.mp4

### Server configuration

**Operating system**: Windows, Linux, MacOS...
Linux
**Web server:** Apache, Nginx...
Apache
**Database:** MariaDB, MySQL, PostgreSQL... 
PostgreSQL
**PHP version:** 7.0, 7.1, 7.2...
7.4.22
**Nextcloud version:** (see Nextcloud admin page)
21.0.4
**Contacts version:** (see Nextcloud apps page)
4.0.2
**Updated from an older Nextcloud or fresh install:**
older Nextcloud
**Signing status:**
```
Login as admin user into your Nextcloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
```

**List of activated apps:**
```
Enabled:
  - accessibility: 1.7.0
  - activity: 2.14.3
  - admin_audit: 1.11.0
  - analytics: 3.6.0
  - announcementcenter: 5.0.0
  - appointments: 1.10.0
  - apporder: 0.13.0
  - calendar: 2.3.2
  - camerarawpreviews: 0.7.12
  - circles: 0.21.4
  - cloud_federation_api: 1.4.0
  - cms_pico: 1.0.15
  - comments: 1.11.0
  - contacts: 4.0.2
  - contactsinteraction: 1.2.0
  - dashboard: 7.1.0
  - dav: 1.17.1
  - deck: 1.4.3
  - duplicatefinder: 0.0.11
  - event_update_notification: 1.3.0
  - external: 3.8.2
  - extract: 1.3.2
  - federatedfilesharing: 1.11.0
  - federation: 1.11.0
  - files: 1.16.0
  - files_automatedtagging: 1.11.0
  - files_markdown: 2.3.4
  - files_mindmap: 0.0.25
[...]

PR #2708, opened 2022-04-09
Fix: Show display name in group dropdown selection

max65482:
Fixes one aspect of #2411.

The dropdown box to select groups/users for sharing now shows the display names (not identifiers).


max65482:
fixed in #2946

PR #2711, opened 2022-04-10
Add search feature on the list

AminulBD:
Close https://github.com/nextcloud/contacts/issues/2796

It's nice to have the search feature on the contact list.

  
Preview:
image

### DIFF
--- a/src/components/ContactsList.vue
+++ b/src/components/ContactsList.vue
@@ -22,7 +22,11 @@
 
 <template>
 	<AppContentList>
-		<div class="contacts-list__header" />
+		<div class="contacts-list__header">
+			<div class="search-contacts-field">
+				<input type="text" :placeholder="t('contacts', 'Search contacts …')" v-model="query">
+			</div>
+		</div>
 		<VirtualList ref="scroller"
 			class="contacts-list"
 			data-key="key"
@@ -63,7 +67,12 @@ export default {
 	data() {
 		return {
 			ContactsListItem,
+			query: '',
 		}
+	},
+
+	mounted() {
+		this.query = this.searchQuery
 	},
 
 	computed: {
@@ -143,8 +152,8 @@ export default {
 		 * @return {boolean}
 		 */
 		matchSearch(contact) {
-			if (this.searchQuery.trim() !== '') {
-				return contact.searchData.toString().toLowerCase().search(this.searchQuery.trim().toLowerCase()) !== -1
+			if (this.query.trim() !== '') {
+				return contact.searchData.toString().toLowerCase().search(this.query.trim().toLowerCase()) !== -1
 			}
 			return true
 		},
@@ -162,5 +171,14 @@ export default {
 // Add empty header to contacts-list that solves overlapping of contacts with app-navigation-toogle
 .contacts-list__header {
 	min-height: 48px;
+}
+
+// Search field
+.search-contacts-field {
+	padding: 5px 10px 5px 40px;
+
+	> input {
+		width: 100%;
+	}
 }
 </style>


szaimen:
Hi @AminulBD thanks for your PR!
Can you please add your div into the `contacts-list__header` div or remove that div and the corresponding css?
Thank you! :)

AminulBD:
Hi @szaimen

Please check now.

image

nimishavijay:
Super cool idea! :)

> One issue though is that this is the place for the New Contact button same as the Add Member button when managing circles

Good point. We could do something similar to Talk where the search and "Create new" button are on the same row in the navigation?

image

We don't use this pattern anywhere else but I like it and I think we could use it more. @jancborchardt any thoughts?

szaimen:
@AminulBD would you be up for implementing the above solution for the memberslist component? The contactslist component looks already good, imo. If you don't want to do it I would recommend to do the enhancements of the memberslist in a follow-up...

szaimen (commented):
I tested this and it seem to work but maybe there should be a delete all button when you have typed something in?
image

jancborchardt:
Yup, great work @AminulBD! :) As @nimishavijay said, if you move it to the top of the navigation column instead of the people column then it’s consistent with Talk.

szaimen:
> if you move it to the top of the navigation column instead of the people column then it’s consistent with Talk.

It would be consistent with talk but does not make sense here imo because the entities are not in the navigation column...

szaimen:
can we bring this forward? Would be a cool enhancement! :)

st3iny:
I strongly agree with what Simon said earlier about the placement. I'd rather have this merged and released as is than to be forgotten. We can improve the placement later when a global decision has been made.

Currently, the space where the search bar was inserted is empty and can/should be used.

src/components/ContactsList.vue
```diff
@@ -22,7 +22,11 @@
 
 <template>
 	<AppContentList>
-		<div class="contacts-list__header" />
+		<div class="contacts-list__header">
+			<div class="search-contacts-field">
+				<input type="text" placeholder="Search contacts..." v-model="query">
```
st3iny, line 27:
```suggestion
				<input type="text" :placeholder="t('contacts', 'Search contacts...')" v-model="query">
```

st3iny (commented):
Tested and works. Please rebase to latest main to fix CI.

AminulBD:
> can we bring this forward? Would be a cool enhancement! :)

@szaimen, I will check further. I was busying with my company tasks.

AminulBD:
> Hi @AminulBD if you want, I can try to take over to get this finished :)

@szaimen yea, it will be great if you do.

szaimen:
@st3iny does this look good to you? :)

src/components/ContactsList.vue
szaimen: done

src/components/ContactsList.vue
```diff
@@ -22,7 +22,11 @@
 
 <template>
 	<AppContentList>
-		<div class="contacts-list__header" />
+		<div class="contacts-list__header">
+			<div class="search-contacts-field">
+				<input type="text" :placeholder="t('contacts', 'Search contacts...')" v-model="query">
```
st3iny, line 27:
```suggestion
				<input type="text" :placeholder="t('contacts', 'Search contacts …')" v-model="query">
```

Ref #2804

szaimen:
> Tested and works. Please squash commits and we are ready.

done

kesselb:
Hey @AminulBD :wave: 

Thanks for your great contribution :+1: We release contacts 4.2 yesterday and already got some positive feedback about the quick search: https://help.nextcloud.com/t/quicksearch-the-better-search-option-instead-of-global-search-imho/142475

PR #2727, opened 2022-05-03
Fix screenshot link in appstore

nickvergessen:
See missing screenshot on https://apps.nextcloud.com/apps/contacts

After merging a new version needs to be published for the image to show up


szaimen (approved):
strange

nickvergessen:
"strange"?

The screenshot is not in this repo here, but in the screenshots repo and that still uses the master branch, so the `s/master/main` replacement when this app switched was too generic?!

szaimen:
> The screenshot is not in this repo here, but in the screenshots repo and that still uses the master branch, so the `s/master/main` replacement when this app switched was too generic?!

ah that makes sense 👍

ChristophWurst (approved):

Issue #1865, opened 2020-10-19
Allow plain-text input for dates


isdnfan:
Adding a birthday to contact is too complicated as the property doesn't accept free input and is limited to the date-picker widget. The widget default to the current date so it is necessary to scroll for lot of year to select the birthday of the contact.

It would be much easier to enter the birthday using free text. The challenge of parsing the text into valid date could be addressed by displaying valid templates like **dd.mm.yyyy** or **mm/dd/yyyy**

Nextcloud 19.0.3 Contacts 3.4.0.

call-me-matt:
I agree, text input should be possible!

By reading your comment, it seems you are scrolling through the years manually one by one. At least that's what I did in the beginning, until I discovered one can simply click on the year:

![birthday](https://user-images.githubusercontent.com/34400929/96502589-4d64c500-1252-11eb-8b30-fd56ec9f0be8.gif)

jangop:
This is probably one of the biggest issues I have with the current Contacts. In my opinion, text input should take precedence over a date picker, because it is faster (arguably), can be copied & pasted conveniently, and does not require an extra click on `OK`.

Of course, this does raise the common question of how to process dates input as plain text, especially considering that partial dates (e.g., without year, or without day, or only year) are valid values. [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) mitigates this issue well, and I think it should be used here (cf. https://tools.ietf.org/html/rfc6350#section-4.3.1):

|                            |              |
|----------------------------|--------------|
| Date                       | `2020-10-30` |
| Date without year          | `--10-30`    |
| Date without day           | `2020-10`    |
| Date without month and day | `2020`       |

jangop:
Maybe rename this issue as `Allow plain-text input for dates`.

meonkeys:
Could we copy the date picker used for when you set an expiration date for a shared file/folder (I think it is the "File sharing" app)? That accepts typed-in dates like `2020-12-30`, and also has a graphical date picker.

skjnldsv:
> Could we copy the date picker used for when you set an expiration date for a shared file/folder

This is the same used :)

meonkeys:
> This is the same used :)

Oh, that's odd, that other date picker *does* accept typed-in date strings.

jangop:
> > This is the same used :)
> 
> Oh, that's odd, that other date picker _does_ accept typed-in date strings.

Very odd indeed. The two pickers also use different format to display current values. 🤔

[...]

PR #2732, opened 2022-05-05
Allow to input the Date by hand

MrPompom:
Add Parse in dateFormat

fixes https://github.com/nextcloud/contacts/issues/1865

### DIFF
--- a/src/components/Properties/PropertyDateTime.vue
+++ b/src/components/Properties/PropertyDateTime.vue
@@ -133,6 +133,9 @@ export default {
 				stringify: (date) => {
 					return date ? this.formatDateTime() : null
 				},
+				parse: (value) => {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Do you know if there is a ticket about this?

How do I test this?

MrPompom:
There are a ticket about this : Allow plain-text input for dates #1865

You can test this by create a contact, add a date like birthday and type the date
image

you can type:
oct 11, 2011
october 11, 2011
10/11/2011
10-11-2011

JuliaKirschenheuter:
@MrPompom Thank you for taking over!

Could you please rename your branch, that it would be a bit precisely?

JuliaKirschenheuter (changes requested):
@MrPompom Thank you for taking over!

Could you please rename your branch, that it would be a bit precisely?

Tested - works)

PR #2737, opened 2022-05-12
Make sure all the fiels are blank when a new contact is created

MrPompom:
Duplicate the default value before replacing it if it's an array

Signed-off-by: Bastien PROMPSY <bastienprompsy@gmail.com>


ChristophWurst:
@MrPompom could you please elaborate on the problem this fixes? For review and testing I would prefer to have some reproduction steps, either here or in a related bug report. Then we can check main and you branch to see if/how the problem is fixed :)

MrPompom:
Before the commit:

When Value is put in the multiple input, the value is not deleted when a new contact is created

image

image


After the commit:

The modification make sure all field are blank when a new contact is created

JuliaKirschenheuter (approved):
Works fine, thank you!

Issue #2408, opened 2021-08-25
Input from the „Company“ field is copied to the „Name“ input field if the name field is empty


jerdesign:
### Describe the bug
When I create a new contact and fill out the company before the name, the name gets overwritten with the text I just entered in the company field.

**To Reproduce**
1. Go to the contacts app and create a new contact
2. Leave the name empty
3. Enter some text in the company field
4. Wait a few seconds
5. The text you just entered in the company field is copied to the name field

**Expected behavior**
Text from different fields is not copied to other fields.

**Actual behavior**
The input from the company field is copied to the name input field if the name field is empty.

**Screenshots**
![CleanShot 2021-08-25 at 16 26 59@2x](https://user-images.githubusercontent.com/4391042/130810082-4ebaa953-b976-4ea1-90fe-59c5126760a2.png)

### Server configuration

**Operating system**: RHEL8

**Web server:** Nginx

**Database:** MariaDB 

**PHP version:** 7.4 fpm

**Nextcloud version:** 22.1.0

**Contacts version:** 4.0.0

**Updated since Nextcloud 20**

**List of activated apps:**
```
Enabled:
  - accessibility: 1.7.0
  - admin_audit: 1.11.0
  - circles: 22.1.0
  - cloud_federation_api: 1.4.0
  - comments: 1.11.0
  - contacts: 4.0.0
  - contactsinteraction: 1.2.0
  - dashboard: 7.1.0
  - dav: 1.18.0
  - encryption: 2.9.0
  - external: 3.9.0
  - federatedfilesharing: 1.11.0
  - files: 1.16.0
  - files_pdfviewer: 2.3.0
  - files_rightclick: 1.1.0
  - files_sharing: 1.13.2
  - files_trashbin: 1.11.0
  - files_versions: 1.14.0
  - files_videoplayer: 1.11.0
  - group_everyone: 0.1.8
  - logreader: 2.7.0
  - lookup_server_connector: 1.9.0
  - nextcloud_announcements: 1.11.0
  - notifications: 2.10.1
  - oauth2: 1.9.0
  - oidc_login: 2.0.4
  - provisioning_api: 1.11.0
  - recommendations: 1.1.0
  - serverinfo: 1.12.0
  - settings: 1.3.0
  - spreed: 12.0.1
  - systemtags: 1.11.0
  - text: 3.3.0
  - theming: 1.12.0
  - theming_customcss: 1.9.0
  - twofactor_backupcodes: 1.10.1
[...]

PR #2739, opened 2022-05-12
Stop overwritting the name of the new contact by the company text

MrPompom:
Delete the condition of missingfn that add the company text to the name field when it's missing

Fixes https://github.com/nextcloud/contacts/issues/2408

### DIFF
--- a/src/services/checks/missingFN.js
+++ b/src/services/checks/missingFN.js
@@ -46,18 +46,6 @@ export default {
 				return true
 			}
 			return false
-		} else if (contact.vCard.hasProperty('org')) {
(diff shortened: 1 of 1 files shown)


JuliaKirschenheuter:
@MrPompom Thank you for taking over =)

could you please give us a bit detailed description, what do you want to reach with this fix?

MrPompom:
This pull request is to fix this issue : https://github.com/nextcloud/contacts/issues/2408#issue-979269722

All the details is already explained in the issue, this fix only stop the company name to become the fullname when it's "New Contact".

st3iny (approved):
Tested and works.

Issue #1813, opened 2020-09-30
Group cannot be deleted while it is selected


nursoda:
### Describe the bug
While testing groups in Contacts I noticed that if one selects a group of contacts, and edits one of the selected group, I cannot delete THAT group from the contact. I think this was different before: In prior version that contact still would remain to be displayed in the details section while it disappeared in the contacts list (since it no longer belonged to this group). Now, It disappears completely (which is OK also), but it should be saved before.

It is possible to select "all contacts" and remove the group then of course. However, if you have many contacts, that's not handy.

**To Reproduce**
1. have one contact 'test' in a group 'test'
2. select it
3. delete its group by "removing" the checkmark within the group dropdown (in the contact details section on the right hand side)

**Expected behavior**
Contact should be saved without the group set that was just deleted.

**Actual behavior**
Removal of this group is not saved.

### Configuration
Arch Linux (current), NGINX 1.18, PHP-FPM 7.4, MariaDB 10.5
Nextcloud 19.0.3.1, Contacts 3.4.0, NC: updated, Signing: OK
Browser: Firefox 81.0 on Arch Linux (current)

skjnldsv:
Thank you so much for testing the 3.4.0 @nursoda :)

PR #2744, opened 2022-05-16
When delete last member from group, it is not update data on server

sazanof:
Fixes https://github.com/nextcloud/contacts/issues/1813

Good afternoon! There is a bug, when I am in the selected group and delete the last contact from it, no data is sent to the server to update a contact.

Corrected in this way.

Thanks!

### DIFF
--- a/src/components/Properties/PropertyGroups.vue
+++ b/src/components/Properties/PropertyGroups.vue
@@ -166,6 +166,11 @@ export default {
 				contact: this.contact,
 				groupName,
 			})
+			const group = this.$store.getters.getGroups.find(search => search.name === groupName)
+			if (group.contacts.length === 0) {
+				this.$emit('update:value', [])
+			}
+
 		},
 
 		/**


GretaD:
I cannot reproduce the initial problem anymore. After updating my main and building it the contacts are added and removed from the group in real time. 
cc @JuliaKirschenheuter

sazanof:
> I cannot reproduce the initial problem anymore. After updating my main and building it the contacts are added and removed from the group in real time. cc @JuliaKirschenheuter

Hello, Greta!
Try to add one user to the group. go to this group(contains one user). try deleting the user from group via multiselect and then refresh the page

JuliaKirschenheuter:
> Hello, Greta! Try to add one user to the group. go to this group(contains one user). try deleting the user from group via multiselect and then refresh the page

Hey @sazanof, 

thank you for taking care!
I have tested it and a update request will be sent. But if there was only one contact in a group, _after deleting and updating a page_ i have 2 error messages:

- Group group not found
- Contact not found

It seems that a group does't exists anymore, but fetch to get a group and users inside of this group will be sent. 
Please try to reproduce it. I think it is another bug, if it is completely different, let me know, i will open another bug.

sazanof:
Hello, @JuliaKirschenheuter !
> have 2 error messages

Yes, there is such a thing.  

Exactly as you said. Perhaps after deleting the last contact from the group, you need to redirect to "All contacts" (change the route to the base one), and also delete the group from the list of groups. 

Why didn't I implement it right away: you see, since a group is some kind of implicit entity that exists only in contact. And if there is not a single contact, in fact there is no this group either (it just hangs on the front-end). What if I want to add other people to the group after I've thrown everyone out? If I don't reload the page, then just this pseudo-group will remain and I will be able to do it. I think I used the translator correctly and explained. =)

Let me know if this is necessary (and what exactly) to implement within the framework of this PR. In theory, as I said above - change the route. 

If yes, then I will try to correct this point as well.

JuliaKirschenheuter:
Thank you once again for your opinion!

I've discussed it internally and this "2 error messages" would be another bug, i will open it. You can leave your bug fixing as it is, which fixes this https://github.com/nextcloud/contacts/issues/1813

> Why didn't I implement it right away: you see, since a group is some kind of implicit entity that exists only in contact. And if there is not a single contact, in fact there is no this group either (it just hangs on the front-end). What if I want to add other people to the group after I've thrown everyone out? If I don't reload the page, then just this pseudo-group will remain and I will be able to do it. I think I used the translator correctly and explained. =)

CardDAV groups are similar to "tags". They could only take place, if there is at least one contact. If last contact goes away from a group, the whole group will disappear. We could not change this behavior.

JuliaKirschenheuter:
Please run ```npm run lint:fix``` to fix linter issues

JuliaKirschenheuter:
> ok, after that commit in PR?

You can do it each time, it does't matter. But please combine at the end all commits to final one, this is a requirement for merging ;)

JuliaKirschenheuter (approved):

PR #2785, opened 2022-06-14
Avoid running social update background job on deleted/disabled users

tcitworld:


call-me-matt (approved):
That is a good improvement, thank you

PR #2798, opened 2022-06-27
Fix mixing jest v27 and v28

ChristophWurst:
Ref https://github.com/nextcloud/contacts/pull/2754#pullrequestreview-997726710

We need to go from v27 to v28 in one jump and bump all related packages.


GretaD (approved):
npm ci && npm run build works fine while on this pr. Yuhu

PR #2801, opened 2022-06-28
Migrate left sidebar to material design icons

GretaD:
before
beforeleftsidebar
after
leftsidebar_contact

please test


ChristophWurst:
Are you going to do the actions in another PR?

Bildschirmfoto vom 2022-06-28 16-07-29

ChristophWurst (approved):
Seems to do the trick otherwise

PR #2806, opened 2022-06-29
Migrate contact details icons to material design icons

GretaD:
ref https://github.com/nextcloud/groupware/issues/38


GretaD:
delivered

PR #2930, opened 2022-09-16
Release v5.0.0-alpha.1

GretaD:
Signed-off-by: greta <gretadoci@gmail.com>


ChristophWurst (changes requested):
No release, please. We only need to *bump* the version.

ChristophWurst:
We will do the first release when the new UI has landed.

PR #2954, opened 2022-09-22
Add organization charts

ChristophWurst:
Updated and continued version of https://github.com/nextcloud/contacts/pull/2726


GretaD:
looks good, i noticed just this small "down/up arrow" that is cut a bit
Screenshot from 2022-09-22 19-04-20

PR #2972, opened 2022-09-23
l10n: Change three dots to an ellipsis

Valdnet:
Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>

### DIFF
--- a/src/components/OrgChart.vue
+++ b/src/components/OrgChart.vue
@@ -56,7 +56,7 @@ export default {
 			})
 		},
 		placeholder() {
-			return t('contacts', 'Select chart...')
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

tcitworld (commented):
I was about to say that I wouldn't have put a space before the ellipsis but it seems it's a mess anyway.
https://en.wikipedia.org/wiki/Ellipsis#In_different_languages

Valdnet:
@rakekniven  Here you go: [English Style Guide](https://ec.europa.eu/info/sites/default/files/styleguide_english_dgt_en.pdf).

Issue #3010, opened 2022-10-03
Can not unfold some people on the org chart


nickvergessen:
### Steps to reproduce

1. Go to the org chart on our instance, unfold the Head of Engineering
2. Try to unfold V…P…
![Bildschirmfoto vom 2022-10-03 09-42-05](https://user-images.githubusercontent.com/213943/193525630-322c09fd-98fd-41c6-83e0-b11b0dda9bd9.png)

### Expected behavior

Should be able to unfold

### Actual behavior

First need to unfold all engineering leads which are after him in the list:

![Bildschirmfoto vom 2022-10-03 09-42-40](https://user-images.githubusercontent.com/213943/193525635-6694ea66-3ba2-4961-a41f-ebc815038e12.png)

### Contact version

5rc1

PR #3024, opened 2022-10-07
Fix org chart card overlapping other card expand button

ChristophWurst:
Fixes https://github.com/nextcloud/contacts/issues/3010

| Before | After |
|---|---|
| Bildschirmfoto vom 2022-10-07 14-39-49 | Bildschirmfoto vom 2022-10-07 14-39-24 |

### DIFF
--- a/src/components/ChartTemplate.vue
+++ b/src/components/ChartTemplate.vue
@@ -66,22 +66,22 @@ export default {
 <style lang="scss" scoped>
 .org-chart-node {
 	background-color: var(--color-main-background);
-	padding-top: 1px;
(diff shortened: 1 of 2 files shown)


st3iny (changes requested):
Tested and works so far.

However, the avatars are not centered anymore.

image

Removing the following line fixes the problem:
https://github.com/nextcloud/contacts/blob/4b0b7b1ee13a9f150cebefd15449b1d4f034bc3a/src/components/ChartTemplate.vue#L75

st3iny (approved):
Tested and works.

PR #3040, opened 2022-10-14
Improve the new contact button design

GretaD:
fixes # [#2966](https://github.com/nextcloud/contacts/issues/2966)
and contact part of https://github.com/nextcloud/groupware/issues/2

### DIFF
--- a/src/components/ContactsList.vue
+++ b/src/components/ContactsList.vue
@@ -23,6 +23,19 @@
 <template>
 	<AppContentList class="content-list">
 		<div class="contacts-list__header">
+			<div>
+				<!-- new-contact-button -->
+				<Button
+					type="primary"
+					button-id="new-contact-button"
+					:disabled="!defaultAddressbook"
+					@click="newContact">
+					<template #icon>
+						<IconAdd :size="20" />
+					</template>
+					{{ t('contacts','New contact') }}
+				</Button>
+			</div>
 			<div class="search-contacts-field">
 				<input v-model="query" type="text" :placeholder="t('contacts', 'Search contacts …')">
 			</div>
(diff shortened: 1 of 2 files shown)


GretaD:
Which version looks better? (Its not styled properly with the right width and stuff, its just for you to see if it should be inline or above each other)

Version 1
Screenshot from 2022-10-14 15-43-17
Version 2
Screenshot from 2022-10-14 15-41-10

ChristophWurst:
> Which version looks better?

Second but first might be necessary for narrow screens

GretaD:
> > Which version looks better?
> 
> Second but first might be necessary for narrow screens

yes, sorry, i just added a comment on my screenshoot

ChristophWurst:
Make them a column for narrow screens and a row for wide screens. In both cases the search input can grow to take all available space.

ChristophWurst:
Before we get into a much more complex topic than necessary, let's have another look at https://github.com/nextcloud/contacts/issues/2966. I believe the only change requested is to left-align the icon and text. Moving the button somewhere else is a bigger change and out of scope.

GretaD:
> Before we get into a much more complex topic than necessary, let's have another look at #2966. I believe the only change requested is to left-align the icon and text. Moving the button somewhere else is a bigger change and out of scope.

ok, because we agreed to change the new button position, i thought to do it in one PR. This is ready for review, but i am fine to change only the button on the sidebar anyway, and merge this one whenever we want to change the position of the button.

ChristophWurst:
Yes we want to but as your screenshots show it's not clear how to move on with the moved button.

Could you create a clean PR for just the alignment of the button icon and text?

src/components/ContactsList.vue
```diff
@@ -157,6 +187,59 @@ export default {
 			}
 			return true
 		},
+		async newContact() {
+			const rev = new VCardTime()
```
ChristophWurst, line 191: where's that code from? it appears moved. please delete the now unused method from the previous component

GretaD: its from here: https://github.com/nextcloud/contacts/blob/main/src/views/Contacts.vue#L38

i dont remember why i didnt remove it, but i dont think its because i forgot. I will test it asap and see

jancborchardt:
@GretaD this seems outdated and could be closed, right? Now that the navigation has the "Create" button and the middle column has the search bar, all seems good.

PR #3070, opened 2022-10-21
Add manually-triggered branch-off workflow

ChristophWurst:
This allows me to branch off from the web UI. Handy when I'm too lazy to do this from the CLI.


ChristophWurst:
let's not

Issue #3058, opened 2022-10-19
Nextcloud 25 organization chart : When a manager's contact file is deleted, it crashes the organization chart display


Jerome-Herbinet:
### Describe the bug

While testing this new feature, I deleted the manager of Dark Sidious and then the chart cannot be displayed anymore. 
To make it appear again, I have to delete de "Manager" field entry in the Dark Sidious contact file.

![2022-10-19_11-42](https://user-images.githubusercontent.com/33763786/196656851-c333bd9f-83e4-4246-8554-a0984b5e43dd.png)

### Steps to reproduce

1. Create an organization chart
2. Delete a contact which is manager of another contact
3. See that you cannot display the chart anymore
4. Delete the manager field entry in the contact file which was underneath the previously deleted contact in the chart
5. See that the chart can be displayed again

### Expected behavior

When a "manager" contact is deleted, all dependencies should be deleted in order to avoid chart explosion.

### Actual behavior

It crashes because of remaining broken dependencies.

### Contact version

5.0.1

### Operating system

CentOS

### PHP engine version

PHP 8.0

### Web server

Nginx

### Database

MySQL

bkrith:
I will check it.. i think the issue is that we scan for the relationships but we don't check if the manager exists.. maybe on deletion we need to scan contacts and make null manager deleted id references.

bkrith:
@ChristophWurst I have a PR for this please review it when you have time. (#3077)

PR #3077, opened 2022-10-23
Update related contacts on manager deletion

bkrith:
Fix issue bug #3058 

Before deletion remove property managersName reference for each related contact.

### DIFF
--- a/src/components/AppContent/ContactsContent.vue
+++ b/src/components/AppContent/ContactsContent.vue
@@ -70,7 +70,7 @@
 		</template>
 
 		<!-- main contacts details -->
-		<ContactDetails :contact-key="selectedContact" :contacts="sortedContacts" />
+		<ContactDetails :contact-key="selectedContact" :sorted-contacts="sortedContacts" :contacts="contacts" />
 	</AppContent>
 </template>
 <script>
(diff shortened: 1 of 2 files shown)


src/components/AppContent/ContactsContent.vue
```diff
@@ -70,7 +70,7 @@
 		</template>
 
 		<!-- main contacts details -->
-		<ContactDetails :contact-key="selectedContact" :contacts="sortedContacts" />
+		<ContactDetails :contact-key="selectedContact" :sorted-contacts="sortedContacts" :contacts="contacts" />
```
ChristophWurst, line 73: is this change really necessary for the fix?

ChristophWurst:
@bkrith please do the following two things

1) Move sort changes out into a separate PR. We can release this fix with the next minor/major. The fix in this PR will be backported.
2) Add logic to catch externally deleted manager. This app is just one of the possible applications that access Nextcloud's CardDAV APIs. If someone deletes a manager on their phone, Thunderbird, etc, then we also don't want to cause an error. I would suggest to just ignore references to other contacts that no longer exist. We don't necessarily have to update all cards.

ChristophWurst:
@bkrith friendly :ping_pong: . Please have a look at the review feedback above when you have time :)

Jerome-Herbinet:
Hi @bkrith ! As @ChristophWurst said, can you take a look at the review feedback above ? Thanks !

Adaniieel15:
Hello, I hope you are well.

In our Nextcloud instance we removed the main administrator and the org chart stopped showing.

I was wondering if they will be looking for a solution for this soon? Or if there are some steps I can take to fix it?

PR #3117, opened 2022-12-01
Revert "Bump vue and vue-template-compiler"

ChristophWurst:
This reverts commit f07b9c238bf31bd450d2af4a933e2becd7079965 and downgrades a few related packages.

We had to downgrade before: https://github.com/nextcloud/contacts/pull/2928.

Until the virtual scroller workds with 2.7 we can not update Vue nor the template compiler.

## How to test
1) Open the app
2) Add a contact if you have none
3) Look at the contacts list

Main: no contacts rendered
Here: contacts rendered

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,9 +37,9 @@
 				"qr-image": "^3.2.0",
 				"string-natural-compare": "^3.0.1",
 				"uuid": "^9.0.0",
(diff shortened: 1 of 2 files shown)


GretaD (approved):
i cannot run npm i on main because of bump, but i can run it here

PR #3165, opened 2023-01-06
chore: scope phpunit

kesselb:
Motivation to scope PHPUnit is the removal of PHP 7.4 support. 

With config.platform.php = 8.0 composer will install a newer PHPUnit stack.  phpunit/php-code-coverage require at least nikic/php-parser 1.14 but 3rdparty nikic/php-parser is 1.10.5.

The incompatible installations cause the coverage run to fail:

`SebastianBergmann\CodeCoverage\StaticAnalysis\CodeUnitFindingVisitor::unionOrIntersectionAsString(): Argument #1 ($type) must be of type PhpParser\Node\ComplexType, PhpParser\Node\UnionType given, called in /home/runner/work/contacts/contacts/apps/contacts/vendor/phpunit/php-code-coverage/src/StaticAnalysis/CodeUnitFindingVisitor.php on line 184`

Signed-off-by: Daniel Kesselberg <mail@danielkesselberg.de>


.github/workflows/phpunit.yml
```diff
@@ -122,7 +122,7 @@ jobs:
 
       - name: Set up PHPUnit
         working-directory: apps/${{ env.APP_NAME }}
-        run: composer i
+        run: composer install && composer bin phpunit install
```
ChristophWurst, line 125: could you do that through the hook? https://github.com/nextcloud/mail/blob/dff7a2751ca340b9f104d9103d291d1bd83aff73/composer.json#L58

ChristophWurst (commented):
:+1: noice

Please add dependabot config for the new vendor-bin

PR #3173, opened 2023-01-13
allow social avatars from telegram, fixes #2393

### LINKED ISSUE #2393: Telegram avatar contacts sync
I really like the idea behind the "Update avatars from social media" feature for contacts. However, I don't believe it currently supports telegram, which natively adds telegram profile links to contacts on iOS. Is there a way that support could be added for this? I currently don't use Facebook and don't see a list anywhere of what other social networks the feature supports.

call-me-matt:
Fix #2393

This PR allows to fetch avatar pictures from telegram messenger profiles.

Q: maybe there is a smarter way to fetch ical, looking for something like:
`this.contact.vCard.getAllProperties(['x-socialprofile', 'impp'])` instead of getting each one and then concatenating them.


st3iny (approved):
Tested and works. Very nice!

PR #3187, opened 2023-01-20
perf(autoloader): Use Composer's authoritative classmap

ChristophWurst:
Theoretically this could speed things up but I'm not able to proof that with blackfire nor simple timings.

## Before optimizing all four Groupware apps

Bildschirmfoto vom 2023-01-20 13-21-55

## After optimizing all four Groupware apps

Bildschirmfoto vom 2023-01-20 13-22-06

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,12 @@
 		"classmap-authoritative": true,
 		"platform": {
 			"php": "7.4"
+		},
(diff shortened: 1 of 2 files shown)


kesselb (approved):

kesselb:
CI failure unrelated

PR #3213, opened 2023-01-30
Fix : Move the labels away from the text fields

Jerome-Herbinet:
The problem :

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,8 @@
 		"ts-jest": "^29.0.5",
 		"ts-loader": "^9.4.2",
 		"typescript": "^4.9.4",
-		"vue-template-compiler": "~2.6"
+		"vue-template-compiler": "~2.6",
+		"webpack": "^5.75.0"
 	},
 	"browserslist": [
 		"extends @nextcloud/browserslist-config"
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
 		"qr-image": "^3.2.0",
 		"string-natural-compare": "^3.0.1",
 		"uuid": "^9.0.0",
-		"vue": "^2.7.14",
+		"vue": "~2.6.14",
 		"vue-click-outside": "^1.1.0",
 		"vue-masonry": "^0.16.0",
 		"vue-material-design-icons": "^5.1.2",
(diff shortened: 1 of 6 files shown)


ChristophWurst:
Could you also post an *after* screenshot? thanks

Jerome-Herbinet:
> Could you also post an _after_ screenshot? thanks
@ChristophWurst

src/components/Properties/PropertyMultipleText.vue
```diff
@@ -142,6 +142,7 @@ export default {
 	user-select: none;
 	text-align: right;
 	background-size: 16px;
+	padding-right: 7px;
```
nimishavijay, line 145: Good catch! Since this is very close to 8px, we could use the variable `calc(var(--default-grid-baseline) * 2)` which would come up to 8px, what do you think? :)

package.json
```diff
@@ -100,7 +100,8 @@
 		"ts-jest": "^29.0.5",
 		"ts-loader": "^9.4.2",
 		"typescript": "^4.9.4",
-		"vue-template-compiler": "~2.6"
+		"vue-template-compiler": "~2.6",
+		"webpack": "^5.75.0"
```
ChristophWurst, line 104: unrelated change? webpack should come as peer of ``@nextcloud/webpack-vue-config``

Jerome-Herbinet: solved !

src/components/Properties/PropertyMultipleText.vue
Jerome-Herbinet: done

package.json
```diff
@@ -64,7 +64,7 @@
 		"qr-image": "^3.2.0",
 		"string-natural-compare": "^3.0.1",
 		"uuid": "^9.0.0",
-		"vue": "~2.6.14",
+		"vue": "^2.7.14",
```
ChristophWurst, line 67: unrelated?

ChristophWurst (changes requested):
revert all changes in package*

package.json
Jerome-Herbinet: revert is done

Issue #2866, opened 2022-07-23
Quotes in TYPE are still being set


nicokaiser:
### Describe the bug

This is essentially #1107, which is still unresolved (but cannot be re-opened):

In the TYPE attribute, quotes are added:

```
EMAIL;TYPE="INTERNET,HOME,pref":user@example.com
TEL;TYPE="HOME,VOICE,pref":+49 123 456789
TEL;TYPE="CELL,VOICE":+49 123 456789
```

This leads to several platforms not interpreting these values correctly. On iOS this entry is being displayed as `"CELL` (presumably split at the `,`).

In https://github.com/nextcloud/contacts/issues/1107#issuecomment-803477441 it is mentioned that changing the vCard version to `4.0` may fix this, so either the version has to be correct, or the quotes have to be removed.

### Steps to reproduce

1. Add a contact
2. Add a cell phone numer
3. Sync to iOS via CardDAV
4. See the entry as `"CELL`

### Expected behavior

The entry type should be `Mobile` (or something localized by iOS) instead of `"CELL`.

### Actual behavior

The entry type is displayed as `"CELL`.

### Contact version

4.2.0

OmarWazzan:
> In https://github.com/nextcloud/contacts/issues/1107#issuecomment-803477441 it is mentioned that changing the vCard version to 4.0 may fix this, so either the version has to be correct, or the quotes have to be removed.

I can confirm that this does not work. I changed the version number on my contacts list before importing at the quotes were still added.

I also tried importing the contacts, exporting them, manually fixing the field, and then re-importing it and the issue still appeared.

However, if I imported a contact, and then reset the TYPE attribute (i.e, from mobile -> none -> mobile), then the field is properly stored.

This leads me to believe that the issue is not with all contacts, but rather with import/export of them.

HooK81:
I had this issues from the beginning, but today I have updated NextCloud to v24 and Contacts app to v4.2.0.

At first the issue was still there. See on my phone "HOME" instead of "domicile" (in french). 
I had to update manually phone type on all my contacts in the NextCloud web app.
Now the problem is gone. 

Types are correct on NextCloud web app and fortunately on my iPhone using iOS 15.6.

I guess there were something recorded badly in the contacts informations.

OmarWazzan:
> I had to update manually phone type on all my contacts in the NextCloud web app.

Unfortunately that's an infeasible solution for those of us with 500+ contacts

nicokaiser:
[...]

PR #3243, opened 2023-02-23
remove quotes on import

caplod:
to comply with https://www.rfc-editor.org/rfc/rfc6350#section-5

see also https://github.com/nextcloud/contacts/issues/2866


st3iny (approved):
Tested and works.

PR #3281, opened 2023-03-20
Allow admin to manage circle settings

artonge:


ChristophWurst (approved):
Code looks good. Can't judge the semantics.

PR #3290, opened 2023-03-29
fix error when contact has no email

caplod:
If you delete the email in a contact and then try to send an email to a group with that contact you get an error:

```TypeError: t.contacts[e].email is null```

here I add a check if the contact has an email.

### DIFF
--- a/src/components/AppNavigation/GroupNavigationItem.vue
+++ b/src/components/AppNavigation/GroupNavigationItem.vue
@@ -153,11 +153,7 @@ export default {
 		 */
 		emailGroup(group) {
 			const emails = []
-			group.contacts.forEach(key => {
-        // break if contact has no email
-				if (this.contacts[key].email == null) { 
-					return
-				}
+			group.contacts.filter(key => this.contacts[key].email !== null).forEach(key => {
 				// The email property could contain "John Doe <john.doe@example.com>", but vcard spec only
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/components/AppNavigation/GroupNavigationItem.vue [hunk 1]
```diff
@@ -154,6 +154,10 @@ export default {
 		emailGroup(group) {
 			const emails = []
 			group.contacts.forEach(key => {
+        // break if contact has no email
+				if (this.contacts[key].email == null) {
```
ChristophWurst, line 158: is the prop `null` or undefined? let's use strong comparison with `===`

src/components/AppNavigation/GroupNavigationItem.vue
```diff
@@ -154,6 +154,10 @@ export default {
 		emailGroup(group) {
 			const emails = []
 			group.contacts.forEach(key => {
```
ChristophWurst, line 156: nitpick: you can add a `.filter` before `forEach` to filter out these contacts

caplod: ok, I added a filter method

ChristophWurst (approved):
:sunglasses: 

If you can squash your commits into one, then we can :shipit:

PR #3302, opened 2023-04-10
Closes #1646 Set focus into new added property

### LINKED ISSUE #1646: Set focus into new added property
I'm always frustrated when I add a new property in a contact, that I have to select the newly created input field with the mouse, before I can start typing the new content for the new property.

Please automatically set the focus of the cursor in the new field

BartBarends:
Closes #1646,

Makes it so that if a user adds a new Property to a contact, it focusses the first input of that property


src/components/ContactDetails/ContactDetailsAddNewProp.vue
```diff
@@ -139,6 +139,15 @@ export default {
 					property.setParameter('type', defaultData.type)
 				}
 			}
+
+			//We want to wait until the new property is rendered
+			this.$nextTick(() => {
+				let lastAddedElementOfType = document.getElementsByClassName("property--last property-" + id)[0];
```
ChristophWurst, line 145: Query selectors are not very Vue-y. Could there be a way to acquire the element via a ref https://vuejs.org/guide/essentials/template-refs.html?

ChristophWurst: We ran into the same issue with https://github.com/nextcloud/contacts/issues/3304 / https://github.com/nextcloud/contacts/pull/3314, and the changes even try to achieve the same goal. https://github.com/nextcloud/contacts/pull/3314 has a wider scope of improvements. Would it be okay to close this PR?

ChristophWurst: FYI https://github.com/vuejs/vue/issues/4225#issuecomment-1199937033 told us that focus selection for custom Vue components is not easy, so we found a tradeoff with component refs and scoped query selectors.

PR #3303, opened 2023-04-12
feat: expose system address book

miaulalala:
Part of: https://github.com/nextcloud/server/issues/19575


lib/Controller/SocialApiController.php
```diff
@@ -66,7 +66,7 @@ public function __construct(IRequest $request,
 	 * @returns {JSONResponse} an empty JSONResponse with respective http status code
 	 */
 	public function setAppConfig($key, $allow) {
-		$permittedKeys = ['allowSocialSync'];
+		$permittedKeys = ['allowSocialSync', 'allowSystemAddressBook'];
```
miaulalala, line 69: I think this is the wrong place to put this.

miaulalala:
This can be closed as the changes are in the DAV backend only. No initial state neccessary.

PR #3314, opened 2023-04-18
fix: Improve design add new Property

JohannesGGE:
Fix: #3304 

Improved design of new Property
- updated rfcProps
- Add new Icons
- focus prop if already exists

|before|after|
|---|---|
|Screenshot from 2023-04-20 11-07-14|Screenshot from 2023-04-20 11-04-57 Screenshot from 2023-04-20 11-05-17|

### DIFF
--- a/src/components/AppNavigation/GroupNavigationItem.vue
+++ b/src/components/AppNavigation/GroupNavigationItem.vue
@@ -156,6 +156,7 @@ export default {
 		 * Open mailto: for contacts in a group
 		 *
 		 * @param {object} group of contacts to be emailed
+		 * @param {string} mode
 		 */
 		emailGroup(group, mode = 'to') {
 			const emails = []
--- a/src/components/ContactDetails/ContactDetailsAddNewProp.vue
+++ b/src/components/ContactDetails/ContactDetailsAddNewProp.vue
@@ -21,38 +21,80 @@
   -->
 
 <template>
-	<div class="property property--without-actions property--last">
-		<!-- title -->
-		<PropertyTitle :icon="'icon-add'" :readable-name="t('contacts', 'Add new property')" />
-
-		<div class="property__row">
-			<div class="property__label" />
-
-			<!-- type selector -->
-			<Multiselect :options="availableProperties"
-				:placeholder="t('contacts', 'Choose property type')"
-				class="property__value"
-				track-by="id"
-				label="name"
-				@input="addProp" />
-		</div>
+	<div>
+		<Actions menu-align="right"
+			event=""
+			type="secondary"
+			:menu-title="t('contacts', 'Add more info')"
+			@click.native.prevent>
+			<template #icon>
+				<IconAdd :size="20" />
+			</template>
+			<template v-if="!moreActionsOpen">
+				<ActionButton v-for="option in availablePrimaryProperties"
+					:key="option.id"
+					class="action--primary"
+					:close-after-click="true"
+					@click.prevent="addProp({id: option.id})">
+					<template #icon>
+						<PropertyTitleIcon :icon="option.icon" />
+					</template>
+					{{ option.name }}
+				</ActionButton>
+				<ActionButton :close-after-click="false"
+					@click="moreActionsOpen=true">
+					<template #icon>
+						<DotsHorizontalIcon :title="t('contacts', 'More fields')"
+							:size="20" />
+					</template>
+					{{ t('contacts', 'More fields') }}
+				</ActionButton>
+			</template>
+			<template v-if="moreActionsOpen">
+				<ActionButton :close-after-click="false"
+					@click="moreActionsOpen=false">
+					<template #icon>
+						<ChevronLeft :title="t('contacts', 'More fields')"
+							:size="20" />
+						{{ t('contacts', 'More fields') }}
+					</template>
+				</ActionButton>
+				<ActionButton v-for="option in availableSecondaryProperties"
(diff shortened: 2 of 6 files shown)


src/components/ContactDetails/ContactDetailsAddNewProp.vue
```diff
@@ -21,38 +21,82 @@
...
+					</template>
+					{{ option.name }}
+				</ActionButton>
+				<ActionButton :close-after-click="false"
+							  @click="moreActionsOpen=true">
+					<template #icon>
+						<DotsHorizontalIcon :title="t('mail', 'More fields')"
+											:size="20" />
+					</template>
+					{{ t('mail', 'More fields') }}
```
ChristophWurst, line 50: s/mail/contacts/

src/components/ContactDetails/ContactDetailsProperty.vue
```diff
@@ -103,6 +112,20 @@ export default {
...
+		bus: {
+			type: Object,
+			required: true,
+		},
+	},
+
+	created() {
+		this.bus.$on('focusProp', async (id) => {
+			if (id === this.propName) {
+				this.focus = true
```
ChristophWurst, line 124: would it be possible to use native focus with `this.refs.component.focus()` or similar? The styling can be done with the pseudo class `:focus`, if not existent already

JohannesGGE: I focus the first input in the prop now :+1:

src/components/AppNavigation/GroupNavigationItem.vue
```diff
@@ -156,6 +156,7 @@ export default {
 		 * Open mailto: for contacts in a group
 		 *
 		 * @param {object} group of contacts to be emailed
+		 * @param mode
```
ChristophWurst, line 159:
```suggestion
		 * @param {string} mode
```

src/components/ContactDetails/ContactDetailsAddNewProp.vue
```diff
@@ -21,38 +21,80 @@
...
-
-			<!-- type selector -->
-			<Multiselect :options="availableProperties"
-				:placeholder="t('contacts', 'Choose property type')"
-				class="property__value"
-				track-by="id"
-				label="name"
-				@input="addProp" />
-		</div>
+	<div>
```
ChristophWurst, line 24: wrapping div has only one element. drop the div?

src/components/ContactDetails/ContactDetailsAddNewProp.vue
```diff
@@ -92,20 +150,43 @@ export default {
...
+		/**
+		 * List of every properties you are allowed to add
+		 * on this contact
+		 *
+		 * @return {object[]}
+		 */
+		availableSecondaryProperties() {
+			return Object.keys(this.properties)
+				// only allow to add multiple properties OR props that are not yet in the contact
+				// .filter(key => this.properties[key].multiple || this.usedProperties.indexOf(key) === -1)
```
ChristophWurst, line 177: dead code?

ChristophWurst:
@JohannesGGE please add before/after screenshots to the PR descriptions for PRs with visual changes :)

src/components/ContactDetails/ContactDetailsAddNewProp.vue
JohannesGGE: I think it's the easiest way to center the popup to the button. Without the div, it is centered to the parent div.

image

src/components/ContactDetails/ContactDetailsProperty.vue
```diff
@@ -374,6 +378,20 @@ export default {
...
+			if (id === this.propName) {
+				this.$nextTick(() => {
+					if (this.$refs.component.$el.querySelector('input') === undefined) {
+						console.warn('no input to focus found')
+					} else {
+						this.$refs.component.$el.querySelector('input').focus()
+					}
+				})
+			}
+		})
```
st3iny, line 392: It's probably a good idea to move the logic to a method and call `this.bus.$off('focus-prop', this.onFocusProp)` in `destroyed()` to prevent leaks.

st3iny (commented):
I did some testing and it looks very nice and works overall.

However, the focusing did not work. I checked the console and there were no warnings (`no input to focus found`).

src/components/ContactDetails/ContactDetailsProperty.vue
JohannesGGE: Didn't it focus at all or just not on adding a new prop? The focusing is just implemented for the case that the prop already exists and it's not a multiple one. (Going to add this for the add case)

src/components/ContactDetails/ContactDetailsProperty.vue
```diff
@@ -374,7 +378,35 @@ export default {
 		},
 	},
 
+	created() {
+		this.bus.$on('focus-prop', (id) => {
+			this.onFocusProp(id)
+		})
```
st3iny, line 384:
```suggestion
		this.bus.$on('focus-prop', this.onFocusProp)
```

The handler won't be unregistered otherwise. Or is this a problem with the logic (especially concerning the usage of `this`)?

st3iny (commented):
Tested and works. There is an issue with masonry but that will be gone soon anyway.

st3iny (approved):
Tested and works.

Issue #655, opened 2018-09-29
Allow to add more properties directly from property


rullzer:
It would be great if I could add more phone numbers without going to `Add new property` selection. Just a + icon below the phone numbers/email/whatever I already have for a contact to add another one of those fields.

devingfx:
> going to `Add new property` selection. 

Hi @rullzer ! Where is this option? I can't find it and I'm unable to add any fields to my contacts :(

PR #3336, opened 2023-04-24
feat(contacts): add add-prop button direct in prop

JohannesGGE:
Fix: #655 

|before|after|
|--|--|
|image|image|


st3iny (approved):
Tested and works.

PR #3337, opened 2023-04-24
feat(contacts): crop for contact picture

JohannesGGE:
Fix: #1662 

Add image cropper for image upload/select from files. Do some refactoring as well.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,6 +39,7 @@
 				"uuid": "^9.0.0",
 				"vue": "~2.6.14",
 				"vue-click-outside": "^1.1.0",
+				"vue-cropperjs": "^4.2.0",
 				"vue-material-design-icons": "^5.2.0",
 				"vue-router": "^3.6.5",
 				"vue-virtual-scroll-list": "^2.3.4",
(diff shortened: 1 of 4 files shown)


nimishavijay (changes requested):
Looks great already! Super nice :) 2 points:

- [ ] We should add a heading in every modal, as a general good practice and for accessibility reasons.
- [ ] Also I'd suggest to have the picture fill the modal as much as possible, without showing a transparent background. Would this be possible? I've pasted a mockup (ignore the icons in the button and their positions, the main idea was to show what I meant with the positioning of the image)

image

Nice work! :)

JohannesGGE:
@nimishavijay I changed the behavior of the cropbox a bit. There is no transparent background anymore but unlike your suggestion the cropbox is still 300px/300px, the dynamic size led to smaller cropboxes in my tests. Let me know if you prefer the old or the new cropbox. 

[Screencast from 27.04.2023 16:50:06.webm](https://user-images.githubusercontent.com/74607597/234901304-edf018c7-dd0f-41ee-b750-17ff4f9faf5b.webm)

nimishavijay:
Hmm, if it's not technically possible to have the entire image shown then the new screencast looks good :) could we make the size of the modal smaller so that there is not so much space on either side?

JohannesGGE:
I did some further testing and found a suitable option, I think.
Only if the picture is much greater in height than in width (3) the modal extend to max height. Do you think that is okay? I believe nobody uses such a picture to crop 
|||
|--|--|
|Screenshot from 2023-04-28 10-53-48|Screenshot from 2023-04-28 10-55-39|

(3)
Screenshot from 2023-04-28 10-39-38

nimishavijay:
That sounds good to me! It is an edge case and we can open a follow up issue about that. Nice work :)

nimishavijay (approved):

ChristophWurst (approved):
Works!

I noticed that upload a pic for a contact with no avatar before doesn't show the uploaded avatar until I navigate to another contact and back

JohannesGGE:
@ChristophWurst Created an issue for this #3351

Issue #2187, opened 2021-05-03
Raw z-app-generated--contactsinteraction--recent URI visible to user


ChristophWurst:
### Describe the bug

Recent contacts show the address but URI instead of the display name.

**To Reproduce**
1. Have recent_contacts enabled
2. Contact someone
3. Open the contacts app
4. Open the entry of this contact
5. See

**Expected behavior**

The contact in an address book "Recently contacted".

**Actual behavior**

![Bildschirmfoto von 2021-05-03 19-07-57](https://user-images.githubusercontent.com/1374172/116908397-234d8000-ac43-11eb-8956-de3ca2cf1c8e.png)

### Server configuration

*The* Nextcloud.

skjnldsv:
You can have a displayname on addressbooks backend :)

ChristophWurst:
That is done in ``\OCA\ContactsInteraction\AddressBook::getProperties``. Is the ``{DAV:}displayname`` not picked up?

skjnldsv:
I'll investigate :mag_right:

pm-pm:
Hello, in my NC21 such a contact is also "read-only".
No chance to delete. A solution would befine. With pleasure a manual workaround. ;-)

ChristophWurst:
That address book is virtual and read-only. Write access is not intended. The app responsible for this lives at https://github.com/nextcloud/server/tree/master/apps/contactsinteraction if you want to have a look yourself.

dcrobertson01:
How do I disable this address book? I now have a contacts list full of people I have no idea who they are or how they got put into my address book, and who I cannot delete, and who are in an addressbook I cannot delete or hide.

dcrobertson01:
I am guessing they are the parents of my son's classmates, and they are in my addressbook now because his teacher does not know how to use BCC.

ChristophWurst:
Disable the `contactsinteraction` app.

pm-pm:
Is there any good solution?

When disabling "contactinteraction app" all Contacts disapeer. :(

What is the root cause and how can it be solved? Some instructions whould be helful.

misterfitz:
I had a similar issue where a close but incorrectly entered email contact was found, I think from an old calendar invite. 

Disabling and re-enabling the `contactinteraction` app worked as @ChristophWurst  suggested here https://github.com/nextcloud/contacts/issues/2187#issuecomment-914258853, thanks!

[...]

PR #3339, opened 2023-04-26
fix(contacts): display name not uri for recent-contacted address book

JohannesGGE:
Fix: #2187 

|before|after|
|--|--|
|Screenshot from 2023-04-26 13-39-04|Screenshot from 2023-04-26 13-39-46|
|Screenshot from 2023-04-26 13-39-17|Screenshot from 2023-04-26 13-40-03|

### DIFF
--- a/src/components/ContactDetails.vue
+++ b/src/components/ContactDetails.vue
@@ -42,7 +42,9 @@
 
 				<!-- fullname -->
 				<template #title>
-					<div v-if="isReadOnly">{{ contact.fullName }}</div>
+					<div v-if="isReadOnly">
+						{{ contact.fullName }}
+					</div>
 					<input v-else
 						id="contact-fullname"
 						ref="fullname"
@@ -59,7 +61,9 @@
 
 				<!-- org, title -->
 				<template #subtitle>
-					<template v-if="isReadOnly">{{ formattedSubtitle }}</template>
+					<template v-if="isReadOnly">
+						{{ formattedSubtitle }}
+					</template>
 					<template v-else>
 						<input id="contact-title"
 							v-model="contact.title"
@@ -226,6 +230,7 @@
 					:is-last-property="true"
 					:property="{}"
 					:hide-actions="true"
+					:is-read-only="isReadOnly"
 					class="property--addressbooks property--last" />
 
 				<!-- Groups always visible -->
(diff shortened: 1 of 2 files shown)


src/components/Properties/PropertySelect.vue
```diff
@@ -91,6 +91,23 @@ export default {
...
+		/**
+		 * Store getters filtered and mapped to usable object
+		 * This is the list of addressbooks that are available to write
+		 *
+		 * @return {{id: string, name: string}[]}
+		 */
+		selectableOptions() {
+			return this.options.
+			filter(option => option.readOnly === undefined || !option.readOnly).
+			map(addressbook => {
```
ChristophWurst, line 103:
```suggestion
			return this.options
				.filter(option => option.readOnly === undefined || !option.readOnly).
				.map(addressbook => {
```

ChristophWurst:
- [ ] BUG: the addressbook prop is no longer read-only. I can change it, but run into an error

## main
Bildschirmfoto vom 2023-04-26 14-02-34

## fix/2187/raw-recent-contacted-uri-visible default state
Bildschirmfoto vom 2023-04-26 14-02-54

## fix/2187/raw-recent-contacted-uri-visible focus state
Bildschirmfoto vom 2023-04-26 14-02-57

JohannesGGE:
@ChristophWurst This seems to be a generel bug with the PropertySelect component because the isReadOnly is not passed, so it's always false.
I fixed it anyway :+1:

src/components/Properties/PropertySelect.vue
```diff
@@ -91,6 +91,23 @@ export default {
...
 	computed: {
+		/**
+		 * Store getters filtered and mapped to usable object
+		 * This is the list of addressbooks that are available to write
+		 *
+		 * @return {{id: string, name: string}[]}
+		 */
+		selectableOptions() {
+			return this.options
+				.filter(option => option.readOnly === undefined || !option.readOnly)
```
st3iny, line 102:
```suggestion
				.filter(option => !option.readOnly)
```

Could be simplified.

src/components/ContactDetails.vue
```diff
@@ -485,17 +490,18 @@ export default {
...
 			return this.addressbooks
-				.filter(addressbook => !addressbook.readOnly && addressbook.enabled)
+				.filter(addressbook => addressbook.enabled)
 				.map(addressbook => {
 					return {
 						id: addressbook.id,
 						name: addressbook.displayName,
+						readOnly: addressbook.readOnly,
 					}
 				})
```
st3iny, line 506: This could have side effects. However, I did not spot any during my testing and the address book options made sense.

st3iny (approved):
Tested and works.

ChristophWurst (approved):

Issue #2925, opened 2022-09-11
in-app search doesn't search in field "notes"


alexanderdd:
### Describe the bug

text from a contacts' "notes" field does not seem to get searched properly by the in-app search

### Steps to reproduce

1. create contact Diana and a few dummy contacts
2. create contact Jim, enter in notes "Diana's dad"
3. (close browser or restart computer or delete cookies or wait a day - not sure - if you just created the contacts it will probably work as expected, but when I use this with contacts from our address book it doesn't work as expected, see also additional info)
4. search for Diana

### Expected behavior

Diana and Jim should be shown in the list, because of the notes in Jim

### Actual behavior

only Diana is shown in the list

### Contact version

 4.2.0

### Operating system

?

### Additional info

* it works as expected in unified search (top right search symbol)
* it works as expected if you first look at Jim's contact details and then do the search (probably there is some kind of cache that gets filled while looking at the contact?)

(it's on managed hosting so I don't know php version etc but I can find out if it is really neccessary)

Nextcloud 23.0.8

alexanderdd:
This is still a bug in NC 25.0.6 contacts 5.2.0

I think this used to work in the past, not sure until which version though...

IMO  it is pretty important to search all fields, not just the name... What do you think?

aurisnoctis:
Can you still reproduce? Others had a similar problem and it seems gone for them: https://github.com/nextcloud/contacts/issues/922

Or could this be only seemingly so due to https://github.com/nextcloud/server/issues/24218? I came here because contacts weren't found with a search for some keyword that's in the "Notes" field. However, I started the search from "All contacts". That's the default name for an automatic group containing all my contacts and that name has a white space by default.

I can indeed reproduce the problem and it shows some strange behavior in the Contacts App of the web GUI:

* I have a certain keyword that's in the title of 1 contacts and in the "Notes" field of 2 other contacts.
* When I search for that keyword, only the one contact is found which has the search phrase in its title.
[...]

PR #3358, opened 2023-05-02
Add NOTES and RELATED to search properties

caplod:
before the search only searches in the following properties:

`'EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME', 'X-MANAGERSNAME', 'TITLE', 'KIND', 'X-ADDRESSBOOKSERVER-KIND'`

so if you search something  in the property 'notes' it is not found.

with this patch, all fields are loaded and searched.

Fixes https://github.com/nextcloud/contacts/issues/2925

### DIFF
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -41,7 +41,7 @@ const isEmpty = value => {
 export const ContactKindProperties = ['KIND', 'X-ADDRESSBOOKSERVER-KIND']
 
 export const MinimalContactProperties = [
-	'EMAIL', 'UID', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME', 'X-MANAGERSNAME', 'TITLE',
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Can't we just include NOTES in the searched properties? I am assuming there is an overhead of fetching all properties to perform the search.

caplod:
notes was just an example.
I have a customer who also uses the field 'Related'  a lot.  
so it would be great to include NOTES and RELATED to search

ChristophWurst:
then let's add the two props to the search properties :+1:

caplod:
Is there anything I can do to get this approved?

ChristophWurst:
@hamza221 could you give this a test? I would like to know if our backend even supports search for NOTES and RELATED. Only selected properties are saved for the oc_cards_properties search index table IIRC.

hamza221 (approved):
Tested search with related and notes

Issue #3351, opened 2023-04-28
Updated avatar only visible after reload/navigating to other contact


JohannesGGE:
### Steps to reproduce

1. Select a contact
2. Click on 'Upload a new picture'/'Choose from Files'
3. Click 'Save'

### Expected behavior

Avatar in contact list as well as in the details view updated.

### Actual behavior

Avatar updated in contact list if there was no avatar before. Otherwise no update.

Avatar didn't update in the details view until another contact is seleted or the page is reloaded.

### PHP engine version

None

### Web server

None

### Database

None

PR #3366, opened 2023-05-03
fix(contacts): display the new avatar on change

JohannesGGE:
Fix: #3351 

When you add, edit, or delete an avatar, it should now update properly.

While working on it I noticed that the avatar doesn't update as well on deletion, even with the fix for a new avatar. Somehow the avatarUrl is not unavailable immediately. I fixed this by setting the avatarUrl undefined on deletion.

### DIFF
--- a/src/components/AppContent/ContactsContent.vue
+++ b/src/components/AppContent/ContactsContent.vue
@@ -63,11 +63,12 @@
 		<template #list>
 			<ContactsList :list="contactsList"
 				:contacts="contacts"
-				:search-query="searchQuery" />
+				:search-query="searchQuery"
+				:reload-bus="reloadBus" />
 		</template>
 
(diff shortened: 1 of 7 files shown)


st3iny (commented):
Tested and works when uploading custom avatars.

Does not work when fetching avatars from social media:
- Avatar in list view will be updated after saving.
- Avatar in details view will be updated after a reload.

Take a look at the following demo:

https://user-images.githubusercontent.com/1479486/236179044-ea7c13c1-d68e-491b-92b8-e48289e42007.mp4

JohannesGGE:
You're right, because I did the cropping before I completely ignored the social media option. I'll fix this as well.

st3iny (approved):
Tested and works. Usage of a bus is not optimal but I see why it is required here.

PR #3388, opened 2023-05-16
Fix: Hide *Delete* action for read-only address books

### LINKED ISSUE #3383: Hide *Delete* action for read-only address books
### Describe the bug

Some address books can't be deleted

Bildschirmfoto vom 2023-05-15 08-52-16
Bildschirmfoto vom 2023-05-15 08-52-20

### Steps to reproduce

1. Run Nextcloud server master
2. Open the contacts app
3. Open the contacts settings
4. Open the actions menu of the Accounts address book

### Expected behavior

No delete action

### Actual behavior

I see a delete action

### Contact version

5.3.0 Beta 2

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

hamza221:
Fixes #3383

### DIFF
--- a/src/components/AppNavigation/Settings/SettingsAddressbook.vue
+++ b/src/components/AppNavigation/Settings/SettingsAddressbook.vue
@@ -94,7 +94,7 @@
 				</template>
 
 				<!-- delete addressbook -->
-				<ActionButton v-if="hasMultipleAddressbooks"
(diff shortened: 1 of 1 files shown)


ChristophWurst (commented):
ABs can be *shared* read-only too. Then the collection should be deletable to allow unsharing.

We have to check if the AB is not shared but ro

Issue #479, opened 2018-02-13
Add new option to hide an address book's contacts without disabling the address book


timreeves:
This issue results from dialog in [Enable/disable address books #412](https://github.com/nextcloud/contacts/pull/412).

The new option already implemented to disable an address book seems to do that on such a low level that it is then disabled per se - for all purposes. For this reason, it is not possible to disable an address book which has been shared with you rather than belonging to you.

In many use cases - team work, small to medium businesses, large contact archives used only seldom - we need a different function: to HIDE the contacts of a particular address book from the VIEW of the logged-in user (available also when not the owner). This is especially true as the historical design of the contacts app is to load all the visible contacts to the front end and then do all the nifty things like selection by category in JS. Which is great with up to (nowadays) about 1000 contacts, but becomes sluggish to unbearable the more you go over the top of that limit.

I would personally prefer that hidden contacts would still be searchable when I type in part of a name or email in the Contacts app search box. And I would certainly want that contacts hidden in the Contacts app (but not disabled) be active for the mail app, e.g. to make sure that mails from contacts known to me are not put into the spam folder.

Finally, in the settings area, if the little "dotdotdot" dropdown for each address book gets a further option "hide/show", then we should really try to explain the difference between Disable and Hide right there where the user has that choice - are tooltips possible there?

Some remote part of me is still dreaming about a backend-based variant of this app - termed by yours truly "Big Contacts" - which would allow working with many thousands of contacts, with the frontend fetching them piece by piece from the backend via traditional lazy loading / infinite scrolling, [see this issue](https://help.nextcloud.com/t/handling-a-large-number-1000-of-contacts-in-the-web-ui/22064).

Be that as it may, the abilty to hide an address book would be a great step on - so thanks John, thanks Henni, thanks Alexander!

mburnicki:
Thanks for the summary, @timreeves . My personal preference would be that searches only be done in the visible address books. Otherwise the search results may be quite confusing.

Concerning the way how to make an address book visible or not, I'd prefer to have a check box beside each name in the address books list rather than have the option hidden in a dotdotdot menu.

It would be useful anyway to see the full list of address books, and a marker (like a checkbox) telling if the contacts from a particular address book are visible in the contacts list, or not.

[...]

Issue #2678, opened 2022-03-16
Make Address Book disable operation responsive


pmarini-nc:
Address Books can be disabled or enabled.

Enabling works just by toggling the enable option
Disabling works only after reloading the whole contacts app page.

It would be better to have disabling acting inmediately (live refresh) as Enabling does.

PR #3399, opened 2023-05-22
enh(addressbooks): add show/hide option for system and shared addressbooks

JohannesGGE:
Fix: #479 #2678

### Changes:
- Show "Show" option for all address books, not just writeable one (#479)
- Load disabled address book contacts to calc sum as shown in the screenshoot
- Refresh contacts on "Show"/Hide (#2678)

### For testing:
- Checkout https://github.com/nextcloud/server/pull/38397
- Show/Hide of "Recently contacted" address book will be tackled in another issue

|before|after|
|--|--|
|image|Screenshot from 2023-05-23 19-41-21|

### DIFF
--- a/src/components/AppNavigation/Settings/SettingsAddressbook.vue
+++ b/src/components/AppNavigation/Settings/SettingsAddressbook.vue
@@ -67,6 +67,21 @@
 					{{ t('contacts', 'Download') }}
 				</ActionLink>
 
+				<template v-if="addressbook.writeProps">
+					<!-- enable/disable addressbook -->
+					<ActionCheckbox v-if="!toggleEnabledLoading"
+						:checked="enabled"
+						@change.stop.prevent="toggleAddressbookEnabled">
+						{{ t('contacts', 'Show') }}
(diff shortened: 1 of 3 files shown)


GretaD (approved):
works

ChristophWurst:
just want to prevent a premature merge. is this PR usable without the change in server?

JohannesGGE:
The show/hide change is just not possible and an error message is shown.

JohannesGGE:
@GretaD Christoph meant the behaviour without the server PR checked out (because there is still a problem).

ChristophWurst:
How about we fetch the address book ACLs and check for the write or property-write ACL. If it's there, show the button. Then this app can work with Nextcloud 28 but also 27, 26 and 25.

JohannesGGE:
The button is now shown depending on the ACLs.
cdav-library PR: https://github.com/nextcloud/cdav-library/pull/789

JohannesGGE:
I think it's fine to have the option to share a hidden address book

GretaD (approved):

PR #3439, opened 2023-06-05
Contacts' import modal improvement

Jerome-Herbinet:
## After 

@nextcloud/designers


jancborchardt (approved):
Looks much better! :)

PR #3441, opened 2023-06-06
More appropriate circle section wording

Jerome-Herbinet:
"user" groups instead of "contact" groups because here are only listed user groups and not contact groups (if I'm not mistaking).

### DIFF
--- a/src/models/constants.ts
+++ b/src/models/constants.ts
@@ -137,7 +137,7 @@ export const CIRCLES_MEMBER_GROUPING = [
 	},
 	{
 		id: `picker-${Type.SHARE_TYPE_GROUP}`,
-		label: t('contacts', 'Contact groups'),
+		label: t('contacts', 'User groups'),
 		share: Type.SHARE_TYPE_GROUP,
 		type: MEMBER_TYPE_GROUP
 	},


jancborchardt (changes requested):
Indeed it does not take "Contact groups" here but just the regular groups from the accounts. It should be simply called "Groups", as we call it like that elsewhere too.

It also needs to be adjusted in the "Add members" modal for the subheading, if you'd like to do that as well @Jerome-Herbinet :) 
image

ArtificialOwl:
> You're right @Jerome-Herbinet. Actually I'm a bit confused since I thought Circles allows you to add both "account groups" and "contact groups" to a circle. The omission of "contact groups" is a bit strange since contacts individually can be added.
> 
> @ArtificialOwl any specific reason here?

It has never been officially requested :-)

Jerome-Herbinet:
Hello @jancborchardt @ArtificialOwl, 
What's the roadmap about this PR and/or the circle enhancement to take in account contact groups ?
Cheers,
Jerome

Jerome-Herbinet:
> Hello @jancborchardt @ArtificialOwl, What's the roadmap about this PR and/or the circle enhancement to take in account contact groups ? Cheers, Jerome

@jancborchardt @ArtificialOwl , any news ?

jancborchardt (commented):
@Jerome-Herbinet since an implementation of being able to add contact groups as well is not currently on the roadmap, I’d suggest to go with the simple wording change for now. And I would say that should simply be "Groups".

> Elsewhere in Nextcloud, when there can be no confusion, we can make do with "Groups" without any problem, but not in this specific case, which I think is an exception.

I think it is better to be consistent – otherwise we don’t have 2 terms, but instead 3.

Jerome-Herbinet:
OK @jancborchardt ; I'll close my PR because @mejo- exactly recently did what you suggest : https://github.com/nextcloud/contacts/pull/4076

jancborchardt:
Ah cool, well thank you both @Jerome-Herbinet @mejo-! :)

PR #3487, opened 2023-06-27
Moved users_picker profile custom picker to contacts

andrey18106:
Moving the [users_picker](https://github.com/nextcloud/users_picker) as it was planned with @DaphneMuller and @julien-nc.

But there is a conflict in NcSelect component ([Vue warn]: You may have an infinite update loop in a component render function.). It works with vue 2.7.14, but not with vue (2.6.14) in contacts, maybe @ChristophWurst can help with it and decide where do we want to move the profile picker.


julien-nc:
Weird thing: `NcSelect` works fine in other places in Contacts even with `vue@2.6.14`.

Also worth mentioning: The profile picker component works fine in https://github.com/nextcloud/users_picker and we can reproduce the problem there when downgrading vue to 2.6.14.

lib/Reference/ProfilePickerReferenceProvider.php
```diff
@@ -0,0 +1,193 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Contacts\Reference;
+
+use OC\Collaboration\Reference\LinkReferenceProvider;
```
ChristophWurst, line 29: private API

lib/Reference/ProfilePickerReferenceProvider.php
```diff
@@ -0,0 +1,193 @@
...
+use OCP\IL10N;
+use OCP\IURLGenerator;
+
+use OCA\Contacts\AppInfo\Application;
+use OCP\Accounts\IAccountManager;
+use OCP\IUserManager;
+
+class ProfilePickerReferenceProvider extends ADiscoverableReferenceProvider {
+
+//	private const RICH_OBJECT_TYPE = Application::APP_ID . '_profile_picker';
```
ChristophWurst, line 43: dead code

lib/Reference/ProfilePickerReferenceProvider.php
```diff
@@ -0,0 +1,193 @@
...
+	 */
+	public function matchReference(string $referenceText): bool {
+		return $this->getObjectId($referenceText) !== null;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function resolveReference(string $referenceText): ?IReference {
+		if ($this->matchReference($referenceText)) {
```
ChristophWurst, line 108: invert the conditions to have an early exit and avoid the many layers of indentation

src/components/ProfilePicker/ProfilePickerReferenceWidget.vue
```diff
@@ -0,0 +1,186 @@
...
+		accessible: {
+			type: Boolean,
+			default: true,
+		},
+	},
+}
+</script>
+
+<style scoped lang="scss">
+.profile_picker-referece {
```
ChristophWurst, line 112: please restructure for BEM

https://scalablecss.com/bem-blocks-within-blocks/

src/components/ProfilePicker/ProfilesCustomPicker.vue
```diff
@@ -0,0 +1,202 @@
...
+import ArrowRightIcon from 'vue-material-design-icons/ArrowRight.vue'
+import UserIcon from './icons/UserIcon.vue'
+
+import NcSelect from '@nextcloud/vue/dist/Components/NcSelect.js'
+import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
+
+import axios from '@nextcloud/axios'
+import { generateOcsUrl, generateUrl } from '@nextcloud/router'
+import debounce from "debounce";
```
ChristophWurst, line 56:
```suggestion
import debounce from 'debounce'
```

src/components/ProfilePicker/ProfilesCustomPicker.vue
```diff
@@ -0,0 +1,202 @@
...
+
+	computed: {
+		options() {
+			if (this.searchQuery !== '') {
+				return this.profiles
+			}
+			return []
+		},
+		noResultText() {
+			return this.loading ? t('contacts', 'Searching...') : t('contacts', 'Not found')
```
ChristophWurst, line 88:
```suggestion
			return this.loading ? t('contacts', 'Searching …') : t('contacts', 'Not found')
```

src/components/ProfilePicker/ProfilesCustomPicker.vue
```diff
@@ -0,0 +1,202 @@
...
+
+	mounted() {
+		this.focusOnInput()
+	},
+
+	methods: {
+		focusOnInput() {
+			setTimeout(() => {
+				this.$refs['profiles-search-input'].$el.getElementsByTagName('input')[0]?.focus()
+			}, 300)
```
ChristophWurst, line 112: https://vuejs.org/api/general.html#nexttick

src/components/ProfilePicker/ProfilesCustomPicker.vue
```diff
@@ -0,0 +1,202 @@
...
+			this.searchQuery = query.trim()
+			this.loading = true
+			await this.debounceFindProfiles(query)
+		},
+		debounceFindProfiles: debounce(function (...args) {
+			this.findProfiles(...args)
+		}, 300),
+		async findProfiles(query) {
+			const url = generateOcsUrl('core/autocomplete/get?search={searchQuery}&itemType=%20&itemId=%20&shareTypes[]=0&limit=20', { searchQuery: query })
+			await axios.get(url).then(res => {
```
ChristophWurst, line 127: you use async-await. replace then-catch with try-catch

src/components/ProfilePicker/ProfilesCustomPicker.vue
```diff
@@ -0,0 +1,202 @@
...
+			})
+				.then((response) => {
+					this.reference = response.data.ocs.data.references[this.resultUrl]
+				})
+				.catch((error) => {
+					console.error(error)
+				})
+				.then(() => {
+					this.loading = false
+				})
```
ChristophWurst, line 162: same here. try-catch-finally will make this code a lot more readable

src/components/ProfilePicker/ProfilesCustomPicker.vue
```diff
@@ -0,0 +1,202 @@
...
+	flex-direction: column;
+	align-items: center;
+	justify-content: space-between;
+	padding: 12px 16px 16px 16px;
+
+	h2 {
+		text-align: center;
+	}
+
+	.footer {
```
ChristophWurst, line 191: BEM

src/components/ProfilePicker/icons/UserIcon.vue
```diff
@@ -0,0 +1,39 @@
...
+		@click="$emit('click', $event)">
+		<svg :fill="fillColor"
+			:width="size"
+			:height="size"
+			enable-background="new 0 0 24 24"
+			version="1.1"
+			viewBox="0 0 24 24"
+			xml:space="preserve"
+			xmlns="http://www.w3.org/2000/svg">
+			<path d="M12,4A4,4 0 0,1 16,8A4,4 0 0,1 12,12A4,4 0 0,1 8,8A4,4 0 0,1 12,4M12,14C16.42,14 20,15.79 20,18V20H4V18C4,15.79 7.58,14 12,14Z" />
```
ChristophWurst, line 16: where's this from? do we need attribution and a license file?

ChristophWurst (changes requested):

DaphneMuller:
> > Moving the [users_picker](https://github.com/nextcloud/users_picker) as it was planned with @DaphneMuller and @julien-nc.
> 
> So what is the plan? Who will maintain that code once it's been moved into this repo?

the contacts maintainers will maintain the code, but we are happy to give the team any code walk-throughs so everybody is comfortable with it.

img/profile.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="#fff"><path d="M12,4A4,4 0 0,1 16,8A4,4 0 0,1 12,12A4,4 0 0,1 8,8A4,4 0 0,1 12,4M12,14C16.42,14 20,15.79 20,18V20H4V18C4,15.79 7.58,14 12,14Z" /></svg>
```
ChristophWurst, line 1: did you create this or do we need license attribution?

src/components/ProfilePicker/ProfilePickerReferenceWidget.vue
```diff
@@ -0,0 +1,164 @@
...
+}
+</script>
+
+<style scoped lang="scss">
+.profile-reference {
+	width: 100%;
+	white-space: normal;
+	display: flex;
+
+	.profile-reference__wrapper {
```
ChristophWurst, line 95:
```suggestion
	&__wrapper {
```

tip: makes it easier to work with BEM in scss

src/components/ProfilePicker/ProfilePickerReferenceWidget.vue
```diff
@@ -0,0 +1,164 @@
+<template>
+	<div class="profile-reference">
+		<div class="profile-reference__wrapper">
+			<div class="profile-card__header">
```
ChristophWurst, line 4:
```suggestion
			<div class="profile-reference__card__header">
```

nit: builds a nice BEM hierarchy

lib/Reference/ProfilePickerReferenceProvider.php
```diff
@@ -0,0 +1,192 @@
...
+
+use OCP\Collaboration\Reference\IReference;
+use OCP\IL10N;
+use OCP\IURLGenerator;
+
+use OCA\Contacts\AppInfo\Application;
+use OCP\Accounts\IAccountManager;
+use OCP\IUserManager;
+
+class ProfilePickerReferenceProvider extends ADiscoverableReferenceProvider {
```
ChristophWurst, line 42: let's add unit tests for this one class and it's logic below :)

ChristophWurst (commented):
Looks good so far!

img/profile.svg
andrey18106: As far as I remember, I took it from Material Design Icons, i dnk what's needed.

ChristophWurst: Licenses like MIT require you to attribute the authors. Find the sources and check the license. If attribution is required we can add a license file like https://github.com/nextcloud/calendar/blob/main/img/LICENSES.md

DaphneMuller:
discussed with Christoph that this issue can be merged after the icon issue is resolved ([https://github.com/nextcloud/contacts/pull/3487#discussion_r1253443738](https://github.com/nextcloud/contacts/pull/3487#discussion_r1253443738?))

will check with @andrey18106

img/profile.svg
andrey18106: @ChristophWurst The icon of profile is from material-design-icons: https://pictogrammers.com/library/mdi/icon/account/

What attribution do we need here? All projects here are under https://pictogrammers.com/docs/general/license/ Pictogrammers Free License, where Icons marked under Apache 2.0.

ChristophWurst: Licenses like MIT and APache 2.0 require you to attribute the authors. We can add a license file like https://github.com/nextcloud/calendar/blob/main/img/LICENSES.md

https://law.stackexchange.com/questions/6048/does-the-apache-2-0-license-require-attribution-on-a-commercial-work-on-which-it

lib/Reference/ProfilePickerReferenceProvider.php
```diff
@@ -0,0 +1,188 @@
...
+
+			// for the Vue reference widget
+			$reference->setRichObject(
+				self::RICH_OBJECT_TYPE,
+				[
+					'user_id' => $userId,
+					'title' => $userDisplayName,
+					'subline' => $userEmail ?? $userDisplayName,
+					'email' => $userEmail,
+					'bio' => isset($bio) && $bio !== '' ? substr_replace($bio, '...', 80, strlen($bio)) : null,
```
ChristophWurst, line 136: I think we should use a multibyte-safe string function, e.g. https://www.php.net/manual/en/function.mb-substr.php and then append `…`. `substr_replace` and `strlen` do not work well with multibytes/emojis.

lib/Reference/ProfilePickerReferenceProvider.php
```diff
@@ -0,0 +1,188 @@
...
+	 * @inheritDoc
+	 */
+	public function resolveReference(string $referenceText): ?IReference {
+		if (!$this->matchReference($referenceText)) {
+			return null;
+		}
+
+		$userId = $this->getObjectId($referenceText);
+		$user = $this->userManager->get($userId);
+		if ($user !== null) {
```
ChristophWurst, line 108: nitpick: if you invert the logic and return early for null users the method has less indentation and becomes easier to read

ChristophWurst (changes requested):
The code looks really good but I notice two critical issues

1) User enumeration restrictions are not valued. You can configure Nextcloud to not allow seeing other people. Yet this picker returns any user objects it can find. Ref https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/file_sharing_configuration.html#file-sharing-configuration. You can test this by restricting your local environment and entering the user ID of another user. That other user should not be selectable. You can check the sharing menu. It values all these restrictions.
2) The user ID is used as search attribute. This works fine for database users or when LDAP records are mapped by a user-readable attribute. However, LDAP/SAML map by UUID by default. So to pick *user a* (LDAP UUID=597ae2f6-16a6-1027-98f4-d28b5365dc14) you have to type the exact string `597ae2f6-16a6-1027-98f4-d28b5365dc14` to find the user. See https://github.com/nextcloud/server/wiki/How-to-test-LDAP%3F for an LDAP-based user base on your dev env.

To address these two issues I would suggest using something like `\OCP\Contacts\IManager::search` instead of `\OCP\IUserManager::get`. It will allow for partial matches (`andre` for `Andrey`) and can help restrict enumeration.

ChristophWurst:
This isn't a light change, so I'll have to allocate resources for the feature.

ChristophWurst:
@andrey18106 I see the security problems have been fixed in your repo. Could you please update this branch? Cheers

DaphneMuller:
Maybe it's better to clarify if we actually want to proceed with this PR given the timeline this is already open?

PR #3505, opened 2023-07-06
deactivate social avatars from twitter

call-me-matt:
I am afraid that with recent changes from twitter, it is not possible to use it as input for our social avatars function :disappointed:


ChristophWurst (changes requested):
Drop the code completely. If it becomes usable again we restore it.

call-me-matt:
True. I removed the twitter part. Facebook is still in as it is working but unreliable.

ChristophWurst (approved):

PR #3527, opened 2023-07-19
fix import loads forever

hamza221:
| Before | After |
|--------|--------|
| [Screencast from 2023-07-19 20-27-24.webm](https://github.com/nextcloud/contacts/assets/40746210/08e26036-82cd-429e-89de-fb3e1025532a) | [Screencast from 2023-07-19 20-25-44.webm](https://github.com/nextcloud/contacts/assets/40746210/a1243391-0793-4106-b307-a7369d2161df) |


GretaD (approved):
workss

Issue #3531, opened 2023-07-31
Improve the quick actions design


GretaD:
### Is your feature request related to a problem? Please describe.

Quick actions are implemented here: https://github.com/nextcloud/contacts/pull/3401. Some design improvement are needed.

- [ ] Improve the horizontal space that the actions take, they should wrap
- [ ] The contact has empty email address fields. This is not common in practice, but still looks weird.
- [ ] Add label for the buttons (its not easy to style it, so need to check with design team for options)
- [ ] Probably some improvement with different screen sizes

### Additional context

ref: https://github.com/nextcloud/contacts/issues/3365

nimishavijay:
From the design review call on 01.08.2023:

**Overall**
* [ ] Placement: Until the shared items is implemented, show the actions in the same column as properties
  - [ ] left aligned with the "Phone", "Email", etc
* [ ] Visible labels not needed for actions, make sure that accessible labels are there, wording can look like:
  - [ ] Email test.test@example.com
  - [ ] Call +4910010010010
  - [ ] View profile

**Mobile**
* [ ] Center the actions in mobile, use flex layout with `justify-items: center;`
* [ ] Use more whitespace below actions 
* [ ] Other small issues in alignment:
  - [ ] Too much space between edit button and 3 dot menu 
  - [ ] Property values are getting ellipsized too fast in view mode, it can be show fully if there is space
  - [ ] Property type has too much space to the right  

**Bugs**
* [ ] If there is an empty email field saved, it should not show in the action menu when clicking on the email button

GretaD:
> [ ] Other small issues in alignment:
> * [ ]  Too much space between edit button and 3 dot menu
> * [ ]  Property values are getting ellipsized too fast in view mode, it can be show fully if there is space
> * [ ]  Property type has too much space to the right

this is not related to this pr. Will open a different issue for those

PR #3532, opened 2023-08-01
Improve quick actions design

GretaD:
fixes: #3531
Screenshot from 2023-08-03 13-18-22

Screenshot from 2023-08-01 13-21-33

### DIFF
--- a/src/components/ContactDetails.vue
+++ b/src/components/ContactDetails.vue
@@ -249,9 +249,13 @@
 						</ActionLink>
 					</Actions>
 					<NcButton v-if="isTalkEnabled && isInSystemAddressBook"
+						:aria-label="(t('contacts', 'Go to talk conversation'))"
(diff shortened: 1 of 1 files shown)


GretaD:
@nimishavijay and @jancborchardt the design part is ready for review, as per screenshot

jancborchardt (changes requested):
On desktop the icons should be left-aligned with the text (name, title, "Phone", etc), currently they are hanging too far left. (As discussed in the call.)

On mobile the actions part looks good – other layout issues as discussed in the call, but different topic.

GretaD:
> On desktop the icons should be left-aligned with the text (name, title, "Phone", etc), currently they are hanging too far left. (As discussed in the call.)

sorry, then i misunderstood the left part. See the new version, this is what you meant?
Screenshot from 2023-08-03 13-18-22

nimishavijay (approved):

PR #3579, opened 2023-08-22
Made Circles info always display

### LINKED ISSUE #3543: Not obvious what Circles are if emptycontent message is not shown
If you do not have Circles yet, there is a nice emptycontent message. But it does not show anymore if some Circles already exist – which likely is the case if you are in a company.

Could we keep showing the message even if Circles are already in the list? And a slight wording enhancement from "users" to "people":
image

auroraanna:
Fixes #3543

### DIFF
--- a/src/components/AppNavigation/RootNavigation.vue
+++ b/src/components/AppNavigation/RootNavigation.vue
@@ -161,8 +161,7 @@
 						@click="onToggleCircles" />
 				</template>
 
-				<p v-else-if="!loading"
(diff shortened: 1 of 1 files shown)


auroraanna:
I think the NcEmptyContent isn't useful for this because in this case, we just want a paragraph, not an icon and a button too.

jancborchardt (approved):
Great, thanks a lot! :) And yep agree, EmptyContent component would be too much there.

PR #3650, opened 2023-10-09
fix v-model of address book select control

### LINKED ISSUE #3611: Selection dialogue for address book import always jumps to "Contacts"
### Describe the bug

I have more than one address book and would like to import a larger number of contacts from a file. Unfortunately, the selection dialogue always jumps to "Contacts" (and then imports them there) and does not let me select another address book.

### Steps to reproduce

1. Go to the contacts settings
2. Click on import button
3. Try to change the address book -> only "Contacts" is selected

### Expected behavior

Another address book should be selectable.

### Actual behavior

Multiple address books are displayed, but regardless of what is selected, it always changes to "Contacts".

### Contact version

at least 5.3.2 and 5.4 affected 

### Operating system

Installed on Archlinux, tested with Firefox on Linux and Android as well as Edge on Windows

### PHP engine version

PHP 8.1

### Web server

Nginx

### Database

MariaDB

### Additional info

_No response_

fuse314:
fixes #3611

### DIFF
--- a/src/components/AppNavigation/Settings/SettingsImportContacts.vue
+++ b/src/components/AppNavigation/Settings/SettingsImportContacts.vue
@@ -38,7 +38,7 @@
 					<h2>{{ t('contacts', 'Import contacts') }}</h2>
 					<NcSelect v-if="!isSingleAddressbook"
 						id="select-addressbook"
-						v-model="selectedAddressbook"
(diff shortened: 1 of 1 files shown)


st3iny (approved):
Tested and works.

Thank you very much :)

Issue #3651, opened 2023-10-10
Tel property with type Voice don't work 


kesselb:
### Describe the bug

The property parameter TYPE=VOICE for the tel property don't work properly.

The data is properly stored in the vcard, but displayed incorrectly.

### Steps to reproduce

1. Create a new contact Alice
2. Add phone number (e.g. +49 030 11223344) with type voice
3. Save the contact
4. Reload the page
5. The type is home

### Expected behavior

The shown type is voice

### Actual behavior

The shown type is home

### Contact version

main

### Additional info

[Screencast from 2023-10-10 19-35-11.webm](https://github.com/nextcloud/contacts/assets/3902676/8ed3c04f-ef7b-4339-baa6-da6494df905b)

https://github.com/nextcloud/contacts/blob/9854dd89cc575ebd23206831ceed3e31c36038d3/src/models/rfcProps.js#L223-L239

https://github.com/nextcloud/contacts/blob/9854dd89cc575ebd23206831ceed3e31c36038d3/src/components/ContactDetails/ContactDetailsProperty.vue#L243-L268

The above code will match the selected types to the types from rfc props to use the right translation. 
this.propModel.options => the options from rfc props. 

Example for type Voice:

| Type | Score |
|--------|--------|
| HOME,VOICE | 1 |
| HOME | 0 |
| WORK,VOICE | 1 |
| WORK | 0 |
| CELL | 0 | 
| CELL,VOICE | 1 |
| WORK,CELL | 0 |
| HOME,CELL | 0 |
|FAX | 0 |
| VOICE | 1 | 

Subsequently, the list is ordered by score, items with score 0 removed and the first item is used as type.

Possible solutions:

1) Reorder rfcProps.js: https://github.com/nextcloud/contacts/pull/3652
2) Extend score calculation to assert that selectedType and type.id.split() have the same length: https://github.com/nextcloud/contacts/pull/3653

PR #3653, opened 2023-10-10
fix: improve matching for tel type parameter

kesselb:
Fix https://github.com/nextcloud/contacts/issues/3651, Close https://github.com/nextcloud/contacts/pull/3652

The old implementation assigns the same score for HOME,VOICE and VOICE for a tel property with parameter voice.

This pull requests adds an aditional point for the items with the same length.

| This PR | Main |
|--------|--------|
| Screenshot from 2023-10-10 20-16-56 | Screenshot from 2023-10-10 20-18-12 |


st3iny (approved):
Tested and works. The additional score makes sense.

PR #3656, opened 2023-10-11
refactor: move type matching to own function

kesselb:
Follow-up for #3653 

- Move matching to own function
- Tests
- Additional check for the bonus point if also the intersection has the same length

### DIFF
--- a/tests/javascript/utils/matchTypes.test.ts
+++ b/tests/javascript/utils/matchTypes.test.ts
@@ -1,0 +1,133 @@
+/**
+ * @copyright Copyright (c) 2023 Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * @author Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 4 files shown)


tests/javascript/utils/matchTypes.test.ts
```diff
@@ -0,0 +1,133 @@
...
+ *
+ */
+
+import { matchTypes } from '../../../src/utils/matchTypes'
+import rfcProps from '../../../src/models/rfcProps.js'
+
+describe('utils/matchTypes test suite', () => {
+
+	describe('impp', () => {
+		it('match', () => {
```
ChristophWurst, line 29:
```suggestion
		it('matches', () => {
```

if you want your test cases to read like English sentences :)

tests/javascript/utils/matchTypes.test.ts
```diff
@@ -0,0 +1,133 @@
...
+				rfcProps.properties.impp.options,
+			)
+
+			expect(match).toMatchObject({
+				type: { id: 'XMPP', name: 'XMPP' },
+				score: 2,
+			})
+		})
+
+		it('no match', () => {
```
ChristophWurst, line 43:
```suggestion
		it('does not match', () => {
```

ChristophWurst (approved):

st3iny (approved):
Fancy typescript code :sparkles:

PR #3669, opened 2023-10-18
Import file FileAction 28 compatibility

### LINKED ISSUE #3621: Import file FileAction 28 compatibility
There is a new API for 28, the import VCF file action needs a bump:
https://nextcloud-libraries.github.io/nextcloud-files/functions/registerFileAction.html

GretaD:
fixes #3621 

- [x] Add initi.ts file
- [x] addInitScript on LoadContactsFilesActions.php
- [x] Migrate files-action.js to the new api
- [x] Check documentation how to migrate OCA.Files.fileActions.setDefault(mime, name)


src/files-action.js
```diff
@@ -37,8 +36,8 @@ window.addEventListener('DOMContentLoaded', () => {
 				window.location = generateUrl(`/apps/contacts/import?file=${absPath}`)
 			},
 		})
+	// to be changed
 		OCA.Files.fileActions.setDefault(mime, name)
```
GretaD, line 40: @skjnldsv  can you please help me with this part and how to adapt i correctly. The doc says: _" ``OCA.Files``: Everything removed but Sidebar and Settings. See the Added API section for replacements."_

But Im not sure how to do it.

skjnldsv: Read the docs on the requirements and available properties: https://nextcloud-libraries.github.io/nextcloud-files/classes/FileAction.html

ChristophWurst: @GretaD https://nextcloud-libraries.github.io/nextcloud-files/classes/FileAction.html#default. Pass it as `default: 'default',` in the file action object above.

src/files-action.js
```diff
@@ -42,3 +44,21 @@ window.addEventListener('DOMContentLoaded', () => {
...
+
+if (nextcloudVersionIs28) {
+	registerFileAction(new FileAction({
+		name,
+		displayName() {
+			return t('contacts', 'Import')
+		},
+		default: 'default',
+		mime,
+		permissions: OC.PERMISSION_READ,
```
ChristophWurst, line 56: I'd prefer if this came from `@nextcloud/files` and not a global variable of server

src/init.d.ts
```diff
@@ -0,0 +1 @@
+export {};
```
GretaD, line 1: this was auto generated, so i will remove it

lib/Listener/LoadContactsFilesActions.php
```diff
@@ -39,5 +39,10 @@ public function handle(Event $event): void {
 		}
 
 		Util::addScript(Application::APP_ID, 'contacts-files-action');
+		// only available since Nc 28
+		if (method_exists(Util::class, 'addInitScript')) {
+			Util::addInitScript(Application::APP_ID, 'init');
+		}
```
st3iny, line 45:
```suggestion
		// only available since Nc 28
		if (method_exists(Util::class, 'addInitScript')) {
			Util::addInitScript(Application::APP_ID, 'contacts-files-action');
		} else {
			Util::addScript(Application::APP_ID, 'contacts-files-action');
		}
```

st3iny: This way you can even delete of init.ts as the existing bundle will be loaded in both cases. The conditional logic is already present in `files-action.js`.

src/files-action.js
```diff
@@ -19,26 +19,51 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
+/* eslint-disable */
```
ChristophWurst, line 22: that's not a good idea. this will just silence all eslint warnings.

let's address them instead.

`@nextcloud/files` is not a dependency of this app. add it.

ChristophWurst (changes requested):

src/files-action.js
GretaD: i was actually trying to dismiss the "import' is only available in ES6", but that was fixed by adding the .jshintrc file. I also added the dependencies that i missed adding 10 times.

.jshintrc
```diff
@@ -0,0 +1,4 @@
+{
```
ChristophWurst, line 1: Contacts doesn't use jshint as far as I can tell. Does this file have any effect?

GretaD: I read the Phpstorm documentation on how to fix this, and it was proposed you either add .jshintrc file, or just _"esversion": 6_, on the package.json. Almost every line of code will show _xxx is available in ES6 (use 'esversion: 6')_ if i dont have this file.

ChristophWurst: What was the problem?

st3iny: You could disable jshint in PHPStorm. I did so too because it was constantly annoying me.

grafik

st3iny (commented):
I tested it and it works no master and stable27. Please address the feedback regarding jshint.

st3iny (approved):
Tested and works. I fixed some minor bugs and merged both if statements.

Please squash all commits and we are ready to merge :)

lib/Listener/LoadContactsFilesActions.php
```diff
@@ -38,6 +38,11 @@ public function handle(Event $event): void {
 			return;
 		}
 
-		Util::addScript(Application::APP_ID, 'contacts-files-action');
+		// only available since Nc 28
+		if (method_exists(Util::class, 'addInitScript')) {
+			Util::addInitScript(Application::APP_ID, 'contacts-files-action');
+		} else {
+			Util::addScript(Application::APP_ID, 'contacts-files-action');
```
skjnldsv, line 45:
```suggestion
			Util::addScript(Application::APP_ID, 'contacts-files-action', 'files');
```

src/files-action.js
```diff
@@ -20,25 +21,47 @@
...
-				const absPath = `${context.dir === '/' ? '' : context.dir}/${fileName}`
-				window.location = generateUrl(`/apps/contacts/import?file=${absPath}`)
-			},
-		})
-		OCA.Files.fileActions.setDefault(mime, name)
-		return
-	}
-	console.error('Unable to register vcf import action')
-})
+if (nextcloudVersionIsGreaterThanOr28) {
```
skjnldsv, line 33: I am not even sure this is needed. Worse case this will register an unused action :shrug:

src/files-action.js
```diff
@@ -20,25 +21,47 @@
...
+	registerFileAction(new FileAction({
+		id: name,
+		displayName: () => t('contacts', 'Import'),
+		default: DefaultType.DEFAULT,
+		mime,
```
skjnldsv, line 38: FileAction doesn't have a `mime`

src/files-action.js
```diff
@@ -20,25 +21,47 @@
...
+		enabled: (nodes) => {
+			return nodes.every((node) => node.mime === mime && (node.permissions & Permission.READ))
```
skjnldsv, line 40: Using `some` with the opposite check is faster as it fill stop at the first false return
Something like 
``` suggestion
			return nodes.some((node) => node.mime !== mime || (node.permissions & Permission.READ) === 0)
```

skjnldsv:
Added some comments @GretaD @st3iny

PR #3670, opened 2023-10-18
Add shared items and events for each contact

### LINKED ISSUE #3364: Add shared items and events for each contact

Currently the information shown for each contact is just the info in the vCard and nothing else. Now that system users are shown in contacts with https://github.com/nextcloud/server/issues/19575 there is the possibility to show additional info for these contacts, such as files or photos shared between the person and the contact, and more.

Google Contacts also has a similar idea with the "You and Person" section

image

For the contacts which are in the system address book, if there are some existing items shared with the person and that contact they can be displayed in view mode. For example, shared files and photos, and shared events. More possibilities are recently sent emails, Talk conversations which are common, etc.

image

In mobile it can be shown below all the vCard info.

n/a

### Additional context

n/a

GretaD:
fixes #3364 

- [x] Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/4662
- [x] Requires https://github.com/nextcloud/related_resources/pull/302 and https://github.com/nextcloud/related_resources/pull/303
- [x] documentation https://github.com/nextcloud/documentation/pull/11270
- [x] required https://github.com/nextcloud/contacts/pull/3695

### DIFF
--- a/src/components/ContactDetails.vue
+++ b/src/components/ContactDetails.vue
@@ -309,13 +309,61 @@
 					class="property--groups property--last"
 					@update:value="updateGroups" />
 			</div>
+			<div v-if="nextcloudVersionAtLeast28" class="related-resources">
+				<NcRelatedResourcesPanel v-if="!filesPanelHasError"
+					provider-id="account"
+					resource-type="files"
+					:description="desc"
+					:limit="5"
+					:header="t('contacts', 'Media shares with you')"
+					:item-id="contact.uid"
+					:primary="true"
+					@has-resources="value => hasRelatedResources = value"
+					@has-error="value => filesPanelHasError = value" />
+				<NcRelatedResourcesPanel v-if="!talkPanelHasError"
+					provider-id="account"
+					resource-type="talk"
+					:description="desc"
+					:limit="5"
+					:header="t('contacts', 'Talk conversations with you')"
+					:item-id="contact.uid"
+					:primary="true"
+					@has-resources="value => hasRelatedResources = value"
+					@has-error="value => talkPanelHasError = value" />
+				<NcRelatedResourcesPanel v-if="!calendarPanelHasError"
+					provider-id="account"
+					resource-type="calendar"
+					:description="desc"
+					:limit="5"
+					:header="t('contacts', 'Calendar events with you')"
+					:item-id="contact.uid"
+					:primary="true"
+					@has-resources="value => hasRelatedResources = value"
+					@has-error="value => calendarPanelHasError = value" />
+				<NcRelatedResourcesPanel v-if="!deckPanelHasError"
+					provider-id="account"
+					resource-type="deck"
+					:description="desc"
+					:limit="5"
+					:header="t('contacts', 'Deck cards with you')"
+					:item-id="contact.uid"
+					:primary="true"
+					@has-resources="value => hasRelatedResources = value"
+					@has-error="value => deckPanelHasError = value" />
+				<NcEmptyContent v-if="!hasRelatedResources"
+					:name="t('contacts', 'No shared items with this contact')">
+					<template #icon>
+						<FolderMultipleImage :size="20" />
+					</template>
+				</NcEmptyContent>
+			</div>
 			<!-- new property select -->
 			<AddNewProp v-if="!isReadOnly"
 				:bus="bus"
 				:contact="contact" />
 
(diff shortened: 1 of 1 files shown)


GretaD:
@nimishavijay this is how it looks like on contacts. We used the same component that is used on Talk and other apps, so the view its exactly the same. Unfortunately, we can split with categories like Files, Calendar and such, but instead, all the shared resources will be listed here.
Are we supposed to show this one mobile? I guess its only web. 
Screenshot from 2023-10-25 21-16-04

GretaD:
Big screen view. Looks like this. I will make the spaces between app shares smaller. But otherwise, will look like this.
Screenshot from 2023-10-27 15-57-18
Im not sure if we should show them on small screen, but this is the only acceptable view i can think of. I would say on small screen, not to show related resources with the argument that we already have quick actions and thats enough, but this is not my decision to make.

Screenshot from 2023-10-27 15-57-38
@nimishavijay now you can review and let me know what you think can be improved :)

ChristophWurst:
The *Anything shared with the same group of people …* is strange for events shared with *this* contact IMO. If this is coming from nc/vue, we might want to add an option to hide or replace the text.

nimishavijay:
Feedback from the rest of the team:
- [ ] "Recently shared items" --> "Media shared with you"
- [ ] "Talk conversations" --> "Talk conversations together"

cc @jancborchardt :) 

@GretaD it would be great if the previews could make it in as they are a pretty important part of this feature. If there is something much more urgent on your todo list then all good, but otherwise the previews would be great to have :)

src/components/ContactDetails.vue
```diff
@@ -307,13 +307,53 @@
 					class="property--groups property--last"
 					@update:value="updateGroups" />
 			</div>
+			<div v-if="nextcloudVersion" class="related-resources">
```
st3iny, line 310:
```suggestion
			<div v-if="nextcloudVersionAtLeast28" class="related-resources">
```

src/components/ContactDetails.vue
```diff
@@ -676,6 +727,9 @@ export default {
 		isInSystemAddressBook() {
 			return this.contact.addressbook.id === 'z-server-generated--system'
 		},
+		nextcloudVersion() {
```
st3iny, line 730:
```suggestion
		nextcloudVersionAtLeast28() {
```

st3iny (commented):
Tested and works so far.

### There are some design issues:

- [ ] The related resource panel does not wrap to the right column. It is rendered below the contacts form.
- [ ] Panels with errors should be hidden. They emit an event `@has-error` which could be leveraged to achieve this. The state of each panel could be saved to a boolean and then the panel could be hidden via `v-if` if the error is true for a given panel.

GretaD:
> Feedback from the rest of the team:
> 
> * [ ]  "Recently shared items" --> "Media shared with you"
> * [ ]  "Talk conversations" --> "Talk conversations together"
> 
There are 2 more apps, calendar and decks. What header do you propose there? Right now is Calendar events and Deck cards.
> cc @jancborchardt :)
> 
> @GretaD it would be great if the previews could make it in as they are a pretty important part of this feature. If there is something much more urgent on your todo list then all good, but otherwise the previews would be great to have :)

If the related resource changes to support preview and the changes we need on the nc/vue are merged and tested today, i can try and make it work for tomorrow(deadline), but this needs some work in different apps and libraries and we might not have the time to do it perfectly and not easily breakable design wise.

jancborchardt:
@GretaD @nimishavijay maybe to simplify, we can go with "[…] with you" for all headings? So:
- Media shares with you
- Talk conversations with you
- Calendar events with you
- Deck cards with you

And no subline. Then it's clear it's things you have in common, in clear language. (Otherwise it sounded like you can see their shares.)

GretaD:
> Tested and works so far.
> 
> ### There are some design issues:
> * [ ]  The related resource panel does not wrap to the right column. It is rendered below the contacts form.
I think this was caused by the error message, because it was too long and not wrapped. Now we hide it when the app is not activated, so the screen will break when the screen is smaller than 1600px wide.

st3iny (commented):
Tested and works. My previous review was addressed.

Unfortunately, we can only merge this after updating nc-vue to 8 :(

ChristophWurst (approved):

PR #3681, opened 2023-10-31
fix(socialavatars): Fix HTTP client usage

ChristophWurst:


call-me-matt (approved):
cannot test at the moment, but code looks fine to me.

PR #3682, opened 2023-10-31
fix(socialavatar): Match base class argument name

ChristophWurst:


hamza221:
Backport?

ChristophWurst:
I would skip because PHP does not throw an error, it's merely Psalm that complained.

PR #3690, opened 2023-11-03
Fix the permission check on files-actions

GretaD:
ref https://github.com/nextcloud/contacts/pull/3669#pullrequestreview-1712054589

### DIFF
--- a/src/files-action.js
+++ b/src/files-action.js
@@ -35,9 +35,12 @@ if (nextcloudVersionIsGreaterThanOr28) {
 		id: name,
 		displayName: () => t('contacts', 'Import'),
 		default: DefaultType.DEFAULT,
-		mime,
 		enabled: (nodes) => {
-			return nodes.every((node) => node.mime === mime && (node.permissions & Permission.READ))
+			if (nodes.length !== 1) {
+				return false
+			}
+			const node = nodes[0]
+			return node.mime === mime && (node.permissions & Permission.READ)
 		},
 		iconSvgInline: () => ContactSvg,
 		async exec(file) {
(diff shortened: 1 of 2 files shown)


src/files-action.js
```diff
@@ -35,9 +35,8 @@ if (nextcloudVersionIsGreaterThanOr28) {
 		id: name,
 		displayName: () => t('contacts', 'Import'),
 		default: DefaultType.DEFAULT,
-		mime,
 		enabled: (nodes) => {
-			return nodes.every((node) => node.mime === mime && (node.permissions & Permission.READ))
+			return nodes.some((node) => node.mime !== mime || (node.permissions & Permission.READ) === 0)
```
ChristophWurst, line 39: this can't be right. this evaluates to true if there is at least one node with a mismatching mime or no read permission.

this needs a negation, no?

skjnldsv: right, negation up front
```suggestion
			return !nodes.some((node) => node.mime !== mime || (node.permissions & Permission.READ) === 0)
```

skjnldsv: Wait, counter argument, considering you do not support `execBatch`, you don't even need to do that.

```suggestion
		if (nodes.length !== 1) {
			return false
		}

		const node = nodes[0]
		return node.mime === mime && (nodes.permissions & Permission.UPDATE) !== 0
```

src/files-action.js
```diff
@@ -35,9 +35,12 @@ if (nextcloudVersionIsGreaterThanOr28) {
...
 		displayName: () => t('contacts', 'Import'),
 		default: DefaultType.DEFAULT,
-		mime,
 		enabled: (nodes) => {
-			return nodes.every((node) => node.mime === mime && (node.permissions & Permission.READ))
+			if (nodes.length !== 1) {
+				return false
+			}
+			const node = nodes[0]
+			return node.mime === mime && (node.permissions & Permission.UPDATE) !== 0
```
ChristophWurst, line 43: read permissions should be sufficient. import doesn't write to the file

PR #3695, opened 2023-11-08
Upgrade to Nextcloud vue 8.0.0

GretaD:
Breaking changes:

- Migrate from deprecated title prop by @raimund-schluessler in #4052 - **done on this pr**
- Remove deprecated popovermenu component by @raimund-schluessler in #4081 - **done before this pr**
- Remove excludeClickOutsideClasses from clickOutsideOptions by @raimund-schluessler in #4088 -  **we dont use it**
- Remove deprecated NcAppNavigationCounter component by @raimund-schluessler in #4096 - **dont use it/done before**
- Rename title props and slots to name by @raimund-schluessler in #4106 - **done on this pr**
- Remove deprecated NcMultiselect and NcMultiselectTags components by @susnux in #4281 - **done in another pr** 
- chore: Drop install entry point and replace it with an Vue Plugin by @susnux in #4349 -**not applicable on contacts(?)**
- feat(NcEmptyContent)!: make empty content centered by default instead of 20vh margin by @ShGKme in #4506
- chore!: Rename Datetime to DateTime in component names by @susnux in #4560 - **done in this pr**
- feat(NcIconSvgWrapper)!: remove ID from svg by @ShGKme in #4607 - **not used on contact**

- [x] remove https://github.com/vuejs/composition-api as its not maintained anymore
- [x] update "vue-template-compiler"" and "vue" to 2.7

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@nextcloud/paths": "^2.1.0",
     "@nextcloud/router": "^2.2.0",
     "@nextcloud/sharing": "^0.1.0",
-    "@nextcloud/vue": "^7.12.6",
+    "@nextcloud/vue": "^8.0.0",
     "b64-to-blob": "^1.2.19",
     "camelcase": "^8.0.0",
     "d3": "^7.8.5",
@@ -65,7 +65,7 @@
     "qr-image": "^3.2.0",
     "string-natural-compare": "^3.0.1",
     "uuid": "^9.0.0",
-    "vue": "~2.6.14",
+    "vue": "~2.7.15",
     "vue-click-outside": "^1.1.0",
     "vue-cropperjs": "^4.2.0",
     "vue-material-design-icons": "^5.2.0",
@@ -89,7 +89,6 @@
     "@nextcloud/typings": "^1.7.0",
     "@nextcloud/webpack-vue-config": "^5.2.0",
     "@types/jest": "^29.5.3",
-    "@vue/composition-api": "^1.7.2",
     "@vue/vue2-jest": "^29.2.5",
     "babel-jest": "^29.6.2",
     "eslint-import-resolver-typescript": "^3.6.0",
@@ -101,7 +100,7 @@
     "ts-jest": "^29.1.1",
     "ts-loader": "^9.5.0",
     "typescript": "^5.1.6",
-    "vue-template-compiler": "~2.6"
+    "vue-template-compiler": "~2.7"
   },
   "browserslist": [
     "extends @nextcloud/browserslist-config"
--- a/src/components/AppNavigation/Settings/SettingsAddressbook.vue
+++ b/src/components/AppNavigation/Settings/SettingsAddressbook.vue
@@ -44,7 +44,7 @@
 			<Button v-if="!addressbook.readOnly"
 				v-tooltip.top="sharedWithTooltip"
 				:class="{'addressbook__share--shared': hasShares}"
-				:title="sharedWithTooltip"
+				:name="sharedWithTooltip"
 				href="#"
 				class="addressbook__share"
 				@click="toggleShare">
--- a/src/components/ContactsList.vue
+++ b/src/components/ContactsList.vue
@@ -180,7 +180,8 @@ export default {
 
 // Search field
 .search-contacts-field {
-	padding: 5px 10px 5px 40px;
+	padding: 5px 10px 5px 50px;
+	margin-top: 4px;
 
 	> input {
 		width: 100%;
--- a/src/components/OrgChart.vue
+++ b/src/components/OrgChart.vue
@@ -4,14 +4,14 @@
 			<h3>
 				{{ t('contacts', 'Chart') }}:
 			</h3>
-			<Multiselect v-model="chart"
+			<NcSelect v-model="chart"
 				class="chart-selection"
 				:disabled="data.length === 1"
 				:options="charts"
 				:allow-empty="false"
 				:searchable="false"
 				:placeholder="placeholder"
-				track-by="id"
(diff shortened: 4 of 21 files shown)


ChristophWurst (approved):

ChristophWurst:
Any incompatibilities with 27 or older?

src/views/Processing/ImportView.vue
```diff
@@ -25,7 +25,7 @@
 		<ProcessingScreen :progress="progress"
 			:total="total"
 			:desc="failed > 0 ? importFailed : ''"
-			:title="total === progress
```
ShGKme, line 28: `ProcessingScreen` is not `@nextcloud/vue` component, its prop was not renamed

susnux:
> Any incompatibilities with 27 or older?

If you use the right sidebar then it will be inconsistent from UI

ShGKme:
> * feat(NcEmptyContent)!: make empty content centered by default instead of 20vh margin by @ShGKme in #4506

This requires either adding `height: 100%` to the `<NcEmptyContent>` or using it in a stretched `flex` container.

ShGKme:
Sorry, suddenly touched a button when the cursor was on the "close" button 🙈

GretaD:
> > Any incompatibilities with 27 or older?
> 
> If you use the right sidebar then it will be inconsistent from UI

we dont :)

miaulalala (commented):
The jumping contacts issue is not resolved yet but I am unsure where that originates. Can't inspect the app with Vue Dev Tools either.

GretaD:
> The jumping contacts issue is not resolved yet but I am unsure where that originates. Can't inspect the app with Vue Dev Tools either.

i cannot reproduce it. if someone else can test and see if they can

package.json
```diff
@@ -65,7 +65,7 @@
     "qr-image": "^3.2.0",
     "string-natural-compare": "^3.0.1",
     "uuid": "^9.0.0",
-    "vue": "~2.6.14",
+    "vue": "~3.3.7",
```
GretaD, line 68: i need  "vue": "~2.7.15", not 3.3 here

ChristophWurst: double check if 2.7 works. we had to downgrade in the past because it did not work with the virtual scroller component used for the contacts list

src/components/ContactsList.vue
```diff
@@ -180,7 +180,8 @@ export default {
 
 // Search field
 .search-contacts-field {
-	padding: 5px 10px 5px 40px;
+	padding: 5px 10px 5px 50px;
+  margin-top: 4px;
```
susnux, line 184:
```suggestion
  	margin-top: 4px;
```

src/components/OrgChart.vue
```diff
@@ -4,7 +4,7 @@
 			<h3>
 				{{ t('contacts', 'Chart') }}:
 			</h3>
-			<Multiselect v-model="chart"
+			<NcSelect v-model="chart"
 				class="chart-selection"
 				:disabled="data.length === 1"
 				:options="charts"
```
susnux, line 10: `track-by="id"` does not exist on NcSelect

package.json
ShGKme: We use `vue-virtual-scroller` in Talk with Vue 2.7, it should work.

We also tried to use `vue-virtual-scroll-list` and it worked in 2.7 if I remember correctly.

If it doesn't work, a minor upgrade may help, I guess.

ChristophWurst: It doesn't work for this app https://github.com/nextcloud/contacts/pull/3117

ChristophWurst: > If it doesn't work, a minor upgrade may help, I guess.

Both packages are at their latest version already

ChristophWurst: > TypeError: this.$route is undefined

is the specific error we get in `ContactsListItem.vue` wrapped with `vue-virtual-scroll-list`.

ChristophWurst: If I refactor the code to props taking the $route info from the parent component, I get a `Unknown custom element: <router-link> - did you register the component correctly? For recursive components, make sure to provide the "name" option.`. So my suspicion is that the virtual scroll list doesn't register the vue router on its items.

raimund-schluessler: Maybe that is actually a problem of `NcListItem` which is used in the scroller and can be a `router-link` but does not register it.

ChristophWurst: Fair point. In my understanding all components are known to the Vue instance. If we construct the instance with the router, router-link should be a known component everywhere. :thinking: https://v3.router.vuejs.org/api/#component-injected-properties

ShGKme: For the record.

The problem was that in the app Webpack resolved imported Vue as `vue.esm.js` while from `vue-virtual-scroll-list` as `vue.common.js`.

So there were 2 different copies of Vue in the bundle, and Vue-Router was installed only in one copy.

Dirty fix: 
```js
webpackConfig.resolve.alias = {
	vue$: 'vue/dist/vue.esm.js',
}
```

src/components/AppNavigation/Settings/SettingsAddressbook.vue
```diff
@@ -26,7 +26,7 @@
 		<li :class="{'addressbook--disabled': !addressbook.enabled}" class="addressbook">
 			<div class="addressbook__content">
 				<!-- addressbook name -->
-				<span class="addressbook__name" :title="addressbook.displayName">
+				<span class="addressbook__name" :name="addressbook.displayName">
```
hamza221, line 29:
```suggestion
				<span class="addressbook__name" :title="addressbook.displayName">
```

hamza221 (approved):

ShGKme (approved):

PR #3715, opened 2023-11-23
feat(deps): Add Nextcloud 29 support

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -39,7 +39,7 @@
 
 	<dependencies>
 		<php min-version="7.4" max-version="8.3" />
-		<nextcloud min-version="25" max-version="28" />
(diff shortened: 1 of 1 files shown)


hamza221:
We're on rc1  https://github.com/nextcloud/contacts/pull/3714

ChristophWurst (approved):
I would not drop 25 in here right before branching off. Let's drop 25 for the next minor of Contacts only

ChristophWurst (changes requested):
wrong radio

ChristophWurst:
all previous pre-releases supported 25 so I would keep this until the final

ChristophWurst (approved):

PR #3748, opened 2023-12-19
release v5.4.3

hamza221:


ChristophWurst (approved):
Let's do this as the final release of the 5.4 line

5.5 seems to fully replace 5.4. There are no Nextcloud versions that only work with 5.4 and not with 5.5: https://apps.nextcloud.com/apps/contacts

Issue #3720, opened 2023-11-26
Double HTML encoding in compound representation of contact's Title and Company


hatelamers:
### Describe the bug

When both title and company fields of a contact are filled and one of them or both contain special characters like **ampersand**, **apostroph**, **quote** or **less-than** or **greater-then** they are double encoded in readonly view.

**Edit mode (both fields)**
![title+company-edit-mode](https://github.com/nextcloud/contacts/assets/97636078/8d95861c-f745-4e9e-b6c9-f84c98cc6ae3)

**Read mode (compound field)**
![title+company-read-mode](https://github.com/nextcloud/contacts/assets/97636078/57ee005d-9a36-4427-b81e-30f02cc18be7)

**Read mode (company only)**
![company-only-read-mode](https://github.com/nextcloud/contacts/assets/97636078/7d143771-717e-44bc-9c65-1272f9fc8ba9)

### Steps to reproduce

1. Open Contacts
2. Select "New Contact" or open an existing one for editing
3. Enter some text containing one or more characters mentioned above in Title and Company fields
4. Save the card
5. In read mode all "special" characters are double encoded

### Expected behavior

Scecial characters rendered as such

### Actual behavior

Special characters rendered as HTML entities

### Contact version

5.4.2

### Operating system

Linux dd46126 5.15.0-88-generic #98-Ubuntu SMP Mon Oct 2 15:18:56 UTC 2023 x86_64 

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MariaDB

PR #3755, opened 2023-12-23
Fix/Contact subtitle encoding

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/3720


GretaD (approved):
works :)

GretaD:
do we need a backport here?

PR #3782, opened 2024-01-12
don't filter results while sharing

avinash-0007:
Share only works by display name right now

- Search by email to share calendar is broken completely as there is a filterBy option in [vue-select](https://vue-select.org/api/props.html#filterby) whose return value defaults to return (label || '').toLocaleLowerCase().indexOf(search.toLocaleLowerCase()) > -1
  - label in our code is set to displayName so only search by displayName works Since we already get our results by
 - searching(DAV & OCS), I don't see why we need the filterBy option in the first place
   - So I just make the return value true
   - In case we do need the option, we should make it check for both displayName and email attributes

### DIFF
--- a/src/components/AppNavigation/Settings/SettingsAddressbookShare.vue
+++ b/src/components/AppNavigation/Settings/SettingsAddressbookShare.vue
@@ -31,6 +31,7 @@
 			:placeholder="placeholder"
 			:class="{ 'showContent': inputGiven, 'icon-loading': isLoading }"
 			:user-select="true"
+			:filter-by="filterResults"
 			open-direction="bottom"
 			track-by="user"
 			label="displayName"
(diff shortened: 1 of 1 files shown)


hamza221:
@avinash-0007  thank you for your contribution, Can you make sure to sign off the commit, to have a green CI

src/components/AppNavigation/Settings/SettingsAddressbookShare.vue
```diff
@@ -107,6 +108,18 @@ export default {
...
 			this.$store.dispatch('shareAddressbook', { addressbook: this.addressbook, user, displayName, uri, isGroup })
 		},
+		/**
+		 * Function to filter results in NcSelect
+		 *
+		 * @param {object} option
+		 * @param {string} label
+		 * @param {string} search
+		 */
+		/* eslint-disable @typescript-eslint/no-unused-vars */
```
ChristophWurst, line 118: I suggest to drop the arguments from the function instead of the eslint line

ChristophWurst (commented):
@avinash-0007 could you please sign-off your commit? Unfortunately we can't do this for you, because it's a mechanism to guarantee that this is your code

PR #3818, opened 2024-02-20
feat: Rebrand circles to teams in the frontend

juliushaertl:
- [x] Replace frontend terms
- [x] Replace icon


ChristophWurst (approved):

juliushaertl:
@ChristophWurst This might have an effect on the release milestone as contacts supports multiple major version but the change here is probably only reasonable for 29+

ChristophWurst:
We can do a major bump and cut off 28

GretaD (approved):

ChristophWurst:
The cutoff is missing. I'll do a follow-up

PR #3837, opened 2024-03-05
enh(sharing): Rename `delete` to `unshare from me` for shared address books

miaulalala:


src/components/AppNavigation/Settings/SettingsAddressbook.vue
```diff
@@ -107,14 +107,23 @@
 				</template>
 
 				<!-- delete addressbook -->
-				<ActionButton v-if="hasMultipleAddressbooks"
+				<ActionButton v-if="hasMultipleAddressbooks && addressbook.url.includes('shared_by')"
```
ChristophWurst, line 110: let's test the principal value: https://github.com/nextcloud/calendar/blob/bb358250c2c641c967bfd66e1987c8180745b82a/src/models/calendar.js#L108C1-L114C3

miaulalala: I don't have the principal for address books - mapDavCollectionToAddressbook only takes the address book. How can I add the principal?

ChristophWurst: I don't see where it's triggered but Contacts dispatches a `PROPFIND /remote.php/dav/principals/users/admin/`. We might already have access to the data :rocket:

src/components/AppNavigation/Settings/SettingsAddressbook.vue
```diff
@@ -107,7 +107,15 @@
 				</template>
 
 				<!-- delete addressbook -->
-				<ActionButton v-if="hasMultipleAddressbooks"
+				<ActionButton v-if="hasMultipleAddressbooks && !addressbook.url.includes(addressbook.owner)"
```
ChristophWurst, line 110: What is the idea behind `!addressbook.url.includes(addressbook.owner)`?

miaulalala: nvm, I got it wrong, I will try and find the principal uri somehow.

GVodyanov:
Created a principals.js store (in Pinia instead of Vuex), and implemented it where needed. @miaulalala Could you please make sure that the structure of the `principalUrl` in `SettingsAddressbook.vue` is indeed as you want it to be in order for the `v-if` to work?

PS for future reference seeing as dealing with this was a bit of a blocker, apparently when having to use stores in `beforeMount`, don't use `mapStores` seeing as it isn't initialized at that time, and the store is undefined. Caused me a bit of troubles 😅

PR #3868, opened 2024-04-05
feat(deps): Add nextcloud 30 support

hamza221:


ChristophWurst (approved):
I would keep this for main only and leave stable6.0 29-

PR #3881, opened 2024-04-11
Add searching capability to ContactPicker

GVodyanov:
I noticed that the search bar in the ContactPicker isn't actually connected to anything
Screenshot from 2024-04-11 20-21-19


hamza221:
I'm not sure if this fix is necessary, 
This is captured on main 

https://github.com/nextcloud/contacts/assets/40746210/08eefd5b-f037-43c3-9908-1489f2984ae0

hamza221:
Also this change shows SAB contacts adding them to a group will fail anyways.

GVodyanov:
Yeah so after some testing I realized I was a bit stupid and did all the testing with read-only contacts, didn't know that was a thing 😅

PR #3925, opened 2024-05-15
chore: create renovate

GretaD:
ref https://github.com/nextcloud/groupware/issues/77
fixes https://github.com/nextcloud/contacts/issues/3916


GretaD:
probably we dont need all those dependencies but i dont know which one we dont need

ChristophWurst (approved):

Issue #2511, opened 2021-11-11
A freshly created group is not persisted after a page reload


pmarini-nc:
### Describe the bug
A freshly created group is not persisted after a page reload

**To Reproduce**
1- add a group
2- add one or more users to the group
3- verify the user is (users are) in the group
4- reload the page

**Expected behavior**
The group just created is there

**Actual behavior**
The group doesn't exist

**Screenshots**

![Peek 2021-11-11 15-39](https://user-images.githubusercontent.com/85954408/141320371-1cd95397-a988-45ef-b889-bc3f2ed2c21d.gif)

### Server configuration

[2021-11-11.md](https://github.com/nextcloud/contacts/files/7521321/2021-11-11.md)

**List of activated apps:**

See system report

**Nextcloud configuration:**

See system report

### Client configuration
**Browser:** Firefox 94

**Operating system:** Ubuntu Desktop 21.04

**CardDAV-clients:** -

miaulalala:
Can you have a look at your nextcloud.log and also try the same process with the dev console open please?

skjnldsv:
This is per design.
Groups are virtual, they only exists in vcard they're written on.
They will persists if you add at least one contacts to it

skjnldsv:
Would require #213

pmarini-nc:
> This is per design. Groups are virtual, they only exists in vcard they're written on. They will persists if you add at least one contacts to it

What do you mean? I have added a contact to the group, but still the group disappears.

pmarini-nc:
> Can you have a look at your nextcloud.log and also try the same process with the dev console open please?

I get this error in nextcloud.log
```
{
  "reqId": "GjeGeWCh3VMLk5l2XflX",
  "level": 0,
  "time": "2021-11-11T15:14:19+00:00",
  "remoteAddr": "10.23.46.1",
  "user": "sbianchi",
  "app": "webdav",
  "method": "PUT",
  "url": "/remote.php/dav/addressbooks/users/sbianchi/z-app-generated--contactsinteraction--recent/3",
  "message": "Node with name '3' could not be found",
  "userAgent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0",
  "version": "22.2.0.2",
  "exception": {
    "Exception": "Sabre\\DAV\\Exception\\NotFound",
    "Message": "Node with name '3' could not be found",
    "Code": 0,
    "Trace": [
      {
[...]

PR #3963, opened 2024-06-11
enh/Improve group adding

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/2511

The idea is to have the add contact modal show up immediately after creating a new group and for it to not show up in the sidebar until you actually added someone, to not give users the "false hope" of seeing it in the sidebar but then reloading the page and seeing that it hadn't been saved.

TODO:
- [x] Make modal open
- [x] Switch to new contact page after modal close
- [x] Close add group modal when add contacts modal opens
- [x] Remove empty groups from sidebar
- [x] When trying to create an already existing group, open modal to add contacts to that group. This is needed in the following scenario:
    - You create a new group but don't add any contacts
    - You therefore do not see in the sidebar
    - You change your mind and want to add contacts to that group
    - You type in the name into the add group and get a "This group already exists error" with no way to continue. With the modal opening you can actually add contacts instead of just having the error.
- [x] After adding contacts to a group you get rerouted to the group view

### DIFF
--- a/src/components/AppNavigation/GroupNavigationItem.vue
+++ b/src/components/AppNavigation/GroupNavigationItem.vue
@@ -5,6 +5,7 @@
 <template>
 	<div class="group-drop-area"
 		data-testid="group-drop-area"
+		v-if="!isEmpty"
 		@drop="onDrop($event, group)"
 		@dragenter.prevent
 		@dragover="onDragOver($event)"
@@ -118,6 +119,10 @@ export default {
 	computed: {
 		contacts() {
 			return this.$store.getters.getContacts
+		},
+
+		isEmpty() {
+			return this.group.contacts.length === 0
 		},
 	},
 
--- a/src/components/AppNavigation/RootNavigation.vue
+++ b/src/components/AppNavigation/RootNavigation.vue
@@ -87,7 +87,7 @@
 					<IconAdd :size="20" />
 				</template>
 				<template #actions>
-					<ActionText>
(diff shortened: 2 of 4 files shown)


hamza221:
Can you add screenshots for designers

hamza221:
redirecting doesn't work maybe because of 
image

GVodyanov:
> redirecting doesn't work maybe because of image

Thanks, should work now

GretaD:
the modal part looks good, but i cannot add a user because when i look for it, it doesnt show. A follow up fix should be also the close modal button :)

Screenshot from 2024-06-18 15-00-19

hamza221:
> the modal part looks good, but i cannot add a user because when i look for it, it doesnt show. A follow up fix should be also the close modal button :)
> 
> Screenshot from 2024-06-18 15-00-19

You can't add SAB contacts if that's the case

GVodyanov:
> the modal part looks good, but i cannot add a user because when i look for it, it doesnt show. A follow up fix should be also the close modal button :)
> 
> Screenshot from 2024-06-18 15-00-19

Yeah, I had already noticed that some time ago and tried "fixing" it (https://github.com/nextcloud/contacts/pull/3881) only to find out it is indeed supposed to work that way

GVodyanov:
Also @GretaD, how does this look?
image

SebastianKrupinski (approved):
Very Nice! 

Working as expected. No Errors in the console.

PR #4007, opened 2024-07-02
fix(deps)!: drop nextcloud 29 support

hamza221:


ChristophWurst (changes requested):
Please make this a semantic breaking commit

ChristophWurst (approved):

Issue #4031, opened 2024-07-15
When creating a new group the arrow isn't aligned with the input space


GVodyanov:
### Steps to reproduce

1. Go to contacts
2. Try to create a new group

### Expected behavior

The arrow should be in the middle

### Actual behavior

![image](https://github.com/user-attachments/assets/cf059ba3-bd29-4873-8b52-31cf76d698ce)

### Contact version

Main

### PHP engine version

None

### Web server

None

### Database

None

GVodyanov:
This somehow fixed itself, yay!

PR #4037, opened 2024-07-17
RootNavigation: align submit add new group icon with input

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4031

| A | B |
| - | - |
| Screenshot from 2024-07-17 17-09-15| Screenshot from 2024-07-17 17-08-37 |

I know that the solution isn't the best, but we have a custom icon size in contacts, so the bug isn't related to nextcloud vue


GVodyanov:
Already fixed (somehow)

PR #4060, opened 2024-07-24
ContactsListItem: make horizontal and  vertical padding equal

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4029

| After | Before |
| - | - |
| Screenshot from 2024-07-24 17-28-20 | Screenshot from 2024-07-24 17-30-27 |

Wasn't able to make this work without a `:deep` selector, if anyone has better ideas...

### DIFF
--- a/src/components/ContactsList/ContactsListItem.vue
+++ b/src/components/ContactsList/ContactsListItem.vue
@@ -178,8 +178,10 @@ export default {
 	}
 }
 
-.list-item-style {
+:deep(.list-item) {
 	list-style: none;
+	padding-inline: calc(var(--default-grid-baseline) * 3);
+	padding-block: calc(var(--default-grid-baseline) * 2);
 }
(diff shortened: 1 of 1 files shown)


GretaD:
on the latest main, it is shown like this: is this still relevant to fix?
Screenshot from 2024-07-24 18-24-08

GVodyanov:
> on the latest main, it is shown like this: is this still relevant to fix?
> Screenshot from 2024-07-24 18-24-08
> 

I would say yes, the right padding is still larger than the top one

GretaD:
> > on the latest main, it is shown like this: is this still relevant to fix?
> > Screenshot from 2024-07-24 18-24-08
> 
> I would say yes, the right padding is still larger than the top one

ohh, i see the issue now. I thoughts was the left space from the avatar.

Issue #3996, opened 2024-06-19
Large teams not managable/usable and button not working


Dennis1993:
### Steps to reproduce

1. Create a team
2. Add members to this team (in my case 42)
3. Add a new member by clicking the button "Add member"
4. In this list not all users are available the list ends after some users
5. Type a new name in the searchbar shows not all users too

Out users are from the active directory

### Expected behavior

All 42 users should be listed and not only ~30 users.
In the left navigation behind the team name the three dots menu is not working if I click on "Add member". Nothing happen. This should be working.

### Actual behavior

Not all users are shown in the list and the search function dont show all users. The button "Add member" is not working on the left sidebar, only in the middle of the screen opens the popup

![image](https://github.com/nextcloud/contacts/assets/1473674/a19a2c1b-590b-472b-86f9-835452066fb2)

![image](https://github.com/nextcloud/contacts/assets/1473674/2fa190a0-b68b-48e6-99b0-6917a8b301c5)

### Server operating system

Ubuntu 20.04

### Web server

Apache

### Database engine version

MariaDB

### PHP engine version

Other

### Nextcloud version

29.0.2

### Updated from an older installed version or fresh install

Updated from an older version

### Browser

Edge 126

### Client operating system

Windows 11

### Web server error log

```Text
empty
```

### Nextcloud log

```Text
empty
```

### Browser log

```Text
empty
```

hamza221:
"Add member" just redirects to to the Team management page now, I agree that it's misleading and should either be renamed or actually open the modal https://github.com/nextcloud/contacts/blob/c4ba7bccfc60c46f266d406985420b336c6e7e94/src/mixins/CircleActionsMixin.js#L143

hamza221:
The other part of the bug should be caused by https://github.com/nextcloud/contacts/pull/3831

jakondacar2ner:
I have the same problem with the listed all users of team. I`d tried several browsers, computers nothing helped.

![Screenshot_3](https://github.com/user-attachments/assets/a6f434f5-819f-46ba-b317-0ef881f4386c)

if somebody handle with this please give the solution.

Dennis1993:
29.0.4 too. App is unusable…

PR #4073, opened 2024-08-05
fix(CircleDetails): Improve calculation of member list

mejo-:
Also add a more icon if the team has more members than displayed.

Related to #3996 

## Screenshots

Before | After
--- | ---
grafik | grafik


ChristophWurst (approved):
Did not test but code looks good!

PR #4075, opened 2024-08-05
fix(MemberList): Don't use VirtualList for listing members

mejo-:
VirtualList would require a fixed container height and using VirtualList here seems like a premature optimization to me.

Fixes: #3996

## Screenshots

Before | After
--- | ---
grafik | grafik
grafik | grafik


mejo-:
Added screenshots to emphasize the visual changes.

ChristophWurst (approved):

PR #4092, opened 2024-08-22
PropertySelect: hide PropertySelects with only one selectable option

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4028

### DIFF
--- a/src/components/Properties/PropertySelect.vue
+++ b/src/components/Properties/PropertySelect.vue
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<div v-if="propModel && showProperty" class="property">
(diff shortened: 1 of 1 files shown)


hamza221:
Just reading the code, I think this would remove any section with a single option, not only addressbook, is it the required behaviour ?

GVodyanov:
> Just reading the code, I think this would remove any section with a single option, not only addressbook, is it the required behaviour ?

Yes. But it should make sense for all the other ones. Contact groups are not affected by this. Out of the details we have, this would affect "manager" and "address book"

SebastianKrupinski (approved):

SebastianKrupinski:
Works!

Issue #2365, opened 2021-07-21
Add a contact (with umlauts) to a group (with umlauts) breaks encoding


kesselb:
### Describe the bug

Add a contact (with umlauts) to a group (with umlauts) breaks encoding

**To Reproduce**

- Create a contact with umlauts (e.g. Mäx)
- Create a group with umlauts (e.g. Täst)
- Add contact Mäx to group Täst

**Expected behavior**

Mäx should be Mäx

**Actual behavior**

Mäx is now MÃ¤x

**Screenshots**

https://user-images.githubusercontent.com/3902676/126527774-5428c7da-1e07-43c0-baf6-910c7c73b2b3.mp4

kangaroo72:
I think, I can confirm.
Actually I have 282 Contacts, and the umlauts are suddenly broken.
What happened??

![image](https://user-images.githubusercontent.com/6906390/194835824-911ca713-d97a-402b-ae32-3a3b45a60323.png)

ChristophWurst:
You tell us :)

Did you update the app? Did you update Nextcloud?

kangaroo72:
Actually I'm on NC 24.0.5 and Contacts 4.2.2

kangaroo72:
Same on 24.0.6
![grafik](https://user-images.githubusercontent.com/6906390/194853782-9825549c-c0e5-41fc-b60f-82bfcca98359.png)

ChristophWurst:
Could you share the redacted .vcf?

kangaroo72:
Here is one affected vcf.
[D40302BD-DDB1-485D-AF45-A6163E922C85.vcf.zip](https://github.com/nextcloud/contacts/files/9748472/D40302BD-DDB1-485D-AF45-A6163E922C85.vcf.zip)

joshtrichards:
Related: #946

Issue #946, opened 2019-02-22
umlauts make name disappear


beinaendi:
Hi everybody,

I am a German speaking person, thus many of my contacts have umlauts like ä, ö, ü. When I download any contact group as vcf, it works perfectly, but prenames or surnames with umlauts are deleted completely.

To reproduce
1. Make a new contact just like "Max Müller"
2. Add the contact to a group
3. Download the Group
4. Import the vcf file in your contacts programm (in my case contacts on OS X)
5. "Max Müller" will appear as "Max" without any surname

This happens on next cloud 15.0.4 with contacts 3.0.3

Best wishes

skjnldsv:
@beinaendi Hello and thanks for your report :)
Could you attach the vcf file here please? (rename to .txt)

beinaendi:
@skjnldsv Hello, of course not ;-) But I made you an example with Max Müller. After downloading it, the surname appears as "M¸ller". Please find the renamed vcf file attached.
[test.txt](https://github.com/nextcloud/contacts/files/2893816/test.txt)

skjnldsv:
@beinaendi okay, so the display of the contact is ok on the nextcloud contacts app, correct?
Only when importing it into your OSX contacts manager it starts to misbehave? 

Because looking at your vcard, `FN:Max Müller` is correct and have the proper umlaut, and `N;SORT-AS=Max Müller:Müller;Max;;;` as well :woman_shrugging:

beinaendi:
@skjnldsv Yes, on the website everything appears perfect. But when I press the download button (arrow down) next to the contact group, this is what happens. I cannot understand why the vcard looks correct on your computer. This is what it looks like on a Mac with macOS Mojave 10.14.3

BEGIN:VCARD
VERSION:3.0
PRODID:-//Sabre//Sabre VObject 4.1.6//EN
UID:urn:uuid:40595350-0230-4a91-9863-9eeccf416771
FN:Max M¸ller
CATEGORIES:test
N;SORT-AS=Max M¸ller:M¸ller;Max;;;
EMAIL;TYPE=PREF:max@mueller.de
TEL;TYPE=CELL:+49 123 456789
ORG:Max AG;
END:VCARD

beinaendi:
So it seems to be an OS X issue. Tried to open the file with TextEdit, BBEdit and Word. Same problem in all programs.

beinaendi:
[...]

PR #4107, opened 2024-09-02
GroupNavigationItem: encode exported contact groups as UTF-8

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/2365
Fix https://github.com/nextcloud/contacts/issues/946

Previously exported contact groups would be encoded either as `us-ascii`, or `iso-8859-1` if they had umlauts or other special characters. I made all of them be exported as `utf-8`, as singular contacts are.

The main issues with this were that when contact groups were encoded as `iso-8859-1` they would have a weird character instead of umlauts and would bug out while importing.


st3iny (approved):
Very weird API. Especially that it doesn't assume utf-8 strings by default.

Tested and works!

Issue #1148, opened 2019-06-23
Support rfc 6474, deathplace & birthplace 


skjnldsv:
See https://tools.ietf.org/html/rfc6474

We already support DEATHDATE

PR #4173, opened 2024-10-07
fix: support for RFC 6474

SebastianKrupinski:
Resolves: #1148

Adds support for RFC 6474

### DIFF
--- a/src/models/rfcProps.js
+++ b/src/models/rfcProps.js
@@ -135,6 +135,12 @@ const properties = {
 		],
 		primary: true,
 	},
+	birthplace: {
+		readableName: t('contacts', 'Place of birth'),
+		icon: 'icon-location',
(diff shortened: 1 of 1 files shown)


GVodyanov (approved):
Cool! Now my future friends from the retirement home will know my deathplace

src/models/rfcProps.js
```diff
@@ -153,6 +159,12 @@ const properties = {
 		},
 		primary: false,
 	},
+	deathplace: {
+		readableName: t('contacts', 'Place of death'),
+		icon: 'icon-location',
+		force: 'text',
```
hamza221, line 165: It can also Be a URI, I'm not sure if `'text'` would break anything

st3iny (approved):

Issue #4164, opened 2024-09-26
Unable to choose an avatar from Files


cwerstuik:
### Describe the bug

I'd like to update a contact's avatar. I can upload the image from my computer, but when I try to select one from my Nextcloud files, there's a problem. I can navigate to the image, but there's no button anyway to actually set the image as the avatar.

### Steps to reproduce

1. Go to any contact in my address book
2. Click on the icon to edit avatar and select "Choose from Files"
3. Navigate to image from the "Pick an Avatar" modal
4. No button exists to actually set the image as the avatar

### Expected behavior

I can look through my Nextcloud files and select an image as a contact's avatar

### Actual behavior

Can navigate to image, but can't see any button to set image as avatar

![Screenshot_20240925_191755](https://github.com/user-attachments/assets/e12e800c-8679-4bca-b0b1-1c5a997fe13e)

### Contact version

6.1.0

### Operating system

Linux

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database

MySQL

hamza221:
maybe related to https://github.com/nextcloud/calendar/pull/6383

PR #4176, opened 2024-10-07
fix:filepicker for contact avatar

hamza221:
fix [#4164](https://github.com/nextcloud/contacts/issues/4164)


GVodyanov (approved):
Bruh how did we not see this before

PR #4186, opened 2024-10-17
chore: add PHP 8.4

SebastianKrupinski:
Fingers crossed

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -20,12 +20,9 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        include:
-          - php-versions: '8.1'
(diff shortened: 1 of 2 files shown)


hamza221:
this is missing
https://github.com/nextcloud/contacts/blob/c49a50f3d568c3a973d986ab2f3bf1da69e902e0/composer.json#L23

SebastianKrupinski:
> this is missing
> 
> https://github.com/nextcloud/contacts/blob/c49a50f3d568c3a973d986ab2f3bf1da69e902e0/composer.json#L23

Just checked our other groupware apps. We haven't increased the max php versions in the composer on them yet, and probably because its not fully released yet.

hamza221 (approved):

PR #4231, opened 2024-11-29
Move users_picker profile custom picker to contacts

julien-nc:
This replaces #3487. I can't push in the branch there.

@ChristophWurst 

* It's rebased on main
* Adjusted the vite config
* The bio processing is now using multi-byte proof functions
* Nitpick addressed: The condition in the reference provider was inverted
* Now check if the profile is enabled (when resolving a link)

About https://github.com/nextcloud/contacts/pull/3487#pullrequestreview-1696541681 I'm not sure I see the issue. The frontend is using the `ocs/v2.php/core/autocomplete/get` endpoint which should respect the sharing settings.

The reference provider now checks if the profile is enabled or not.
It seems fine to resolve any `/u/USER_ID` link as long as the profile is enabled.


SebastianKrupinski:
Hi @julien-nc 

How do I test? What am I looking for.

julien-nc:
@SebastianKrupinski This adds a "profile picker" entry in the smart picker and a reference widget (link preview) for `/u/USER_ID` links.

ChristophWurst:
Dear @julien-nc,

Thank you for your contribution!

> About [#3487 (review)](https://github.com/nextcloud/contacts/pull/3487#pullrequestreview-1696541681) I'm not sure I see the issue. The frontend is using the `ocs/v2.php/core/autocomplete/get` endpoint which should respect the sharing settings.

That is correct for the part of searching. But once a reference is selected, there is no more check in ProfilePickerReferenceProvider.
I could abuse that to render a reference to a user I would otherwise not see. The reference contains any non-private data.

I suggest we test this specific scenario and see how the code behaves. I just want to have clarity on this point before the code goes into the main branch.

ChristophWurst:
@julien-nc on a higher level, what do you think about moving the app/code into server instead? The picker picks *users*, not *contacts*. There is nothing specific to contacts (or event teams) in the code.

julien-nc:
@ChristophWurst Yes we can move it to server.

I'd still like to have your opinion on the last changes. We now check the visibility of the profile fields with the ProfileManager. This respects the scopes and the visibility settings.

> I could abuse that to render a reference to a user I would otherwise not see. The reference contains any non-private data.

Now the preview can't display any information that would not be visible by clicking the link to browse the profile page.

lib/Reference/ProfilePickerReferenceProvider.php
```diff
@@ -0,0 +1,180 @@
...
+				'headline' => $this->profileManager->isProfileFieldVisible(IAccountManager::PROPERTY_HEADLINE, $user, $currentUser) ? $headline->getValue() : null,
+				'location' => $isLocationVisible ? $location->getValue() : null,
+				'location_url' => $isLocationVisible ? $this->getOpenStreetLocationUrl($location->getValue()) : null,
+				'website' => $this->profileManager->isProfileFieldVisible(IAccountManager::PROPERTY_WEBSITE, $user, $currentUser) ? $website->getValue() : null,
+				'organisation' => $this->profileManager->isProfileFieldVisible(IAccountManager::PROPERTY_ORGANISATION, $user, $currentUser) ? $organisation->getValue() : null,
+				'role' => $this->profileManager->isProfileFieldVisible(IAccountManager::PROPERTY_ROLE, $user, $currentUser) ? $role->getValue() : null,
+				'url' => $referenceText,
+			]
+		);
+		return $reference;
```
ChristophWurst, line 140: With these new checks it looks like we are only exposing data that should be visible to the active user :+1: 

Looks good

ChristophWurst:
Is the plan to move the *user* picker into the *Contacts* app still?

julien-nc:
No. Let's decide how to move it to server. Should it be a shipped app or in the core? I leaning towards a shipped app to isolate it (and make the move simpler :grin: ).

ChristophWurst:
I'm fine with either. Shipped app is easy to set up, but cumbersome to maintain. Moving to server is more work now but easier to keep alive in the long run because it will live in a very active repository.

julien-nc:
It was first moved to a new shipped app (user_picker) and finally moved in the profile shipped app:
* https://github.com/nextcloud/server/pull/58518

Issue #4090, opened 2024-08-20
"Import contacts" button shows in navigation when a team is selected


jancborchardt:
### Describe the bug

Meta groups on top: All fine | Contact groups: All fine | Teams: "Import contacts" button shows
-|-|-
![image](https://github.com/user-attachments/assets/b50eea28-798b-49e5-bb5d-6ce9524a808c)|![image](https://github.com/user-attachments/assets/e39536b0-3abc-46d9-a08b-b253d95ca1ec)|![image](https://github.com/user-attachments/assets/3e6e985a-0d42-46c5-8277-a295023cb23f)

### Steps to reproduce

1. Go to Contacts
2. Click on any team
3. Notice the "Import contacts" button appearing on the top of the navigation

### Expected behavior

The button "Import contacts" only shows on emptycontent, and in the settings.

### Actual behavior

The button "Import contacts" also shows when a team is selected from the left navigation.

### Contact version

6.1.0-alpha.2

### PHP engine version

None

### Web server

None

### Database

None

PR #4236, opened 2024-12-01
Add proper routing for circles, removing the import button when not needed

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4090

### DIFF
--- a/src/views/Contacts.vue
+++ b/src/views/Contacts.vue
@@ -11,7 +11,7 @@
 			:selected-group="selectedGroup"
 			:selected-contact="selectedContact">
 			<div class="import-and-new-contact-buttons">
-				<SettingsImportContacts v-if="!loadingContacts && isEmptyGroup && !isChartView" />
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@GretaD @hamza221 can we get this merged?

src/views/Contacts.vue
```diff
@@ -193,6 +196,8 @@ export default {
 				return this.sortedContacts
 			} else if (this.selectedGroup === GROUP_NO_GROUP_CONTACTS) {
 				return this.ungroupedContacts.map(contact => this.sortedContacts.find(item => item.key === contact.key))
+			} else if (this.selectedGroup === ROUTE_CIRCLE) {
+				return this.circles
```
hamza221, line 200: I think this too hacky, because you're returning circles to contact lists when ContactList should be empty. 
I think a better approach would be to and a condition here 
https://github.com/nextcloud/contacts/blob/fix/circlesRouting/src/components/AppNavigation/Settings/SettingsImportContacts.vue#L22

Issue #4148, opened 2024-09-16
contacts-index.css blocked due to MIME type (“text/html”) mismatch


melroy89:
### Describe the bug

Incorrect, missing contacts-index.css file when on the files tab when using Nextcloud 30.

```sh
The resource from “https://cloud.melroy.org/apps/contacts/css/contacts-index.css?v=1b4503c0-5” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff). files
```

When I visit t his css file, I got redirect to: https://cloud.melroy.org/apps/contacts/All%20contacts/4af1a2e9-c232-4fd9-8c12-62ff515bed78~default. And it shows me:

```sh
Error: page.tabs[tab.id] is undefined undefined
```

Followed by:  `group css not found`

As well as: 

```
Error: Debounced method called with different contexts.
    d https://cloud.melroy.org/apps/contacts/js/index-BYtwKcsR.chunk.mjs:2
    mounted https://cloud.melroy.org/apps/contacts/js/contacts-main.mjs?v=305544fb-5:3
    qt https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    It https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    insert https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    bt https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    _l https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    _update https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    r https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    get https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    run https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    tf https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    Kn https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
    hn https://cloud.melroy.org/apps/contacts/js/index-b2_Rf36J.chunk.mjs:2
[...]

PR #4244, opened 2024-12-05
fix: remove unused style

solracsf:
Fix https://github.com/nextcloud/contacts/issues/4148

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -11,7 +11,6 @@ use OCA\Contacts\AppInfo\Application;
 use OCA\Contacts\Service\GroupSharingService;
 use OCA\Contacts\Service\SocialApiService;
 use OCP\App\IAppManager;
-
 use OCP\AppFramework\Controller;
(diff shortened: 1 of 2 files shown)


SebastianKrupinski:
Tested Works.

Before:
Screenshot 2024-12-05 141310

After:
image

SebastianKrupinski (approved):
Tested. Works. Thanks for fixing this.

Also thank you for updating the constructor!

solracsf:
Thanks. I'm not familiar with backports policy in this repo but please consider backporting it 👍🏻

SebastianKrupinski:
We normally try to back port fixes 3 versions for all repos, for the contacts app that would be 6.1, 6.0, 5.5.

The only issue is that with the constructor changes, back porting this fix is not possible, version 6.0 supports a min php version of 7.4 which does not support the constructor variable declarations. Also the removal of the addInitScript check makes pack porting to 6.0 impossible as 6.0 supports a min Nc version of 27.

So long story short we can't really back port these changes. Sorry

PR #4261, opened 2024-12-19
fix: specify origin of request when search for users

### LINKED ISSUE #2370: specify origin of request when search for users
It would help to have a confirmation that the request is coming from the Contacts App when running 

`GET /ocs/v2.php/apps/files_sharing/api/v1/sharees?format=json&itemType=file&search=test&perPage=25&shareType[]=0&shareType[]=1&shareType[]=4&shareType[]=7&lookup=false`

The idea is that when sharing a file, the same request is executed and should only returns Circles to which the current user is a member. But in the contacts app, the request should returns also visible Circles, in case the current user wants to add a Circles he is not currently member as a Member.

I think changing the `itemType` from '`file` to `contacts` will be enough.
It is not the perfect solution but it will work better, unless we switch to another endpoint to catch the list of users.

Right now, this is how the Circles App detect the origin of the request is from the contacts app:
https://github.com/nextcloud/circles/pull/724/files#diff-85950fb08bcb016579e9da6fb82f2335314fe42638d0ae76f5cb8828681ebfb8R101-R103

hamza221:
fix #2370

### DIFF
--- a/src/services/collaborationAutocompletion.js
+++ b/src/services/collaborationAutocompletion.js
@@ -21,7 +21,7 @@ export const getSuggestions = async function(search) {
 	const request = await axios.get(generateOcsUrl('apps/files_sharing/api/v1/sharees'), {
 		params: {
 			format: 'json',
-			itemType: 'file',
+			itemType: 'contacts',
 			search,
 			perPage: maxAutocompleteResults,
 			shareType,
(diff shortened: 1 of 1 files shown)


st3iny (approved):
Not sure how to test though.

PR #4299, opened 2025-01-14
Typo: Adressbooks -> Addressbooks

IntGrah:


IntGrah:
Or even "address books"?

hamza221:
Please don't change the translation files manually, instead change where the strings originate from in the app source code

hamza221 (changes requested):
Please don't change the translation files manually, instead change where the strings originate from in the app source code

ChristophWurst:
Hi @IntGrah!
Thank you very much for you contribution. As Hamza said, we don't modify the l10n files manually. There is an automation for it. So please undo the changes in `l10n/`. The other file change is fine and can be integrated.

ChristophWurst:
@IntGrah friendly ping

PR #4309, opened 2025-01-23
feat(deps): Add Nextcloud 32 support

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

PR #4310, opened 2025-01-24
Revert "feat(deps): Add Nextcloud 32 support"

hamza221:
Reverts nextcloud/contacts#4309


ChristophWurst (approved):
Makes sense after reading https://github.com/nextcloud/contacts/pull/4309#issuecomment-2612779377

PR #4311, opened 2025-01-24
chore: add support for nextcloud 32

hamza221:


ChristophWurst (approved):
IMO these commits should be `feat`. With release automation we'd want a minor bump at least.

PR #4334, opened 2025-02-07
fix: Make circle members visible again by moving them out of the modal

juliusknorr:
As discussed between @sorbaugh and @jancborchardt this moves the team member list out of the modal and onto the team page again.

Screenshot 2025-02-07 at 09 58 57

### DIFF
--- a/src/components/AppContent/CircleContent.vue
+++ b/src/components/AppContent/CircleContent.vue
@@ -4,41 +4,20 @@
 -->
 
 <template>
-	<AppContent v-if="!circle">
-		<EmptyContent :name="t('contacts', 'Please select a team')">
+	<AppContent>
+		<EmptyContent v-if="!circle" :name="t('contacts', 'Please select a team')">
 			<template #icon>
 				<AccountGroup :size="20" />
 			</template>
 		</EmptyContent>
-	</AppContent>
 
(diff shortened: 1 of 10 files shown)


juliusknorr:
I'm stuggling with a bug that is also there on main but would be good to address with this one as it is quite noticable:

When adding a new member it is not showing up immediately in the list, is suspect some vue/vuex reactivity issue but could not figure out why the list does not update.

jancborchardt (approved):
Looks good design-wise for now. We could think about making the user list 2-column so the 3-dot action menu is not so far off to the right, but that can be done in a follow-ilup.

Also cc @marcoambrosini for review.

GVodyanov (commented):
It works well for me, I didn't experience the reactivity issue when adding new users. If you could provide some more information on how to reproduce it, I could look into it.

marcoambrosini (approved):

juliusknorr:
> It works well for me, I didn't experience the reactivity issue when adding new users. If you could provide some more information on how to reproduce it, I could look into it.

I just built using `npm run watch` and then created a new circle and added a user. Using firefox, but if that is not reproducable, maybe we are good and it is just something odd on my setup? Maybe vue dev tools interfering?

hamza221 (approved):

PR #4380, opened 2025-03-12
Adding error on invalid email-address for contact-editor

### LINKED ISSUE #4379: Contact Email Address not cleaned/validated
### Describe the bug

Email Entry is not cleaned on save, when email-adresses are invalid mail delivery fails as unknown error.

The validation could also apply to phone number, address... but i think email is the most critical, because of the automatic nextcloud messages.

This seems like missing a common feature? Is this on purpose or was just never a reason to implement.

_I recently had a problem where it would not send calender invites for contacts bc. they contained spaces and i could not find the problem at first._

### Steps to reproduce

1. Go to new or existing contact
2. Set email entry to invalid address (spaces, no @, no tld...) 
3. Save
4. Cause Email-Send Action (calender invitation, ...)

### Expected behavior

It should'nt be possible to have contacts with invalid email-adresses.

- [x] Trimming spaces out of email (if some format accidently adds it)
- [ ] Validating and rejecting invalid email-adresses on edit
    - [x] In WebUi
    - [ ] On File-Import
    - [ ] On CardDav Sync
    - [ ] ?Provide way to fix current entrys?
    

### Actual behavior

You can easily add invalid email addresses to contacts and nextcloud tries to send emails to them but fails as unknown.

### Contact version

7.0.3

### Operating system

_No response_

### PHP engine version

PHP 8.3

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

The profile editor already has a validation system, so can be copied.
Phone Number and Address Validation may be nice, but probably cause problems bc. of different typing, and are not critical, as nextcloud does not use them.

timedin-de:
fixes #4379

- [x] Trimming spaces out of email (if some format accidently adds it)
- [x] Validating and rejecting invalid email-adresses on edit

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,6 +41,7 @@
         "pinia": "^2.3.1",
         "string-natural-compare": "^3.0.1",
         "uuid": "^11.1.0",
+        "validator": "^13.12.0",
         "vue": "~2.7.16",
         "vue-click-outside": "^1.1.0",
         "vue-cropperjs": "^4.2.0",
@@ -18845,6 +18846,14 @@
       "integrity": "sha512-Kvp459HrV2FEJ1CAsi1Ku+MY3kasH19TFykTz2xWmMeq6bk2NU3XXvfJ+Q61m0xktWwt+1HSYf3JZsTms3aRJg==",
       "dev": true
     },
+    "node_modules/validator": {
+      "version": "13.12.0",
+      "resolved": "https://registry.npmjs.org/validator/-/validator-13.12.0.tgz",
+      "integrity": "sha512-c1Q0mCiPlgdTVVVIJIrBuxNicYE+t/7oKeI9MWLj3fh/uq2Pxh/3eeWbVZ4OcGW1TUf53At0njHw5SMdA3tmMg==",
+      "engines": {
+        "node": ">= 0.10"
+      }
+    },
     "node_modules/vfile": {
       "version": "5.3.7",
       "resolved": "https://registry.npmjs.org/vfile/-/vfile-5.3.7.tgz",
@@ -33623,6 +33632,11 @@
           "dev": true
         }
       }
+    },
+    "validator": {
+      "version": "13.12.0",
+      "resolved": "https://registry.npmjs.org/validator/-/validator-13.12.0.tgz",
+      "integrity": "sha512-c1Q0mCiPlgdTVVVIJIrBuxNicYE+t/7oKeI9MWLj3fh/uq2Pxh/3eeWbVZ4OcGW1TUf53At0njHw5SMdA3tmMg=="
     },
     "vfile": {
       "version": "5.3.7",
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "pinia": "^2.3.1",
     "string-natural-compare": "^3.0.1",
     "uuid": "^11.1.0",
+    "validator": "^13.12.0",
     "vue": "~2.7.16",
     "vue-click-outside": "^1.1.0",
     "vue-cropperjs": "^4.2.0",
--- a/src/components/Properties/PropertyText.vue
+++ b/src/components/Properties/PropertyText.vue
@@ -57,6 +57,22 @@
 					@mousemove="resizeHeight"
 					@keypress="resizeHeight" />
 
+				<!-- email with validation-->
+				<NcTextField v-else-if="propName === 'email'"
+					ref="email"
+					:class="{'property__value--with-ext': haveExtHandler}"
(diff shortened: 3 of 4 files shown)


timedin-de:
i implemented the email validator (copied from user settings) to the ui. If an invalid email is entered theres a error note and on safe the value will be reverted to the previous value.

This is all what can be done here, will continue a independent pull request in backend.

src/constants/AccountPropertyConstants.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+// eslint-disable-next-line no-control-regex
+export const VALIDATE_EMAIL_REGEX = /^(?!(?:(?:\x22?\x5C[\x00-\x7E]\x22?)|(?:\x22?[^\x5C\x22]\x22?)){255,})(?!(?:(?:\x22?\x5C[\x00-\x7E]\x22?)|(?:\x22?[^\x5C\x22]\x22?)){65,}@)(?:(?:[\x21\x23-\x27\x2A\x2B\x2D\x2F-\x39\x3D\x3F\x5E-\x7E]+)|(?:\x22(?:[\x01-\x08\x0B\x0C\x0E-\x1F\x21\x23-\x5B\x5D-\x7F]|(?:\x5C[\x00-\x7F]))*\x22))(?:\.(?:(?:[\x21\x23-\x27\x2A\x2B\x2D\x2F-\x39\x3D\x3F\x5E-\x7E]+)|(?:\x22(?:[\x01-\x08\x0B\x0C\x0E-\x1F\x21\x23-\x5B\x5D-\x7F]|(?:\x5C[\x00-\x7F]))*\x22)))*@(?:(?:(?!.*[^.]{64,})(?:(?:(?:xn--)?[a-z0-9]+(?:-+[a-z0-9]+)*\.){1,126}){1,}(?:(?:[a-z][a-z0-9]*)|(?:(?:xn--)[a-z0-9]+))(?:-+[a-z0-9]+)*)|(?:\[(?:(?:IPv6:(?:(?:[a-f0-9]{1,4}(?::[a-f0-9]{1,4}){7})|(?:(?!(?:.*[a-f0-9][:\]]){7,})(?:[a-f0-9]{1,4}(?::[a-f0-9]{1,4}){0,5})?::(?:[a-f0-9]{1,4}(?::[a-f0-9]{1,4}){0,5})?)))|(?:(?:IPv6:(?:(?:[a-f0-9]{1,4}(?::[a-f0-9]{1,4}){5}:)|(?:(?!(?:.*[a-f0-9]:){5,})(?:[a-f0-9]{1,4}(?::[a-f0-9]{1,4}){0,3})?::(?:[a-f0-9]{1,4}(?::[a-f0-9]{1,4}){0,3}:)?)))?(?:(?:25[0-5])|(?:2[0-4][0-9])|(?:1[0-9]{2})|(?:[1-9]?[0-9]))(?:\.(?:(?:25[0-5])|(?:2[0-4][0-9])|(?:1[0-9]{2})|(?:[1-9]?[0-9]))){3}))\]))$/i
```
hamza221, line 12: I'd rather switching to a package such as https://www.npmjs.com/package/address-rfc2822, instead of using a regex here

hamza221:
Thank you for your contribution, it's really helpful to have the check, 
I think it would be better if it doesn't fail silently and revert on save on wrong address. 
Maybe we can a `showError()`toast and keep the same mechanism ( reverting -> saving ) ? 
I can also help adding that If you want me too.

src/constants/AccountPropertyConstants.ts
hamza221: Looks good, thank you

timedin-de:
> Maybe we can a `showError()`toast and keep the same mechanism ( reverting -> saving ) ? I can also help adding that If you want me too.

Yep theoretically i think its easy to implement, but a problem would be that the verification is implemented inside of the email input => for multiple wrong email inputs it will show multiple errors. Would this be a acceptable solution?

I'm not really sure how to implement it in the ContactDetails (or parent) to have it fire only once.

hamza221:
> > Maybe we can a `showError()`toast and keep the same mechanism ( reverting -> saving ) ? I can also help adding that If you want me too.
> 
> Yep theoretically i think its easy to implement, but a problem would be that the verification is implemented inside of the email input field => for multiple wrong email inputs it will show multiple errors. Would this be a acceptable solution?
> 
> I'm not really sure how to implement it in the ContactDetails (or parent) to have it fire only once.

I was thinking to trigger only on `onSave( )` in contactDetails,  by watching `localContact`'s property or simpler adding a simplified `errors`prop array that gets amended with `propretyName -> bool` and is checked on save

PR #4381, opened 2025-03-12
fix: fix team members colum layout in chromium based browsers

icewind1991:
Looks like chrome has some issues, taken the solution from https://stackoverflow.com/questions/3322891/why-is-chrome-cutting-off-text-in-my-css3-multi-column-layout.

Before:

Screenshot_20250312_195902

After: 

Screenshot_20250312_195706

### DIFF
--- a/src/components/MemberList/MemberListItem.vue
+++ b/src/components/MemberList/MemberListItem.vue
@@ -345,5 +345,6 @@ export default {
 .members-list-item {
 	user-select: none;
 	box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


provokateurin (approved):

hamza221 (approved):

susnux (approved):

susnux:
should be backported but not sure which is the 29 branch, `stable6.1` or `stable6.0`?

PR #4412, opened 2025-04-04
refactor: use server container for contacts manager

hamza221:
The app uses internals of another app. With https://github.com/nextcloud/contacts/pull/4394 we had to adjust the constructor arguments because they changed.
This PR replaces the direct instantiation with a service locator.

### DIFF
--- a/lib/Service/SocialApiService.php
+++ b/lib/Service/SocialApiService.php
@@ -11,10 +11,7 @@ namespace OCA\Contacts\Service;
 
 use OCA\Contacts\AppInfo\Application;
 use OCA\Contacts\Service\Social\CompositeSocialProvider;
-
-use OCA\DAV\CardDAV\CardDavBackend;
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

lib/Service/SocialApiService.php
```diff
@@ -9,12 +9,10 @@
 
 namespace OCA\Contacts\Service;
 
+use OC\ServerContainer;
```
ChristophWurst, line 12: There is an interface

ChristophWurst (changes requested):

lib/Service/SocialApiService.php
ChristophWurst: Try ``\Psr\Container\ContainerInterface``

ChristophWurst (approved):

Issue #4416, opened 2025-04-08
Exchange cloud IDs between remote users


redblom:
### Is your feature request related to a problem? Please describe.

This feature addresses the somewhat cumbersome process of retrieving the cloud ID of a remote user you want to federatively share with. 
This feature will enable the user to send an invitation to a remote user to exchange cloud IDs. It basically extends the contacts app with the option to exchange with a remote user name, email and cloud ID resulting in the creation of contact info on each other's system (sort of like exchanging a contact card, including cloud ID). From then on both users can share with each other as easy as with local users.
The features relies on an implementation of the Invitation Workflow (see https://github.com/sara-nl/nc-collaboration).

### Describe the solution you'd like

Some mock-up screenshots of what we're after:

![Image](https://github.com/user-attachments/assets/c91f6e93-4198-4bc8-bd41-d061252f5243)

ArtificialOwl:
Hello @redblom, could it be a duplicate (but yet different approach) of https://github.com/nextcloud/server/pull/51113 ?

hiSandog:
A useful next step would be a minimal fixture for Exchange cloud IDs between remote users that includes one failing input and one nearby control case. That would make it clear whether the fix addresses the reported edge case without widening behavior unexpectedly. `/invite-accepted` is a useful concrete anchor because it is named directly in the report.

PR #4417, opened 2025-04-08
feat: exchange cloud ID

redblom:
**~~This is a work in progress~~**
This PR implements https://github.com/nextcloud/contacts/issues/4416

### About
The goal of this feature is to allow for easy remote contact information exchange (cloud ID, email, name) between 2 users from different instances/organizations based on an invitation request.
The feature is an implementation of the invitation workflow from the OCM specification (https://datatracker.ietf.org/doc/draft-ietf-ocm-open-cloud-mesh/)<br><br>

A user (sender) creates the invitation. This invitation consists of an invite link and (encoded) invite code. The invitation is sent to another user (receiver) via email or through another channel, eg. chat.
After receiving the invitation, the receiver can accept the invitation by following the invite link or by pasting the (encoded) invite code into a manual accept form. 
After accepting, the exchange of user information will take place (OCM `/invite-accepted` endpoint) resulting in the creation of both users in each other's contacts list.
Following the invite link in the email is the most convenient way to accept. It will redirect the receiver via a WAYF (Where Are You From) page immediately to an accept dialog.  
Note: Manual acceptance of the invitation works with both the invite link and (encoded) invite code.

### Screenshots
1_contacts-app-main-page-Mike
Contacts page with new buttons 'Invite contact', 'Accept invite' and 'All invites'.<br><br>

2_fill-in-invite-form
Click 'Invite contact' and fill in form fields:

- Invite label for reference in the Invites list (optional)
- Recipient email (required)
- Personal message that will be sent with the email (optional)
- Check to send copy of the invite email to yourself (optional)

3_invite-sent-list
'All invites' button displays the invites sent.<br><br>

3a_alt-way-to-share-invite
It's also possible to share the invite by sending the invite link or code through an alternative channel (eg. chat).<br><br>

4_invite-email-to-Pete
This what the email to the receiver of the invite looks like.<br><br>

5_copy-of-invite-email-to-Pete
This is what the copy of the email to the sender of the invite looks like.<br><br>

6_wayf-page
Clicking on the invite link will direct the receiver to a WAYF (Where Are You From) page.<br>
Here you choose your institute. You can also fill in your institute's provider address if you are not on the list. A discovery process will run to check whether your provider supports OCM.<br><br>

7_login-after-wayf-page-select-Pete
Login after selecting your institute.<br><br>

8_accept-invite
You will be redirected to an invite accept dialog.<br><br>

9_invite-accepted-Pete
On acceptance the contact information exchange will take place via the OCM `/invite-accepted` endpoint. This will lead to the creation of a new contact (the sender) on the receiver's end.<br><br>

10_invite-accepted-Mike
And on the sender end the receiver contact info will also be added.<br><br>

11_alt-invite-without-email-Jimmie
When the config setting `ocm_invites_optional_mail` is set to false, creating an invite will give you by default the option to share the invite through an alternative channel (eg. chat).<br><br>

12_alt-invite-no-email-copy-code
Copy the invite link or code and send it via your channel of choice.<br><br>

12a_alt-invite-with-email
You may still decide to send the invite via email by clicking the 'Send via email' button.<br><br>

13_alt-manual-accept-invite-Dan
The receiver of the invite can click 'Accept invite' and manually paste the invite link, code, or encoded invite code into the accept form.<br>
And this will lead to the exchange of contact information via the OCM /invite-accepted endpoint.<br><br>

14_alt-manual-invite-accepted-Dan
After contact info exchange the sender will be added as a new contact on the receiver's end.<br><br>

15_alt-manual-invite-accepted-Jimmie
And the receiver will be added as a new contact on the sender's end.<br><br>

### OCM invites admin settings
ocm-invites-admin-settings

### DIFF
--- a/lib/Controller/FederatedInvitesController.php
+++ b/lib/Controller/FederatedInvitesController.php
@@ -1,0 +1,985 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Contacts\Controller;
+
+use Exception;
+use GuzzleHttp\Exception\RequestException;
+use OCA\Contacts\AppInfo\Application;
+use OCA\Contacts\Db\FederatedInvite;
+use OCA\Contacts\Db\FederatedInviteMapper;
+use OCA\Contacts\Exception\ContactAlreadyExistsException;
+use OCA\Contacts\Service\FederatedInvitesService;
+use OCA\Contacts\WayfProvider;
+use OCA\FederatedFileSharing\AddressHandler;
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\AnonRateLimit;
+use OCP\AppFramework\Http\Attribute\BruteForceProtection;
+use OCP\AppFramework\Http\Attribute\FrontpageRoute;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
+use OCP\AppFramework\Http\Attribute\PublicPage;
+use OCP\AppFramework\Http\Attribute\UserRateLimit;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\Http\JSONResponse;
+use OCP\AppFramework\Http\RedirectResponse;
+use OCP\AppFramework\Http\TemplateResponse;
+use OCP\AppFramework\Services\IInitialState;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Defaults;
+use OCP\Http\Client\IClientService;
+use OCP\IL10N;
+use OCP\IRequest;
+use OCP\IURLGenerator;
+use OCP\IUserSession;
+use OCP\Mail\IMailer;
+use OCP\Mail\IMessage;
+use OCP\OCM\Exceptions\OCMProviderException;
+use OCP\OCM\Exceptions\OCMRequestException;
+use OCP\OCM\IOCMDiscoveryService;
+use OCP\Util;
+use Psr\Log\LoggerInterface;
+use Sabre\DAV\UUIDUtil;
+use Throwable;
+
+/**
+ * Controller for federated invites related routes.
+ *
+ */
+
+class FederatedInvitesController extends Controller {
+	public function __construct(
+		IRequest $request,
+		private AddressHandler $addressHandler,
+		private Defaults $defaults,
+		private FederatedInviteMapper $federatedInviteMapper,
+		private FederatedInvitesService $federatedInvitesService,
+		private IClientService $httpClient,
+		private IInitialState $initialState,
+		private IMailer $mailer,
+		private IOCMDiscoveryService $discovery,
+		private IUserSession $userSession,
+		private WayfProvider $wayfProvider,
+		private ITimeFactory $timeFactory,
+		private IL10N $il10,
+		private IURLGenerator $urlGenerator,
+		private LoggerInterface $logger,
+	) {
+		parent::__construct(Application::APP_ID, $request);
+	}
+
+	/**
+	 * Returns all open (not yet accepted) invites.
+	 *
+	 * @return JSONResponse
+	 */
+	#[NoAdminRequired]
+	#[FrontpageRoute(verb: 'GET', url: '/ocm/invitations')]
+	public function getInvites(): JSONResponse {
+		if (($disabled = $this->requireOcmInvitesEnabled()) !== null) {
+			return $disabled;
+		}
+
+		$uid = $this->userSession->getUser()->getUID();
+		try {
+			$_invites = $this->federatedInviteMapper->findOpenInvitesByUid($uid);
+			$invites = [];
+			foreach ($_invites as $invite) {
+				if ($invite instanceof FederatedInvite) {
+					array_push(
+						$invites,
+						$invite->jsonSerialize()
+					);
+				}
+			}
+			return new JSONResponse($invites, Http::STATUS_OK);
+		} catch (Exception $e) {
+			$this->logger->error("An unexpected error occurred loading invites for user with uid=$uid. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+			return new JSONResponse([
+				'code' => 'ocm_invites_fetch_failed',
+				'message' => 'Could not load invites.',
+			], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
+	/**
+	 * Deletes the invite with the specified token.
+	 *
+	 * @param string $token the token of the invite to delete
+	 * @return JSONResponse with data signature ['token' | 'message'] - the token of the deleted invitation or an error message in case of error
+	 */
+	#[NoAdminRequired]
+	#[FrontpageRoute(verb: 'DELETE', url: '/ocm/invitations/{token}')]
+	public function deleteInvite(string $token): JSONResponse {
+		if (($disabled = $this->requireOcmInvitesEnabled()) !== null) {
+			return $disabled;
+		}
+
+		$uid = $this->userSession->getUser()->getUID();
+		try {
+			$invite = $this->federatedInviteMapper->findInviteByTokenAndUid($token, $uid);
+			$this->federatedInviteMapper->delete($invite);
+			return new JSONResponse(['token' => $token], Http::STATUS_OK);
+		} catch (DoesNotExistException $e) {
+			$this->logger->warning("Could not find invite with token=$token for user with uid=$uid", ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'Invite not found'], Http::STATUS_NOT_FOUND);
+		} catch (Exception $e) {
+			$this->logger->error("An unexpected error occurred deleting invite with token=$token. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'An unexpected error occurred trying to delete the invite'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
+	/**
+	 * Results in displaying the invite accept dialog upon following the invite link
+	 * by redirecting to the index page with the required invite accept dialog parameters.
+	 *
+	 * @param string $token
+	 * @param string $providerDomain
+	 * @return RedirectResponse
+	 */
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	#[FrontpageRoute(verb: 'GET', url: FederatedInvitesService::OCM_INVITE_ACCEPT_DIALOG_ROUTE)]
+	public function inviteAcceptDialog(string $token = '', string $providerDomain = ''): RedirectResponse {
+		return new RedirectResponse($this->urlGenerator->linkToRoute('contacts.page.index', [
+			'token' => $token,
+			'providerDomain' => $providerDomain,
+		]));
+	}
+
+	/**
+	 * Creates an invitation to exchange contact info with the remote user.
+	 *
+	 * @param string $email the recipient email address to send the invitation to (optional)
+	 * @param string $message the optional message to send with the invitation
+	 * @return JSONResponse with data signature ['invite' | 'message'] - the invite url or an error message in case of error.
+	 */
+	#[NoAdminRequired]
+	#[UserRateLimit(limit: 60, period: 3600)]
+	#[BruteForceProtection(action: 'ocmInviteCreate')]
+	#[FrontpageRoute(verb: 'POST', url: '/ocm/invitations')]
+	public function createInvite(string $email = '', string $message = ''): JSONResponse {
+		if (($disabled = $this->requireOcmInvitesEnabled()) !== null) {
+			return $disabled;
+		}
+
+		// Enforce email required when optional mail is disabled
+		if (empty($email) && !$this->federatedInvitesService->isOptionalMailEnabled()) {
+			return new JSONResponse(['message' => $this->il10->t('Email address is required.')], Http::STATUS_BAD_REQUEST);
+		}
+
+		$uid = $this->userSession->getUser()->getUID();
+		if (!empty($email)) {
+			$validationError = $this->validateEmail($email);
+			if ($validationError !== null) {
+				return $validationError;
+			}
+
+			$this->cleanupSupersededInvitesForRecipient($uid, $email);
+
+			// check for existing open invite for the specified email, only if email provided
+			$existingInvites = $this->federatedInviteMapper->findOpenInvitesByRecipientEmail(
+				$uid,
+				$email,
+			);
+			if (count($existingInvites) > 0) {
+				$this->logger->error("An open invite already exists for user with uid $uid and for recipient email $email", ['app' => Application::APP_ID]);
+				return new JSONResponse(['message' => $this->il10->t('An open invite already exists.')], Http::STATUS_CONFLICT);
+			}
+		}
+
+		$invite = new FederatedInvite();
+		$invite->setUserId($uid);
+		$token = UUIDUtil::getUUID();
+		$invite->setToken($token);
+		// created-/expiredAt in seconds
+		$invite->setCreatedAt($this->timeFactory->now()->getTimestamp());
+		$invite->setExpiredAt($this->federatedInvitesService->getInviteExpirationDate($invite->getCreatedAt()));
+		if (!empty($email)) {
+			$invite->setRecipientEmail($email);
+		}
+		$invite->setAccepted(false);
+		$inserted = false;
+		try {
+			$this->federatedInviteMapper->insert($invite);
+			$inserted = true;
+		} catch (Throwable $e) {
+			if ($this->isDuplicateConstraintException($e)) {
+				if (!empty($email) && $this->cleanupSupersededInvitesForRecipient($uid, $email) > 0) {
+					try {
+						$this->federatedInviteMapper->insert($invite);
+						$inserted = true;
+					} catch (Throwable $retry) {
+						if (!$this->isDuplicateConstraintException($retry)) {
+							$this->logger->error('An unexpected error occurred saving a new invite after stale cleanup. Stacktrace: ' . $retry->getTraceAsString(), ['app' => Application::APP_ID]);
+							return new JSONResponse(['message' => 'An unexpected error occurred creating the invite.'], Http::STATUS_INTERNAL_SERVER_ERROR);
+						}
+					}
+				}
+				if (!$inserted) {
+					return new JSONResponse(['message' => $this->il10->t('An open invite already exists.')], Http::STATUS_CONFLICT);
+				}
+			} else {
+				$this->logger->error('An unexpected error occurred saving a new invite. Stacktrace: ' . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+				return new JSONResponse(['message' => 'An unexpected error occurred creating the invite.'], Http::STATUS_INTERNAL_SERVER_ERROR);
+			}
+		}
+		$senderProvider = $this->federatedInvitesService->getProviderFQDN();
+
+		// Only send email if email address provided
+		if (!empty($email)) {
+			/** @var JSONResponse */
+			$response = $this->sendInvitationEmail($token, $senderProvider, $email, $message);
+			if ($response->getStatus() !== Http::STATUS_OK) {
+				// delete invite in case sending the email has failed
+				try {
+					$this->federatedInviteMapper->delete($invite);
+				} catch (Exception $e) {
+					$this->logger->error("An unexpected error occurred deleting invite with token $token. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+					return new JSONResponse(['message' => 'An unexpected error occurred creating the invite.'], Http::STATUS_INTERNAL_SERVER_ERROR);
+				}
+				return $response;
+			}
+		}
+
+		// invite url, use token instead of email for routing
+		$inviteUrl = $this->urlGenerator->getAbsoluteURL(
+			$this->urlGenerator->linkToRoute('contacts.page.index') . 'ocm-invites/' . $token
+		);
+		return new JSONResponse(['invite' => $inviteUrl], Http::STATUS_OK);
+	}
+
+	/**
+	 * Accepts the invite and creates a new contact from the inviter.
+	 * On success the user is redirected to the new contact url.
+	 *
+	 * @param string $token the token of the invite
+	 * @param string $provider the provider of the sender of the invite
+	 * @return JSONResponse with data signature ['contact' | 'message'] - the new contact url or an error message in case of error
+	 */
+	#[NoAdminRequired]
+	#[UserRateLimit(limit: 60, period: 3600)]
+	#[BruteForceProtection(action: 'ocmInviteAccept')]
+	#[FrontpageRoute(verb: 'PATCH', url: '/ocm/invitations/{token}/accept')]
+	public function acceptInvite(string $token = '', string $provider = ''): JSONResponse {
+		if (($disabled = $this->requireOcmInvitesEnabled()) !== null) {
+			return $disabled;
+		}
+
+		if ($token === '' || $provider === '') {
+			$this->logger->error("Both token and provider must be specified. Received: token=$token, provider=$provider", ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'Both invite code and provider must be specified.'], Http::STATUS_BAD_REQUEST);
+		}
+		$localUser = $this->userSession->getUser();
+		if ($localUser === null) {
+			return new JSONResponse(['message' => $this->il10->t('Could not accept invite because no authenticated user was found.')], Http::STATUS_UNAUTHORIZED);
+		}
+		$provider = $this->normalizeProviderBase($provider);
+		if ($provider === null) {
+			return new JSONResponse(['message' => $this->il10->t('The invite provider is invalid or not allowed.')], Http::STATUS_BAD_REQUEST);
+		}
+		$recipientProviderFqdn = $this->federatedInvitesService->getProviderFQDN();
+		// do not accept an invite from this instance
+		$senderProviderFqdn = parse_url($provider)['host'];
+		$this->logger->debug("acceptInvite() - $recipientProviderFqdn === $senderProviderFqdn");
+		if ($recipientProviderFqdn === $senderProviderFqdn) {
+			return new JSONResponse(['message' => $this->il10->t('Unable to accept this invitation. You are trying to accept an outgoing invitation.')], Http::STATUS_BAD_REQUEST);
+		}
+		$userId = $localUser->getUID();
+		$email = $localUser->getEMailAddress();
+		$name = $localUser->getDisplayName();
+		if ($recipientProviderFqdn === '' || $userId === '' || $email === '' || $name === '') {
+			$this->logger->error("All of these must be set: recipientProvider: $recipientProviderFqdn, email: $email, userId: $userId, name: $name", ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'Could not accept invite, user data is incomplete.'], Http::STATUS_UNPROCESSABLE_ENTITY);
+		}
+		$cloudId = '';
+		try {
+			// accept the invite by calling provider OCM /invite-accepted
+			// this returns a response with the following data signature: ['userID', 'email', 'name']
+			// @link https://cs3org.github.io/OCM-API/docs.html?branch=v1.1.0&repo=OCM-API&user=cs3org#/paths/~1invite-accepted/post
+			$client = $this->httpClient->newClient();
+			/**
+			 * @var \OCP\OCM\ICapabilityAwareOCMProvider $discovered
+			 *
+			 */
+			$discovered = $this->discovery->discover($provider);
+			if (
+				$discovered->hasCapability('invites')
+				|| $discovered->hasCapability('invite-accepted')
+				|| $discovered->hasCapability('/invite-accepted')
+			) {
+
+				$response = $this->discovery->requestRemoteOcmEndpoint(
+					null,
+					$provider,
+					'/invite-accepted',
+					[
+						'recipientProvider' => $recipientProviderFqdn,
+						'token' => $token,
+						'userID' => $userId,
+						'email' => $email,
+						'name' => $name
+					],
+					'POST',
+					$client
+				);
+				$responseData = $response->getBody();
+				$data = json_decode($responseData, true);
+				if (
+					!is_array($data)
+					|| !isset($data['userID'], $data['email'], $data['name'])
+					|| !is_string($data['userID'])
+					|| !is_string($data['email'])
+					|| !is_string($data['name'])
+					|| trim($data['userID']) === ''
+					|| trim($data['email']) === ''
+					|| trim($data['name']) === ''
+				) {
+					$this->logger->warning('Invalid /invite-accepted payload from provider', [
+						'app' => Application::APP_ID,
+						'provider' => $provider,
+						'payload' => $responseData,
+					]);
+					return new JSONResponse(['message' => $this->il10->t('Could not accept invite because the remote provider returned an invalid response.')], Http::STATUS_BAD_GATEWAY);
+				}
+
+				$cloudId = $data['userID'] . '@' . $this->addressHandler->removeProtocolFromUrl($provider);
+
+				$contactRef = $this->federatedInvitesService->createNewContact(
+					$cloudId,
+					$data['email'],
+					$data['name'],
+					null
+				);
+				if (!isset($contactRef)) {
+					$this->logger->error('Remote invite acceptance succeeded but local contact creation failed', [
+						'app' => Application::APP_ID,
+						'token' => $token,
+						'provider' => $provider,
+						'cloudId' => $cloudId,
+						'userId' => $userId,
+					]);
+					return new JSONResponse([
+						'code' => 'ocm_invite_local_contact_create_failed',
+						'message' => $this->il10->t('The remote provider accepted the invite, but this server could not create the local contact.'),
+					], Http::STATUS_BAD_GATEWAY);
+				}
+				$key = base64_encode($contactRef);
+				$contactUrl = $this->urlGenerator->getAbsoluteURL(
+					$this->urlGenerator->linkToRoute('contacts.page.index') . $this->il10->t('All contacts') . '/' . $key
+				);
+				return new JSONResponse(['contact' => $contactUrl], Http::STATUS_OK);
+			} else {
+				$this->logger->error('Provider: ' . $provider . ' does not support invites.', ['app' => Application::APP_ID]);
+				return new JSONResponse(['message' => 'Provider: ' . $provider . ' does not support invites.'], Http::STATUS_BAD_REQUEST);
+			}
+		} catch (ContactAlreadyExistsException $e) {
+			return new JSONResponse(['message' => 'Contact with cloudID ' . $cloudId . ' already exists.'], Http::STATUS_CONFLICT);
+		} catch (RequestException $e) { // this should catch OCM API request exceptions
+			$this->logger->error('/invite-accepted returned an error: ' . $e->getMessage(), ['app' => Application::APP_ID]);
+			/**
+			 * 400: Invalid or non existing token
+			 * 409: Invite already accepted
+			 */
+			$statusCode = $e->getResponse() !== null
+				? $e->getResponse()->getStatusCode()
+				: null;
+			switch ($statusCode) {
+				case Http::STATUS_BAD_REQUEST:
+					return new JSONResponse(['message' => 'Invalid, non-existing, or expired invite code.'], Http::STATUS_BAD_REQUEST);
+				case Http::STATUS_CONFLICT:
+					return new JSONResponse(['message' => 'Invite already accepted'], Http::STATUS_CONFLICT);
+			}
+			$this->logger->error("An unexpected error occurred accepting invite with token=$token and provider=$provider. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'An unexpected error occurred trying to accept invite.'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		} catch (OCMProviderException|OCMRequestException|Exception $e) {
+			$this->logger->error("An unexpected error occurred accepting invite with token=$token and provider=$provider. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'An unexpected error occurred trying to accept invite'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
+	/**
+	 * Re-sends an existing invite email while preserving invite lifetime metadata.
+	 *
+	 */
+	#[NoAdminRequired]
+	#[UserRateLimit(limit: 30, period: 3600)]
+	#[BruteForceProtection(action: 'ocmInviteResend')]
+	#[FrontpageRoute(verb: 'PATCH', url: '/ocm/invitations/{token}/resend')]
+	public function resendInvite(string $token): JSONResponse {
+		if (($disabled = $this->requireOcmInvitesEnabled()) !== null) {
+			return $disabled;
+		}
+
+		$uid = $this->userSession->getUser()->getUID();
+		try {
+			$invite = $this->federatedInviteMapper->findInviteByTokenAndUid($token, $uid);
+		} catch (DoesNotExistException $e) {
+			$this->logger->warning("Could not find invite with token=$token for user with uid=$uid", ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'Invite not found'], Http::STATUS_NOT_FOUND);
+		} catch (Exception $e) {
+			$this->logger->error("An unexpected error occurred loading invite with token=$token. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'An unexpected error occurred trying to resend the invite'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		// Cannot resend if no email address
+		if (empty($invite->getRecipientEmail())) {
+			return new JSONResponse(['message' => $this->il10->t('Cannot resend: no email address')], Http::STATUS_UNPROCESSABLE_ENTITY);
+		}
+		if ($this->isInviteAccepted($invite)) {
+			return new JSONResponse([
+				'code' => 'ocm_invite_already_accepted',
+				'message' => $this->il10->t('Invite has already been accepted.'),
+			], Http::STATUS_CONFLICT);
+		}
+		if ($this->isInviteExpired($invite)) {
+			return new JSONResponse([
+				'code' => 'ocm_invite_expired',
+				'message' => $this->il10->t('Invite has expired. Please create a new one.'),
+			], Http::STATUS_GONE);
+		}
+
+		$sendDate = date('Y-m-d', $invite->getCreatedAt());
+		$initiatorDisplayName = $this->userSession->getUser()->getDisplayName();
+		// a resend notification that refers to the previously sent invite
+		$message = $this->il10->t(
+			'This is a copy of an invite sent to you previously by %1$s on %2$s',
+			[
+				$initiatorDisplayName,
+				$sendDate
+			]
+		);
+		$senderProvider = $this->federatedInvitesService->getProviderFQDN();
+		/** @var JSONResponse */
+		$response = $this->sendInvitationEmail($token, $senderProvider, $invite->getRecipientEmail(), $message);
+		if ($response->getStatus() !== Http::STATUS_OK) {
+			$this->logger->error("An unexpected error occurred resending the invite with token $token. HTTP response status: " . $response->getStatus(), ['app' => Application::APP_ID]);
+			return $response;
+		}
+
+		// the invite url, use token instead of email for routing
+		$inviteUrl = $this->urlGenerator->getAbsoluteURL(
+			$this->urlGenerator->linkToRoute('contacts.page.index') . 'ocm-invites/' . $invite->getToken()
+		);
+		return new JSONResponse(['invite' => $inviteUrl], Http::STATUS_OK);
+	}
+
+	/**
+	 * Attaches a recipient email to an existing link-only invite and sends the
+	 * invitation email. Refreshes the creation and expiration timestamps so the
+	 * recipient receives a fresh expiry window. Reverts both the email and the
+	 * timestamps if the mailer fails, so a failed call leaves the invite as it
+	 * was before.
+	 *
+	 * @param string $token the invite token
+	 * @param string $email the recipient email address
+	 * @param string $message the optional message to include in the email
+	 * @return JSONResponse the serialized invite on success or an error message
+	 */
+	#[NoAdminRequired]
+	#[UserRateLimit(limit: 30, period: 3600)]
+	#[BruteForceProtection(action: 'ocmInviteAttachEmail')]
+	#[FrontpageRoute(verb: 'PATCH', url: '/ocm/invitations/{token}/email')]
+	public function attachEmailAndSend(string $token, string $email = '', string $message = ''): JSONResponse {
+		if (($disabled = $this->requireOcmInvitesEnabled()) !== null) {
+			return $disabled;
+		}
+
+		$uid = $this->userSession->getUser()->getUID();
+		try {
+			$invite = $this->federatedInviteMapper->findInviteByTokenAndUid($token, $uid);
+		} catch (DoesNotExistException $e) {
+			$this->logger->warning("Could not find invite with token=$token for user with uid=$uid", ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'Invite not found'], Http::STATUS_NOT_FOUND);
+		} catch (Exception $e) {
+			$this->logger->error("An unexpected error occurred loading invite with token=$token. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'An unexpected error occurred attaching the email.'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		if ($this->isInviteAccepted($invite)) {
+			return new JSONResponse([
+				'code' => 'ocm_invite_already_accepted',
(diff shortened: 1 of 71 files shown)


redblom:
General layout plan:
 - dedicated button to send an invite to a remote user - below 'New contact' button
 - dedicated button on cloud ID component for existing remote contacts that have no cloud ID set
 - new sent invites will be displayed in the left panel - with a revoke option

Status:
 - invite accept dialog route implemented

redblom:
> Thanks for the contribution, @redblom!
> 
> > The features relies on an implementation of the Invitation Workflow (see https://github.com/sara-nl/nc-collaboration).
> 
> Is there any way this could work without introducing an app dependency?

Short answer: No, that is to say, the actual cloud ID exchange is not a responsibility of the contacts app but the contacts app (the cloud ID field) is a logical place where to have this option (as per earlier discussion with the Design Team).
The feature actually depends on the implementation of OCM `/invite-accepted`. This is currently in the process of being implemented for nextcloud server core (https://github.com/nextcloud/server/pull/51113) so cloud ID exchange can become an integral part of nextcloud. Up until then there should be an app present implementing this.
The idea is to make the option not visible to the user when the platform can not do the actual cloud ID exchange (by configuration).

ChristophWurst:
Thanks for the clarification

redblom:
Do we want to 'cc' the invitation email to the sender as well (by default or configurable) ?
Eg. in case the receiver loses the invitation mail it would allow the sender to simply resend the invitation from the email client instead of having to create a new one (and revoke the old one) from the contacts app.
\* Another (possible future) option would be a resend button on the invitation details page.

mickenordin:
> Do we want to 'cc' the invitation email to the sender as well (by default or configurable) ? Eg. in case the receiver loses the invitation mail it would allow the sender to simply resend the invitation from the email client instead of having to create a new one (and revoke the old one) from the contacts app. * Another (possible future) option would be a resend button on the invitation details page.

I think a "resend button" makes sense, maybe along with a check box that is unchecked by default to send the invitation to the sender as well.

redblom:
> > Do we want to 'cc' the invitation email to the sender as well (by default or configurable) ? Eg. in case the receiver loses the invitation mail it would allow the sender to simply resend the invitation from the email client instead of having to create a new one (and revoke the old one) from the contacts app. * Another (possible future) option would be a resend button on the invitation details page.
> 
> I think a "resend button" makes sense, maybe along with a check box that is unchecked by default to send the invitation to the sender as well.

We do not persist the optional personal message with the invite. So that's lost for a resend. But in order to correlate to the previously sent invite, what we can do is to use a new message like:
`"This is a copy of an invite send to you previously by $initiatorDisplayName on $sendDate"`

Fwiw I prefer this to persisting the message.

mickenordin:
Adding some screenshots here to show you what it looks like, this is the dialog for adding a new invitation:
2025-09-25T16:01:24,021984991+02:00
This is what it looks like after it is sent:
2025-09-25T16:01:40,518818898+02:00
This is the default Where-Are-You-From page, when no federation has been configured
2025-09-25T16:05:10,821218256+02:00
This is what the WAYF-page looks like with a federation configured:
2025-09-25T16:06:40,570032364+02:00
It is also possible to add multiple federations:
2025-09-25T16:14:34,777039486+02:00
This is what the type-to-search feature looks like:
2025-09-25T16:14:50,748780755+02:00

mickenordin:
> Thanks for the contribution, @redblom!
> 
> > The features relies on an implementation of the Invitation Workflow (see https://github.com/sara-nl/nc-collaboration).
> 
> Is there any way this could work without introducing an app dependency?

All the necessary plumbing is in Nextcloud 32.0.0.0rc4 so no external apps needed anymore.

redblom:
@ArtificialOwl could you give it a go, the review.

lib/WayfProvider.php
```diff
@@ -0,0 +1,120 @@
...
+	/**
+	 * Returns all mesh providers from cache if possible.
+	 *
+	 * @return array an array containing all mesh providers
+	 */
+
+
+	public function getMeshProvidersFromCache(): array {
+		$json = $this->appConfig->getValueString(Application::APP_ID, 'federations_cache');
+		$data = json_decode($json, true);
```
ArtificialOwl: you can use getValueArray() / setValueArray() and set the config value as `lazy`

lib/Service/FederatedInvitesService.php
```diff
@@ -0,0 +1,57 @@
...
+		private IAppConfig $appConfig,
+		private IClientService $httpClient,
+		private IURLGenerator $urlGenerator,
+		private LoggerInterface $logger,
+	) {
+	}
+
+	public function isOcmInvitesEnabled():bool {
+		$val = $this->appConfig->getValueBool(Application::APP_ID, 'ocm_invites_enabled', FederatedInvitesService::OCM_INVITES_ENABLED_BY_DEFAULT);
+		$boolval = (is_string($val) ? filter_var($val, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) : (bool)$val);
```
ArtificialOwl: `getValueBool()` will always returns `boolean`

lib/Db/FederatedInviteMapper.php
```diff
@@ -0,0 +1,72 @@
...
+
+use OCA\CloudFederationAPI\Db\FederatedInviteMapper as DbFederatedInviteMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @template-extends QBMapper<FederatedInvite>
+ */
+class FederatedInviteMapper extends DbFederatedInviteMapper {
```
ArtificialOwl: not a huge fan of this,
if the DB is created and managed from cloud_federation_api, could this request be also integrated to the app, and calling an API from the cloud_federation_api to select/update/delete data from/to database ?

ArtificialOwl (commented):
quick review, not fully complete, but already some elements to rework

redblom:
@ArtificialOwl modified to work with Nextcloud 33 OCM endpoint request event.

redblom:
List of modifications - basically moving everything around the `invite-accepted` capability to this app:
 - ~~move implementation of `invite-accepted` capability to this app~~ - done
 - ~~move registering `invite-accepted` capability to this app~~ - done
 - ~~move setting/getting InviteAcceptDialog property to this app~~ - done
 - ~~rename `lib/Migration/Version1016Date202502262004.php`~~ - done

redblom:
> * [ ]  Ping me and I'll enable the workflows

@miaulalala - ping !

redblom:
@miaulalala - What's wrong with the sign-off, maybe co-author is not allowed ?

miaulalala:
> @miaulalala - What's wrong with the sign-off, maybe co-author is not allowed ?

No that's not the issue, one setting has Anton P., the other Anton P (no dot) 😅

redblom:
> > @miaulalala - What's wrong with the sign-off, maybe co-author is not allowed ?
> 
> No that's not the issue, one setting has Anton P., the other Anton P (no dot) 😅

Pfff :flushed:

appinfo/routes.php
```diff
@@ -5,10 +5,22 @@
...
  */
 
+use OCA\Contacts\Service\FederatedInvitesService;
+
 return [
 	'routes' => [
 		['name' => 'contacts#direct', 'url' => '/direct/contact/{contact}', 'verb' => 'GET'],
 		['name' => 'contacts#directcircle', 'url' => '/direct/circle/{singleId}', 'verb' => 'GET'],
+
+		['name' => 'federated_invites#get_invites', 'url' => '/ocm/invitations', 'verb' => 'GET'],
```
artonge, line 15: Those can be replaced by the following attribute. See suggestion in `FederatedInvitesController`

```php
	#[Route(type: Route::TYPE_FRONTPAGE, verb: 'GET', url: '/ocm/invitations')]
```

lib/Controller/FederatedInvitesController.php
```diff
@@ -0,0 +1,485 @@
...
+		);
+	}
+
+	/**
+	 * Returns all open (not yet accepted) invites.
+	 *
+	 * @return JSONResponse
+	 */
+	#[NoAdminRequired]
+	public function getInvites(): JSONResponse {
```
artonge, line 107:
```suggestion
	#[Route(type: Route::TYPE_FRONTPAGE, verb: 'GET', url: '/ocm/invitations')]
	public function getInvites(): JSONResponse {
```

lib/Controller/FederatedInvitesController.php
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	/**
+	 * Deletes the invite with the specified token.
+	 *
+	 * @param string $token the token of the invite to delete
+	 * @return JSONResponse with data signature ['token' | 'message'] - the token of the deleted invitation or an error message in case of error
+	 */
+	#[NoAdminRequired]
+	public function deleteInvite(string $token): JSONResponse {
```
artonge, line 128: And so on

```suggestion
	#[Route(type: Route::TYPE_FRONTPAGE, verb: 'DELETE', url: '/ocm/invitations/{token}')]
	public function deleteInvite(string $token): JSONResponse {
```

lib/Command/DisableOcmInvites.php
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$isEnabled = $this->appConfig->getValueBool('contacts', 'ocm_invites_enabled');
+		if (!$isEnabled) {
+			$output->writeln('OCM Invites already disabled.');
+			return self::SUCCESS;
+		}
+
+		$this->appConfig->setValueBool('contacts', 'ocm_invites_enabled', false);
```
artonge, line 36: Let's introduce a lexicon in Contacts so that new settings are tracked.

Documentation: https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/config/lexicon.html
Example: https://github.com/nextcloud-gmbh/server/blob/04172b855de465bc661034b32724ca394ffee3fe/apps/settings/lib/ConfigLexicon.php#L22

```suggestion
		$this->appConfig->setValueBool(App::ID, 'ocm_invites_enabled', false);
```

lib/Command/DisableOcmInvites.php
```diff
@@ -0,0 +1,41 @@
...
+		$this->appConfig->deleteKey('core', ConfigLexicon::OCM_INVITE_ACCEPT_DIALOG);
```
artonge, line 37: This is a clean up of the old key? If so this should go into a repair step: https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/repair.html

lib/Command/EnableOcmInvites.php
```diff
@@ -0,0 +1,42 @@
...
+class EnableOcmInvites extends Command {
+	public function __construct(
+		protected IAppConfig $appConfig,
+	) {
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('contacts:enable-ocm-invites')
```
artonge, line 24: Overall, this is nice for discoverability to have those commands, but they could be replaced by `occ config:app:set contacts ocm_invites_enabled --value false/true --type boolean`. I would consider removing them, or making it a single command like the `occ maintenance:mode` command.

lib/Cron/UpdateOcmProviders.php
```diff
@@ -0,0 +1,36 @@
...
+
+use OCA\Contacts\AppInfo\Application;
+use OCA\Contacts\WayfProvider;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\TimedJob;
+use OCP\IAppConfig;
+
+class UpdateOcmProviders extends TimedJob {
+	// Run every five minutes
+	private int $expire_time = 5 * 60;
```
artonge, line 20: On which occasion will the data change? I am wondering if once a day would be enough. Or even let the admin trigger a command for that?

lib/Migration/Version8004Date20260130131217.php
artonge, line 1: Is the schema different from the old table? If not, could be we simply reuse the old table instead of dropping it?

lib/Controller/FederatedInvitesController.php
```diff
@@ -0,0 +1,485 @@
...
+	 */
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	public function inviteAcceptDialog(string $token = '', string $providerDomain = ''): TemplateResponse {
+		$this->initialState->provideInitialState('inviteToken', $token);
+		$this->initialState->provideInitialState('inviteProvider', $providerDomain);
+		$this->initialState->provideInitialState('acceptInviteDialogUrl', FederatedInvitesService::OCM_INVITE_ACCEPT_DIALOG_ROUTE);
+
+		return $this->index();
+	}
```
artonge, line 158: I would not extend the `PageController` just for that. Maybe simply redirect to the index, and add some query parameters to the URL?

lib/Controller/FederatedInvitesController.php
```diff
@@ -0,0 +1,485 @@
...
+			return new JSONResponse(['message' => 'An unexpected error occurred trying to accept invite.'], Http::STATUS_NOT_FOUND);
+		} catch (OCMProviderException|OCMRequestException|Exception $e) {
+			$this->logger->error("An unexpected error occurred accepting invite with token=$token and provider=$provider. Stacktrace: " . $e->getTraceAsString(), ['app' => Application::APP_ID]);
+			return new JSONResponse(['message' => 'An unexpected error occurred trying to accept invite'], Http::STATUS_NOT_FOUND);
+		}
+	}
+
+	/**
+	 * Resets the creation and expiration dates, and sends a new invite to the recipient.
+	 *
```
artonge, line 319:
```suggestion
```

src/components/Ocm/OcmAcceptForm.vue
```diff
@@ -0,0 +1,129 @@
...
+    return {
+      invite: "",
+      error: "",
+    };
+  },
+  methods: {
+    parseInvite(str) {
+      function tryParse(s) {
+        const idx = s.lastIndexOf("@");
+        if (idx === -1) return null;
```
artonge, line 66: Please do not save a few characters at the expense of readability.

```suggestion
        if (idx === -1) {
            return null;
        }
```

artonge: Same applies to the rest of the code.

src/css/wayf.scss
artonge, line 1: Can't we have that in `Wayf.vue`?

src/store/ocminvites.js
```diff
@@ -0,0 +1,123 @@
...
+const actions = {
+    fetchOcmInvites(context) {
+		axios.get(generateUrl('/apps/contacts/ocm/invitations'))
+			.then(response => {
+				context.commit('appendInvites', response.data)
+				context.commit('sortInvites', response.data)
+			})
+			.catch((error) => {
+				logger.error('Error fetching OCM invites: ' + error)
+			})
```
artonge, line 37: We usually try to use `try/catch` nowadays

src/models/ocminvite.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * SPDX-FileCopyrightText: 2018 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export default class OcmInvite {
```
artonge, line 8: Any reason to wrap the data in a class?

src/store/ocminvites.js
artonge, line 1: Let's use typescript as much as possible.

src/wayf.js
```diff
@@ -0,0 +1,23 @@
...
+import { loadState } from "@nextcloud/initial-state";
+import { translate as t, translatePlural as n } from "@nextcloud/l10n";
+import Wayf from "./components/Ocm/Wayf.vue";
+
+if (!document.body.id) {
+  document.body.id = 'body-public';
+}
+
+
+document.addEventListener("DOMContentLoaded", () => {
```
artonge, line 16: I don't think you need to wrap that code in a listener.

src/wayf.js
```diff
@@ -0,0 +1,23 @@
...
+  const props = loadState("contacts", "wayf");
+  const app = createApp(Wayf, props);
+  app.config.globalProperties.t = t;
+  app.config.globalProperties.n = n;
+  app.mount("#contacts-wayf");
+});
+import './css/wayf.scss'
```
artonge, line 23: Let's move imports at the top of the file.

src/wayf.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+import { createApp } from "vue";
+import { loadState } from "@nextcloud/initial-state";
+import { translate as t, translatePlural as n } from "@nextcloud/l10n";
+import Wayf from "./components/Ocm/Wayf.vue";
+
+if (!document.body.id) {
+  document.body.id = 'body-public';
+}
```
artonge, line 13: Why is this needed?

src/models/constants.ts
```diff
@@ -130,13 +135,13 @@ export const CIRCLES_MEMBER_GROUPING = [
...
-		share: ShareType.Remote,
-		type: MEMBER_TYPE_USER,
-	},
+	// TODO: implement federated
+	// {
+	// id: `picker-${ShareType.Remote}`,
+	// label: t('contacts', 'federated users'),
+	// share: ShareType.Remote,
+	// type: MEMBER_TYPE_USER
+	// },
```
artonge, line 144: This is not working anymore?

lib/Listener/OcmDiscoveryListener.php
```diff
@@ -0,0 +1,43 @@
...
+	 * This handler will register the capability invite-accepted
+	 * and set the invite accept dialog url.
+	 *
+	 * @param Event $event an event of type LocalOCMDiscoveryEvent
+	 * @return void
+	 */
+	public function handle(Event $event): void {
+		if ($event instanceof LocalOCMDiscoveryEvent) {
+			$event->addCapability('invite-accepted');
+			$inviteAcceptDialog = $this->appConfig->getValueString('core', ConfigLexicon::OCM_INVITE_ACCEPT_DIALOG);
```
artonge, line 37: Can't this be a Contacts app settings instead of consuming the private one?

lib/Service/FederatedInvitesService.php
```diff
@@ -0,0 +1,183 @@
...
+		private IUserManager $userManager,
+		private IUserSession $userSession,
+		private FederatedInviteMapper $federatedInviteMapper,
+		private LoggerInterface $logger,
+		private SocialApiService $socialApiService,
+	) {
+	}
+
+	public function isOcmInvitesEnabled():bool {
+		return $this->appConfig->getValueBool(Application::APP_ID, 'ocm_invites_enabled', FederatedInvitesService::OCM_INVITES_ENABLED_BY_DEFAULT);
```
artonge, line 49: Let's use a lexicon.

lib/Service/SocialApiService.php
```diff
@@ -225,6 +229,64 @@ public function updateContact(string $addressbookId, string $contactId, ?string
...
+			foreach ($addressBooks as $_addressBook) {
+				// TODO properly resolve the correct addressbook to add the contact to
+				// Resolve by uri seems a bit risky ... can we be sure the uri equals 'contacts' ?
+				// Perhaps add to the first 'non system' addressbook we find ?
+				// (although we still would like to add to the 'Contacts' addressbook I guess)
+				if ($_addressBook->getUri() === 'contacts') {
+					$addressBook = $_addressBook;
+					break;
+				}
+			}
```
artonge, line 267: Maybe @ChristophWurst can give us the better way of doing this

lib/WayfProvider.php
```diff
@@ -0,0 +1,125 @@
...
+	) {
+	}
+
+	/**
+	 * Returns all mesh providers.
+	 *
+	 * @return array an array containing all mesh providers
+	 */
+	public function getMeshProviders(): array {
+		$urls = preg_split('/\s+/', trim($this->appConfig->getValueString(Application::APP_ID, 'mesh_providers_service')));
```
artonge, line 35: Lexicon

lib/WayfProvider.php
```diff
@@ -0,0 +1,125 @@
...
+		return $federations;
+	}
+
+	/**
+	 * Returns all mesh providers from cache if possible.
+	 *
+	 * @return array an array containing all mesh providers
+	 */
+	public function getMeshProvidersFromCache(): array {
+		$data = $this->appConfig->getValueArray(Application::APP_ID, 'federations_cache', [], true);
```
artonge, line 93: Lexicon

lib/Cron/UpdateOcmProviders.php
artonge: Or could this be done during the Wayf request?

miaulalala:
@redblom - can you check on artonge's review comments please?

redblom:
> @redblom - can you check on artonge's review comments please?

:+1:  working on it.

lib/Cron/UpdateOcmProviders.php
redblom: > Or could this be done during the Wayf request?

The idea was to not do that. It's the reason to have the cache.

Further, an admin trigger would require the admin to be notified, meaning no guarantees there.
But yes we're looking for some sane default value ...

redblom:
> @redblom - can you check on artonge's review comments please?

@miaulalala We've now covered them all we think.

lib/Service/SocialApiService.php
redblom: Addressed in the latest push.

lib/Listener/OcmDiscoveryListener.php
redblom: Addressed in the latest push.

lib/Migration/Version8004Date20260130131217.php
redblom: Addressed in https://github.com/nextcloud/server/pull/57853, see https://github.com/nextcloud/server/pull/57853#discussion_r3173052270

redblom:
Rebased and some cleanup.

miaulalala:
Hi @redblom! CI is currently failing on two fronts:

**1. Missing `#[Override]` attributes (Psalm)**
Multiple methods in `Cron/UpdateOcmProviders.php`, the two new Listener classes, and the Migration files are missing `#[Override]` attributes. You can fix most of these automatically by running `composer psalm:update-baseline` and committing the result.

**2. NPM lint error in `src/wayf.js`**
Line 13 uses a top-level arrow function — change it to a regular `function` declaration to satisfy the `antfu/top-level-function` rule, then run `npm run lint:fix`.

lib/Command/OcmInvitesConfig.php
artonge, line 1: Do we really want a command for that? Why is the already existing command to set config values not enough?

If the reason is for documentation, then I think we should drop the custom command and simply extend the documentation.

lib/Cron/UpdateOcmProviders.php
artonge: > The idea was to not do that. It's the reason to have the cache.

Is this because the operation is slow?
You can also [manually register](https://docs.nextcloud.com/server/latest/developer_manual/basics/backgroundjobs.html#registering-manually) a `QueuedJob` for this kind of use case.

artonge:
@redblom there are still some unresolved conversations from the last time I reviewed, can you address them?

lib/Command/OcmInvitesConfig.php
redblom: @artonge - Hmm maybe we misunderstood your suggestion here. A dedicated command (with documentation) seemed useful to us.

But is extending the docs the preferred way then?

redblom:
@artonge 
> Is this because the operation is slow?

Yes, it can be slow-ish, so definitely for a responsive wayf page.

> You can also [manually register](https://docs.nextcloud.com/server/latest/developer_manual/basics/backgroundjobs.html#registering-manually) a `QueuedJob` for this kind of use case.

Will have a look at that.

lib/Controller/FederatedInvitesController.php
redblom: @artonge : you mean like so?

```
class PageController extends Controller {
	...

	public function index(string $token = '', string $providerDomain = ''): TemplateResponse {
		if(isset($token) && isset($providerDomain)) {
			return $this->handleInviteAcceptDialog($token, $providerDomain);
		}
		...
	}

	private function handleInviteAcceptDialog(string $token = '', string $providerDomain = ''): TemplateResponse {
		$this->initialState->provideInitialState('inviteToken', $token);
		$this->initialState->provideInitialState('inviteProvider', $providerDomain);
		$this->initialState->provideInitialState('acceptInviteDialogUrl', FederatedInvitesService::OCM_INVITE_ACCEPT_DIALOG_ROUTE);
		return new TemplateResponse(Application::APP_ID, 'main');
	}
...
```

lib/Cron/UpdateOcmProviders.php
redblom: @artonge - In the docs I read that QueuedJobs are for one time jobs that may be triggered (added/removed) following an event. In this case we are dealing with job that deals with retrieving data (the wayf providers list) from an external service. That service does not emit an event when the list of providers is updated, so we must consult it periodically. We also can't rely on the responsiveness of that service, this is why we cache the retrieved data (the list of providers).
How can QueuedJob help in this usecase?

lib/Command/OcmInvitesConfig.php
artonge: > But is extending the docs the preferred way then?

I would say yes, as in this case this is doing nothing else than setting the config value.

lib/Controller/FederatedInvitesController.php
```diff
@@ -0,0 +1,1066 @@
...
+			return new DataResponse(['error' => 'OCM discovery failed', 'base' => $base], Http::STATUS_NOT_FOUND);
+		}
+	}
+
+	/**
+	 * Accepts the invite and creates a new contact from the inviter.
+	 * On success the user is redirected to the new contact url.
+	 *
+	 * @param string $token the token of the invite
+	 * @param string $provider the provider of the sender of the invite
```
artonge: The argument is not part of the method signature

lib/Controller/FederatedInvitesController.php
```diff
@@ -0,0 +1,1066 @@
...
+	/**
+	 * Sends a copy of the invite email to the sender.
+	 *
+	 * @param string $token the invite token
+	 * @param string $senderProvider this provider
+	 * @param string $senderAddress the sender's email address
+	 * @param string $recipientAddress the original recipient's email address
+	 * @param string $message the optional message included in the invite
+	 */
+	private function sendCcEmail(string $token, string $senderProvider, string $senderAddress, string $recipientAddress, string $message): void {
```
artonge: Isn't the logic really close to the `sendMail` method?
Can't we add a `$cc` argument to `sendMail` to benefit from the work that was already done there?

lib/WayfProvider.php
```diff
@@ -0,0 +1,148 @@
...
+
+		$found = [];
+		foreach ($urls as $url) {
+			if ($url === '') {
+				continue;
+			}
+			try {
+				$res = $this->httpClient->newClient()->get($url);
+				$code = $res->getStatusCode();
+				if (!($code >= 200 && $code < 400)) {
```
artonge:
```suggestion
				if (!($code >= Http::STATUS_OK && $code < ...)) {
```

lib/WayfProvider.php
```diff
@@ -0,0 +1,148 @@
...
+					continue;
+				}
+				$data = json_decode($res->getBody(), true);
```
artonge, line 72: Where is the format of the answer defined?

lib/Controller/FederatedInvitesController.php
```diff
@@ -0,0 +1,1066 @@
...
+			$appManager,
+			$compareVersion,
+			$groupSharingService,
+		);
+	}
+
+	/**
+	 * Returns all open (not yet accepted) invites.
+	 *
+	 * @return JSONResponse
```
artonge, line 82: Phpdoc are not necessary if the native php types are used, you can remove them.

lib/Controller/FederatedInvitesController.php
artonge: Yes, and looking at the variable that you provide:
- `inviteToken` seems to be accessible from the frontend already
- same for `inviteProvider`

So maybe you only need to provide `acceptInviteDialogUrl`?

nimishavijay:
This seems very comprehensive, nice! :) From what I understand this is a way to exchange contacts with people from other organizations directly from the contacts app. Design wise I have a few questions and some feedback:

Thinking about the design and flow itself, some questions come to mind:
- [ ] There seem to be a lot of ways of sending invitations: via email, via invite link, via invite code. Is there a necessity to have an invite code? Could we have just one way: invite link and have options like "copy link" and "share via email" for that one link? It would simplify a lot of decision making for the user. They would only have to click on a link to exchange contacts in every case. Does that make sense?
- [ ] Since the goal is to NOT look at the invites page (there shouldn't be invitations that are sent and pending acceptance, and once you have exchanged contacts there's no reason to look at the invitation again) I am not sure if having an entire view in the main content is the way to go. What do you think about moving it into a modal?
- [ ] Similar reasoning for the buttons too: the combined visual weight of the invitations is even more than of the CTA of the whole app! So what do you think about collapsing all invites into one item called "External invites" and that opens up a modal with the list of invitations and an option to create?
- [ ] instead of having an option to send a copy of the invite to the sender's own address, why not have the invite message shown in the UI? It can be collapsed by default to save space of course

Visual design feedback to be more consistent with Nextcloud design:
- [ ] We want to be accessible, so the headings for the modals/dialogs must be h2. I see that some of them are different like h5. If you want them to look smaller you can override the styles, but it should follow the heading order
- [ ] The buttons in a dialog a usually positioned as: primary action button on the right end and the cancel button on the left end for left-to-right languages. Your buttons seem to be the other way around, which could lead to accidentally clicking on cancel. The primary action button should also always be the primary variant so that the color stands out in the dialog
- [ ] Also try to use existing variables or multiples of existing variables for padding, margin and gaps as much as possible. Instead of 4px, use `var(--default-grid-baseline)` 
- [ ] In the "where are you from" page, consider using the [NcFormbox](https://nextcloud-vue-components.netlify.app/#/Components/NcFormBox) component, it would automatically adjust colours, spacing, etc


But either way this is really cool. Interested to know what you think of the design feedback! :)

redblom:
> But either way this is really cool. Interested to know what you think of the design feedback! :)

@nimishavijay - Thank you for this valuable feedback! We're on it.

lib/Controller/FederatedInvitesController.php
redblom: :+1:  https://github.com/nextcloud/contacts/pull/4417/commits/eb7e7af8a4c7018f7725d5f3e77c4166eb5c53bf

lib/Controller/FederatedInvitesController.php
redblom: :+1:  https://github.com/nextcloud/contacts/pull/4417/commits/ea98855205bad78153a91291b5ef83fe5b8e4045

lib/Controller/FederatedInvitesController.php
redblom: :+1:  https://github.com/nextcloud/contacts/pull/4417/commits/ea98855205bad78153a91291b5ef83fe5b8e4045

`inviteToken` and `inviteProvider` are not already accessible.

lib/WayfProvider.php
redblom: :+1: https://github.com/nextcloud/contacts/pull/4417/commits/2a3d1fee57a3b70740647f52de7173fc382548a6

lib/Controller/FederatedInvitesController.php
redblom: ?? I don't see that (much) in other places 
Also: https://docs.nextcloud.com/server/stable/developer_manual/getting_started/coding_standards/php.html#comments

lib/WayfProvider.php
redblom: :+1: https://github.com/nextcloud/contacts/pull/4417/commits/88ae1b584d3dee6de0d5d60dc338e855a7af1b41

lib/Controller/FederatedInvitesController.php
redblom: Indeed!
:+1: https://github.com/nextcloud/contacts/pull/4417/commits/1438c7270c3ea15ec212da3cf0b14fa076f3947f

lib/Command/OcmInvitesConfig.php
redblom: :+1: https://github.com/nextcloud/contacts/commit/d1e34e7b382cada3bdf852df8f29b7b7e7677d22

redblom:
@artonge - addressed all your reviews.
I have two questions left where I would like your input:
https://github.com/nextcloud/contacts/pull/4417#discussion_r3274687716
https://github.com/nextcloud/contacts/pull/4417#discussion_r3302996398

lib/Cron/UpdateOcmProviders.php
miaulalala: TimedJob is the right choice - but split it in two:

Every 15 minutes, check the list if new hosts / instances were added. Only add those to the list and check their capabilities, don't read every host.

Every 24 hours run a full sync on every other host that's in the mesh and re- write the capabilities to the cache. Outside of working hours would be best.

src/css/wayf.scss
redblom: :+1: https://github.com/nextcloud/contacts/pull/4417/commits/622904d00a6b85e51a933e718dd79a5160c00a7c

redblom:
> This seems very comprehensive, nice! :) From what I understand this is a way to exchange contacts with people from other organizations directly from the contacts app. Design wise I have a few questions and some feedback:
> 
> ...

#### Re-design of the layout:
main-01-main drawio

_3 buttons collapsed into 1 'OCM invitations' and 1 permanent sub item 'All invitations' that shows the nr. of pending invitations_


main-01-main-dropdown drawio

_Sub menu displays the options to create and accept invites_

main-01-main-create-invitation-show-email-preview-button drawio

_A preview of the email can be shown_

main-01-main-create-invitation-show-email-preview drawio

_The email preview_

main-01-main-all-invitations drawio

_Options to copy link, (encoded) invite_

lib/Cron/UpdateOcmProviders.php
redblom: :+1: https://github.com/nextcloud/contacts/pull/4417/commits/9142baf7210868a4a3e6116d7cf9683859fddd8a

lib/Cron/UpdateDeltaOcmProviders.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
miaulalala:
```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

Please also add a short description about what this job does

lib/Cron/UpdateOcmProviders.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
miaulalala:
```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

Same here, please add a description

lib/Db/FederatedInvite.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
miaulalala:
```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

lib/Db/FederatedInviteMapper.php
```diff
@@ -0,0 +1,183 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
miaulalala: same

lib/Exception/ContactExistsException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
miaulalala: same

lib/Service/FederatedInvitesService.php
```diff
@@ -0,0 +1,257 @@
...
+		if (trim($recipientProvider) === '' || trim($userID) === '' || trim($email) === '' || trim($name) === '') {
+			return new JSONResponse(['message' => 'Could not accept invite, user data is incomplete.', 'error' => true], Http::STATUS_BAD_REQUEST);
+		}
+
+		try {
+			$invitation = $this->federatedInviteMapper->findByToken($token);
+		} catch (DoesNotExistException) {
+			$response = ['message' => 'Invalid or non existing token', 'error' => true];
+			$status = Http::STATUS_BAD_REQUEST;
+			$response = new JSONResponse($response, $status);
```
miaulalala:
```suggestion
			$response = new JSONResponse($response, Http::STATUS_BAD_REQUEST);
```

Same for all others

lib/Exception/ContactExistsException.php
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Contacts\Exception;
+
+use Exception;
+
+class ContactExistsException extends Exception {
```
miaulalala:
```suggestion
class ContactAlreadyExistsException extends Exception {
```

lib/Service/SocialApiService.php
```diff
@@ -234,6 +239,130 @@ public function updateContact(string $addressbookId, string $contactId, ?string
...
+				[
+					'FN' => $name,
+					'EMAIL' => $email,
+					'CLOUD' => $cloudId,
+				],
+				$addressBook->getKey()
+			);
+			return $newContact;
+		} catch (Exception $e) {
+			$this->logger->error('An exception occurred creating a federated contact: ' . $e->getTraceAsString(), ['app' => Application::APP_ID]);
```
miaulalala:
```suggestion
			$this->logger->error('An exception occurred creating a federated contact: ' . $e->getMessage()(), ['exception' => $e]);
```

miaulalala (changes requested):

lib/Cron/UpdateDeltaOcmProviders.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

lib/Cron/UpdateOcmProviders.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

lib/Db/FederatedInvite.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

lib/Db/FederatedInviteMapper.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

lib/Exception/ContactExistsException.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

lib/Service/FederatedInvitesService.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

lib/Exception/ContactExistsException.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

lib/Service/SocialApiService.php
redblom: https://github.com/nextcloud/contacts/pull/4417/commits/06636c830032e77eceac9f0543dfe7b6341b96ea

redblom:
#### GUI after modifications:

Screenshot from 2026-06-21 15-08-59
New External invitations header and context menu

Screenshot from 2026-06-21 14-34-23
Email preview option

Screenshot from 2026-06-21 14-34-37
Preview email

miaulalala:
Feedback by @jancborchardt:

**Must haves**
- [x] Make "All invitations" a list item below the address book list instead of a separate menu section:
image

- [x] Make adding a new invitation or accepting an invitation as a separate button next to the "new contact" button - example from Talk:
image

- [x] Make "email address" the first input field in the Invitation modal and remove the label
- [x] Replace "Provider" with "Select your server" as a heading. Remove subtext from heading.
- [x] Put accept buttons on the rightmost side, and be a primary button (if not fixed yet)
- [x] Add "create invitation" and "accept invitation" on empty content page when no invitations are open
- [x] Add title to Invitation and Accept modal
- [x] Remove icon from cancel button in Invitation and Accept Modal
- [x] No full stop in headings
- [ ] Give good user feedback for errors - such as an invite not being able to be accepted
	- [ ] Also hide the button to accept an invite if the capability is not there (may already be done) 

**Nice to have** - will not block merge
- [x] Remove extra container around recipient email and personal message so all form element are on one level
- [ ] Change input field for adding a custom provider: show a "+ Add custom provider" button -> show input field only when button is clicked
- [ ] Rename "Continue" button to "Accept invite" button to "provider" page
- [ ] Rename "Cancel" to "Decline invite" for Accept modal
- [ ] Reduce width of provider list on invitation acceptance page
- [ ] Remove cloud icon from provider / server list
- [ ] Add border radius consistently for "provider" page

**Future enhancement and ideas**
-  [ ] store chosen server in localstorage
-  [ ] moving federated couldId field from ...more to the immediate menu
-  hand down who is sending the invitation via link hash or similar - to be discussed with the OCM working group

jancborchardt:
https://github.com/nextcloud/contacts/pull/4417#issuecomment-4789621159

FYI for reference @nimishavijay this is design feedback I gave in the call. I’ll also take over to do a last review pass over it after the changes are done. :)

redblom:
Screenshots belonging to https://github.com/nextcloud/contacts/pull/4417/commits/66bca00d00442271865ed60da2260dffa6ffaed1


Screenshot from 2026-06-26 15-30-26
OCM invitations menu moved to top.


Screenshot from 2026-06-26 15-32-09
New invitation form: label field removed, title fixed, removed extra container


Screenshot from 2026-06-26 15-32-39
Accept invite form: title fixed


Screenshot from 2026-06-26 15-33-13
All invitations moved

Screenshot from 2026-06-26 15-33-45
Options to create/accept invites on empty invitations page

Screenshot from 2026-06-26 15-34-55
Correct message when trying to accept an invite you've created yourself


Screenshot from 2026-06-26 15-53-52
WAYF page: new title, removed sub text

jancborchardt (commented):
Nice @redblom! Looks very good and basically good to go!

Only have 3 tiny pieces of feedback based on those changes, but they are also not a blocker:
- [x] The nav entry should be called "External invitations" instead of "All invitations" as that is clearer, no?
- [x] In the modal for "Create invitation", the button says "Send invite", but everywhere else we use the language "invitation". So better to use "Send invitation" here too.
- [ ] On the emptycontent page "There are no invites", there are 2 primary buttons. There should only ever be 1 primary button, in this case it makes most sense to be "Create invitation". (And they can be next to each other rather than below each other.)

redblom:
@jancborchardt - Thank you for the fast review !!

> Nice @redblom! Looks very good and basically good to go!
> 
> Only have 3 tiny pieces of feedback based on those changes, but they are also not a blocker:
> 
>     * [ ]  The nav entry should be called "External invitations" instead of "All invitations" as that is clearer, no?
> 
>     * [ ]  In the modal for "Create invitation", the button says "Send invite", but everywhere else we use the language "invitation". So better to use "Send invitation" here too.
> 
>     * [ ]  On the emptycontent page "There are no invites", there are 2 primary buttons. There should only ever be 1 primary button, in this case it makes most sense to be "Create invitation". (And they can be next to each other rather than below each other.)

Screenshots with implementation:
Screenshot from 2026-06-28 12-05-41

Screenshot from 2026-06-28 12-06-33

miaulalala:
@jancborchardt can you +1 if you're happy with the changes? @hamza221 I'd apprechiate a review as well. Should be merged only after the server PR is in.

jancborchardt (approved):
Thought I already approved, here you go. :)

ChristophWurst:
> Should be merged only after the server PR is in.

Didn't look at the full diff but sounds like this needs a version guard then. The Contacts app is distributed via the app store. It may be updated earlier than an update updates the server. That must not lead to an incompatibility.

redblom:
> > Should be merged only after the server PR is in.
> 
> Didn't look at the full diff but sounds like this needs a version guard then. The Contacts app is distributed via the app store. It may be updated earlier than an update updates the server. That must not lead to an incompatibility.

Firstly, the unit tests that test the OCM invitation flow do not account for nextcloud versions below 35 and I guess that's why those tests currently fail. So that needs to be fixed. 
... assuming we don't want `nextcloud min-version` to be `35`. 

Second, @ChristophWurst, what version guard implementation do you have in mind?

ChristophWurst:
A guard to make this app usable when deployed on 33/34. Perhaps the feature can just be hidden for those.

redblom:
> A guard to make this app usable when deployed on 33/34. Perhaps the feature can just be hidden for those.

Perfect. That has been accounted for. ~~Will do~~.

The feature is hidden by default.

Also there are two backport PRs for 33/34 ([#62460](https://github.com/nextcloud/server/pull/62460), #[62486](https://github.com/nextcloud/server/pull/62486) ) for this waiting for acceptance.

redblom:
Rebase.

redblom:
Merging [#62460](https://github.com/nextcloud/server/pull/62460) and #[62486](https://github.com/nextcloud/server/pull/62486) should make the CI build succeed.

miaulalala (approved):

redblom:
```
Run composer run psalm -- --threads=1 --monochrome --no-progress --output-format=github
> psalm '--threads=1' '--monochrome' '--no-progress' '--output-format=github'
Error: lib/Controller/FederatedInvitesController.php:299:40: TypeDoesNotContainType: '' cannot be identical to non-empty-string (see https://psalm.dev/056)
Error: lib/Controller/FederatedInvitesController.php:299:75: TypeDoesNotContainType: '' cannot be identical to non-empty-string (see https://psalm.dev/056)
------------------------------
2 errors found
```
This does not seem to occur locally. What's going on here ??

ChristophWurst:
> This does not seem to occur locally. What's going on here ??

CI installs stub packages dynamically. You might have Psalm running against another version. See here https://github.com/nextcloud/contacts/blob/59ea0379f2b9efdbb73fa7362f02af4723cd3636/.github/workflows/psalm-matrix.yml#L73-L74

lib/Service/SocialApiService.php
```diff
@@ -107,30 +111,22 @@ protected function addPhoto(array &$contact, string $imageType, string $photo) {
...
 			}
 		}
-
-		$addressBookIsUpdatable = $addressBook !== null
-			&& ($addressBook->getPermissions() & Constants::PERMISSION_UPDATE);
-
-		if (!$addressBookIsUpdatable) {
-			return null;
-		}
-
```
hamza221, line 133: Wouldn't this cause `updateContact`down below to fail on readonly addressbooks ?

lib/Service/SocialApiService.php
```diff
@@ -252,7 +252,131 @@ public function updateContact(string $addressBookId, string $contactId, ?string
...
+	 * @param string $cloudId the cloud id of the federated contact
+	 * @param string $email the email of the federated contact
+	 * @param string $name the display name of the federated contact
+	 * @param string $userId the uid of the local (inviter) user
+	 *
+	 * @return array|null the created contact array, or null if a contact with
+	 *                    that cloud id already exists or there is no suitable
+	 *                    writable address book for the inviter
+	 */
+	public function createFederatedContact(string $cloudId, string $email, string $name, string $userId): ?array {
```
hamza221: is this used anywhere ?

src/components/AppNavigation/Settings/SettingsImportContacts.vue
```diff
@@ -307,6 +316,10 @@ export default {
 			this.isOpened = false
 			this.loading = false
 		},
+
+		getOcmInvitesActiveClass() {
+
+		},
```
hamza221: Leftover?

src/components/Ocm/OcmInviteAccept.vue
```diff
@@ -0,0 +1,102 @@
+<!--
+  - SPDX-FileCopyrightText: 2018 Nextcloud GmbH and Nextcloud contributors
```
hamza221:
```suggestion
  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

src/components/Ocm/OcmInviteDetails.vue
```diff
@@ -0,0 +1,443 @@
+<!--
+  - SPDX-FileCopyrightText: 2018 Nextcloud GmbH and Nextcloud contributors
```
hamza221:
```suggestion
  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

lib/Service/SocialApiService.php
redblom: @hamza221 - I'm on all of these. There's definitely some technical dept that has developed.

redblom: Done. Reverted.

lib/Service/SocialApiService.php
redblom: Removed, fixed tests.

src/components/AppNavigation/Settings/SettingsImportContacts.vue
redblom: Removed

src/components/Ocm/OcmInviteAccept.vue
redblom: All fixed.

src/components/Ocm/OcmInviteDetails.vue
redblom: All fixed.

icewind1991:
Please prefix your table names with the appid in the future.

The "federated_invites" table name was previously used already by `cloud_federation_api` and the duplication is leading to migration conflicts.

redblom:
> Please prefix your table names with the appid in the future.
> 
> The "federated_invites" table name was previously used already by `cloud_federation_api` and the duplication is leading to migration conflicts.

@icewind1991 - Understood.
We leaving this one as is right?

icewind1991:
> We leaving this one as is right?

Yes, attempting to change it now would only lead to more issues.

PR #4425, opened 2025-04-08
fix(lint-php-cs): use minimum available php version

hamza221:
reverting #4422 and opting for #6831 to fix php lint job


st3iny (approved):
Thanks, this is a bit cleaner.

PR #4461, opened 2025-05-13
Revert "chore(deps-dev): bump vite from 5.4.19 to 6.3.4"

hamza221:
Breaks router 
revert #4453


st3iny (approved):
We should be compatible with Vite 6 but for some reason it breaks the router. Tested and works.

Would be nice to still make this works as there are some vulnerabilities in the old Vite version.

Issue #4392, opened 2025-03-17
Bulk actions on contacts


schiessle:
It would be great if I could select muultiple contacts in the list (similar to the file listing) and than perform bulk actions like:
- delete the selected addresses
- move them to a different address book
- ...

ChristophWurst:
Potential duplicate of https://github.com/nextcloud/contacts/issues/277.

jancborchardt:
Closing as duplicate of https://github.com/nextcloud/contacts/issues/277 – transferred the "Planned" and "Priority: Medium" status to the other issue, which had more info on Sprint etc anyway.

schiessle:
@GVodyanov Just out of curiosity. The finished pull request only show the delete action for selected files. Are other actions also supported? The main driver of my feature request was that I had to move multiple contacts to a different address book

ChristophWurst:
Delete is the MVP action for multiselect. Other actions can be added later on.

PR #4477, opened 2025-05-26
Add multiactions in the contact list

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4392
Fix https://github.com/nextcloud/contacts/issues/277

image
image

How to interact:
- Regular avatar click: select one at a time
- Shift click on body: select one at a time and range selection
- Shift click on avatar: only range selection

Range selection works by clicking on two extremes and everything in the middle will be selected, exactly like in mail.

There is also a small dialogue when deleting contacts.

### DIFF
--- a/src/components/ContactsList.vue
+++ b/src/components/ContactsList.vue
@@ -5,32 +5,81 @@
 
 <template>
 	<AppContentList class="content-list">
+		<NcDialog :open="showDeleteConfirmationDialog"
+			:name="n(
+				'contacts',
+				'Delete {number} contact',
+				'Delete {number} contacts',
+				multiSelectedContacts.size,
+				{ number: multiSelectedContacts.size }
+			)"
+			:buttons="buttons"
+			no-close>
+			{{ t('contacts', 'Are you sure you want to proceed?') }}
+			<NcNoteCard v-if="readOnlyMultiSelectedCount"
+				type="info"
+				:text="n('contacts',
+					'Please note that {number} contact is read only and won\'t be deleted',
+					'Please note that {number} contacts are read only and won\'t be deleted',
+					readOnlyMultiSelectedCount,
+					{ number: readOnlyMultiSelectedCount })" />
+		</NcDialog>
+
 		<div class="contacts-list__header">
 			<div class="search-contacts-field">
 				<input v-model="query" type="text" :placeholder="t('contacts', 'Search contacts …')">
 			</div>
 		</div>
+		<transition name="contacts-list__multiselect-header">
+			<div v-if="isMultiSelecting" class="contacts-list__multiselect-header">
+				<NcButton type="tertiary"
+					:title="t('contacts', 'Unselect {number}', { number: multiSelectedContacts.size })"
+					:close-after-click="true"
+					@click.prevent="unselectAllMultiSelected">
+					<IconSelect :size="16" />
+				</NcButton>
+				<NcButton type="tertiary"
+					:disabled="!isAtLeastOneEditable"
+					:title="deleteActionTitle"
+					:close-after-click="true"
+					@click.prevent="attemptDeleteAllMultiSelected">
+					<IconDelete :size="16" />
+				</NcButton>
+			</div>
+		</transition>
+
 		<VirtualList ref="scroller"
 			class="contacts-list"
 			data-key="key"
 			:data-sources="filteredList"
 			:data-component="ContactsListItem"
 			:estimate-size="68"
-			:extra-props="{reloadBus}" />
+			:extra-props="{reloadBus, onSelectMultipleFromParent: onSelectMultiple, onSelectRangeFromParent: onSelectRange }" />
 	</AppContentList>
 </template>
 
 <script>
-import { NcAppContentList as AppContentList } from '@nextcloud/vue'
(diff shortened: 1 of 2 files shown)


src/components/ContactsList.vue
```diff
@@ -5,32 +5,84 @@
...
+				'contacts',
+				'Delete {number} contact',
+				'Delete {number} contacts',
+				multiSelectedContacts.size,
+				{ number: multiSelectedContacts.size }
+			)"
+			:buttons="buttons"
+			no-close>
+			{{ t('contacts', 'Are you sure you want to proceed?') }}
+			<span v-if="readOnlyMultiSelectedCount" class="content-list-dialog-warning">
```
hamza221, line 19: **Non blocking**.
Can we prevent selection of readonly cards instead ?

src/components/ContactsList.vue
```diff
@@ -5,32 +5,84 @@
...
+		</NcDialog>
+
 		<div class="contacts-list__header">
 			<div class="search-contacts-field">
 				<input v-model="query" type="text" :placeholder="t('contacts', 'Search contacts …')">
 			</div>
 		</div>
+		<div v-if="isMultiSelecting" class="contacts-list__multiselect-header">
+			<NcButton type="tertiary"
+				:title="n('contacts', 'Unselect {number}', 'Unselect {number}', multiSelectedContacts.size, { number: multiSelectedContacts.size })"
```
hamza221, line 36:
```suggestion
				:title="n('contacts', 'Unselect {number} card', 'Unselect {number} cards', multiSelectedContacts.size, { number: multiSelectedContacts.size })"
```
Or just
`t('contacts', 'Unselect {number}', { number: multiSelectedContacts.size }`

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -156,13 +174,20 @@ export default {
 				params: { selectedGroup: this.selectedGroup, selectedContact: this.source.key },
 			})
 		},
+		onSelectMultiple() {
+			// This weirdness of passing a function as a prop is because the VirtualList extra-props prop object does not support listening to custom events (afaik)
```
hamza221, line 178: Why not use an event bus ?

hamza221 (changes requested):
I found this bug 
| clicking on the icon | clicking on the components   |
|--------|--------|
|image |image |

GVodyanov:
> I found this bug
> clicking on the icon 	clicking on the components
> https://github.com/user-attachments/assets/653755d1-ee75-4395-aac6-5f73c2b9375a 	https://github.com/user-attachments/assets/17c95786-5a70-4258-ab41-9460687ad746

Hey! are you sure that's the right video @hamza221 😅?

Either way, yes clicking on the avatar has a different action, like in mail. Is it not working intuitively?

src/components/ContactsList/ContactsListItem.vue
GVodyanov: I think it's discouraged in Vue3, having said that what I'm doing here isn't exactly standard either, but it might be more future proof.

hamza221:
> Hey! are you sure that's the right video @hamza221 😅?


Mb I uploaded the wrong video, please have another look I updated it

hamza221:
> Either way, yes clicking on the avatar has a different action, like in mail. Is it not working intuitively?

IMO selection should either work the same way when clicking on the component and Avatar, or not work at all when clicking the component. 
Also Shift+ click is usually inclusive of the last element pressed.

GVodyanov:
Notes from meeting:
- Only show actions if at least one contact supports that action

Advice from Nimisha:
- Make the check a little smaller
- Add a short animation when showing the actions


EDIT: done

GVodyanov:
image

@nimishavijay Here's the checkboxes becoming a little smaller, delete being grayed out because it can't be applied, and on hover it says "Please select at least one editable contact to delete".

I also added the same animation as in mail but quicker.

hamza221 (changes requested):
The above mentioned bug is still unresolved 
https://github.com/user-attachments/assets/b2e6a3e9-fc45-4c9e-93f3-45d9058986fc

GVodyanov:
> The above mentioned bug is still unresolved https://github.com/user-attachments/assets/b2e6a3e9-fc45-4c9e-93f3-45d9058986fc
> 
> Edit: Mail app for comparison
> Screen.Recording.2025-06-20.at.18.22.19.mov

Ahn I see, fixed now, thanks!

GVodyanov:
> Nice! Looks great with the smaller checkmarks now.
> 
> Only change would be in the confirmation dialog. We can use an NcNoteCard component in the warning style to show the read only warning, that should make it more eye-catching :)

Good idea!

Here's how I made it look now
image

src/components/ContactsList.vue
```diff
@@ -56,6 +105,23 @@ export default {
 		return {
 			ContactsListItem,
 			query: '',
+			multiSelectedContacts: new Map(),
+			refreshKey: 0, // used to force re-render of the list when search query changes, can be removed in vue3
```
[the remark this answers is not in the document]
GVodyanov: Nope, good catch

PR #4521, opened 2025-06-22
Add merging duplicate contacts

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/70

### DIFF
--- a/src/components/ContactsList.vue
+++ b/src/components/ContactsList.vue
@@ -5,32 +5,81 @@
 
 <template>
 	<AppContentList class="content-list">
+		<NcDialog :open="showDeleteConfirmationDialog"
+			:name="n(
+				'contacts',
+				'Delete {number} contact',
+				'Delete {number} contacts',
+				multiSelectedContacts.size,
+				{ number: multiSelectedContacts.size }
+			)"
+			:buttons="buttons"
+			no-close>
+			{{ t('contacts', 'Are you sure you want to proceed?') }}
+			<NcNoteCard v-if="readOnlyMultiSelectedCount"
+				type="info"
+				:text="n('contacts',
+					'Please note that {number} contact is read only and won\'t be deleted',
+					'Please note that {number} contacts are read only and won\'t be deleted',
+					readOnlyMultiSelectedCount,
+					{ number: readOnlyMultiSelectedCount })" />
+		</NcDialog>
+
 		<div class="contacts-list__header">
 			<div class="search-contacts-field">
 				<input v-model="query" type="text" :placeholder="t('contacts', 'Search contacts …')">
 			</div>
 		</div>
+		<transition name="contacts-list__multiselect-header">
+			<div v-if="isMultiSelecting" class="contacts-list__multiselect-header">
+				<NcButton type="tertiary"
+					:title="t('contacts', 'Unselect {number}', { number: multiSelectedContacts.size })"
+					:close-after-click="true"
+					@click.prevent="unselectAllMultiSelected">
+					<IconSelect :size="16" />
+				</NcButton>
+				<NcButton type="tertiary"
+					:disabled="!isAtLeastOneEditable"
+					:title="deleteActionTitle"
+					:close-after-click="true"
+					@click.prevent="attemptDeleteAllMultiSelected">
+					<IconDelete :size="16" />
+				</NcButton>
+			</div>
+		</transition>
+
 		<VirtualList ref="scroller"
 			class="contacts-list"
 			data-key="key"
 			:data-sources="filteredList"
 			:data-component="ContactsListItem"
 			:estimate-size="68"
-			:extra-props="{reloadBus}" />
+			:extra-props="{reloadBus, onSelectMultipleFromParent: onSelectMultiple, onSelectRangeFromParent: onSelectRange }" />
 	</AppContentList>
 </template>
 
 <script>
-import { NcAppContentList as AppContentList } from '@nextcloud/vue'
+import { NcAppContentList as AppContentList, NcButton, NcDialog, NcNoteCard } from '@nextcloud/vue'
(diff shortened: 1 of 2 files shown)


nimishavijay:
Looks really good! :) The layout of the conflicts dialog is really nice! I have some questions and a couple of points to make it more understandable:

- [x] Wording changes:
  - [x] `Merge contacts summary` --> `Confirm merging contacts`
  - [x] In the warning notecard we can provide more info --> `The selected contacts have conflicting information. Choose which information to keep`

- [x] Layout/spacing changes:
  - [x] We can group each conflict by adding a horizontal divider line, or a container in --background-dark around each set of conflicts 
  - [x] Move the merge button to the bottom as that's where we keep all our buttons generally
  - [x] We can sort the conflicts on top so the user doesn't have to scroll all the way to the end to make sure there are no conflicts. What do you think?
  - [x] Once the user has selected the info, I see that the yellow `(✖️)`  icon turns white, nice touch! Additionally, we can change the icon to the `(✔️)` icon (many times changing just the color may not be accessible)
  - [x] Once all conflicts have been resolved, instead of removing the warning, we can turn it into a success notecard with `Contacts can be merged`

Questions: 
- [x] Is this dialog shown even if there are no conflicts? I'm thinking there can be 3 states: 
  - [x] when the 2 contacts are exactly the same: just a simple text-only confirmation dialog 
  - [x] when they are not the same but there are no conflicts (eg. one contact has phone and email and another has phone, email and address): show a summary like this with all the info that the new contact will have (essentially a preview of the merged contact)
  - [x] one for when there are conflicts: this dialog
  - [x] possibly 1 and 2 can be combined, so that we always see the preview of the merged contact
- [x] What happens when you try to merge more than 2 contacts (is it even possible)? Are there more columns?

GVodyanov:
Hey @nimishavijay, thanks a lot for the comments, I really like them! I'll send some more screenshots with the changes once I finish all the properties.

Here are the answers to your questions:

> Is this dialog shown even if there are no conflicts? I'm thinking there can be 3 states: 

So for the multiple states, what I thought of doing is this modal format for all cases, even when there aren't conflicts, just to see a summary. Currently, when there are no conflicts, you won't have any checkboxes, so it's basically read only and just a confirmation dialog.

Something to keep in mind is that even if there are no direct "conflicts", we can still give the user some margin of choice. For instance, you might have a situation in which one contact has a work number and the other has a home number. They technically aren't a conflict, and can be combined, but if they happen to be the same number the user might want to only include one of the two. I just talked about this with Christoph and I have to finish thinking all the possibilities through.

> What happens when you try to merge more than 2 contacts (is it even possible)? Are there more columns?

Doing more than 2 contacts wouldn't be terribly difficult, it would just require having horizontal scrolling I guess, but we decided that this PR will only do 2 contacts at a time, maybe in the future we can expand on that.

src/components/ContactsList/Merging.vue
```diff
@@ -0,0 +1,648 @@
...
+					}
+				}
+			})
+
+			contactToSave.jCal = jCal
+
+			await this.$store.dispatch('deleteContact', { contact: this.contactsList[0] })
+			await this.$store.dispatch('deleteContact', { contact: this.contactsList[1] })
+
+			await this.$store.dispatch('addContact', contactToSave)
```
GVodyanov, line 513: I can't understand how to modify and then update an existing contact object. Modifying the jCal like this directly is the only thing that I was able to figure out, but is there a better way of doing this?

Also when I go to add the contact I get an error saying `invalid REV`, is it maybe because the URL has to be updated somehow? 

Any help would be appreciated.

src/components/ContactsList/Merging.vue
```diff
@@ -0,0 +1,629 @@
...
+					finalProperties[name].forEach((property) => {
+						contactToSave.vCard.updatePropertyWithValue(name, this.getPropertyValue(property))
+					})
+				}
+			})
+
+			contactToSave.groups = this.selectedGroups
+			contactToSave.addressbook = this.contactsList[this.chosenAddressBook.id].addressbook
+
+			await this.$store.dispatch('addContact', contactToSave)
```
GVodyanov, line 494: Can't figure out how to make this update on the server too, any help would be appreciated

hamza221:
Not a blocker, but when the same property has the same value in both cards, it would make sense to bypass the conflict resolution 
image

GVodyanov:
> Not a blocker, but when the same property has the same value in both cards, it would make sense to bypass the conflict resolution image

I love javascript

image

Fixing now...

src/components/ContactsList.vue
```diff
@@ -45,6 +52,13 @@
 					@click.prevent="attemptDeleteAllMultiSelected">
 					<IconDelete :size="16" />
 				</NcButton>
+				<NcButton type="tertiary"
+					:disabled="!areTwoEditable"
+					:title="mergeActionTitle"
+					:close-after-click="true"
+					@click.prevent="initiateContactMerging">
+					<IconSetMerge :size="18" />
```
ChristophWurst, line 60: 20? https://github.com/nextcloud/mail/issues/11438

ChristophWurst (commented):
Looks very well structured code-wise :clap:

MrRinkana:
Awesome! Thank you a lot!

Coming in 7.2.6 or more of a 7.3 thing?

hamza221:
> Awesome! Thank you a lot!
> 
> 
> 
> Coming in 7.2.6 or more of a 7.3 thing?

7.3.0 and rc is planned this week.

christiansass:
so the idea here is that we’re going to test the two contact merge functions in the wild for now. without any issues we can look at expanding that into multi contact merge? 

Only reason I ask is the first time this showed up for me I had four versions of one contact that I would love to have combined at once. 

I’m guessing it under this system I would just continue to merge two down until you had only one remaining?

PR #4557, opened 2025-07-13
[stable7.1] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (10 unfixable, 11 only fixable manually using --force)


hamza221:
No more releases from 7.1

PR #4577, opened 2025-07-23
fix: saving vcard fails due wrong rev value

kesselb:
This looks like another issue from the ical 2.2.0 update. An ICAL.VCardTime object is mapped to the date-and-or-time type. Assigning such an object to contact.rev triggers an error deep within ical.js when trying to "undecorate" the object with the given vcard design set.[^1]

The current version would also work by changing `contact.rev = rev` to `contact.rev = rev.toICALString()` because that would skip the undecorating path.

As I understand it, date-and-or-time is available only in vCard 4. Therefore, I used ICAL.Time, which should be available in vCard 3 as the date-time type.

[^1]:
https://github.com/kewisch/ical.js/blob/ac16646655277764a6a09ccfab88152ea3c62280/lib/ical/property.js#L208-L210


src/store/contacts.js
```diff
@@ -331,9 +331,10 @@ const actions = {
 		validate(contact)
 
 		// Update REV
-		const rev = new ICAL.VCardTime()
-		rev.fromUnixTime(Date.now() / 1000)
-		contact.rev = rev
+		// ICAL.Time.now() would also work, but I’m using fromJsDate with useUTC=true on purpose
+		// to create an ical date in UTC rather than in a local timezone for better interoperability.
+		const rev = ICAL.Time.fromJSDate(new Date(), true)
+		contact.rev = rev.toICALString()
```
kesselb, line 337: It seems that toICALString() is generating an invalid value like `REV:20250723T15040Z` (the time part is one digit short).

kesselb:
Sebastian's taking care of it in https://github.com/nextcloud/contacts/pull/4573

PR #4606, opened 2025-08-08
feat: create and add guests to teams

leftybournes:
Depends on: nextcloud/guests#1371

Allow users to invite guests directly through the teams dialog.

Before:
image

After:
image

### DIFF
--- a/src/components/EntityPicker/EntityPicker.vue
+++ b/src/components/EntityPicker/EntityPicker.vue
@@ -9,14 +9,29 @@
 		<!-- Wrapper for content & navigation -->
 		<div class="entity-picker">
 			<!-- Search -->
-			<div class="entity-picker__search">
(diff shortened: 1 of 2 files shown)


GVodyanov:
Hello! Thanks a lot for your PR

GVodyanov:
I tried making it work, but I don't have any experience with the Guest app. I see that `OCA.Guests` is undefined, even though I can see the guests in the invitation list and inviting from files works. I'm assuming this has to do with an issue from my configuration and not the code though, so if you could give advice on how to make this run it would be appreciated. Thanks!

leftybournes:
This is part of a collection of pull requests that I opened for the guest creation feature. You will need changes from https://github.com/nextcloud/guests/pull/1371 in the guests app for the changes here to work.

PR #4609, opened 2025-08-11
Revert "fix: use built now function instead of manually creating date"

hamza221:
discovered in #4521 
reverts https://github.com/nextcloud/contacts/pull/4573 

According to the pre-existing logic : 
https://github.com/nextcloud/contacts/blob/fix/invalid-rev/src/services/checks/invalidREV.js#L24-L30 
`vcardtime` is expected for v3 contacts and `icaltime` is expected for v4
With the reverted commit that was not respected. According to that logic even the repaired rev is wrong. 

creating rev from unix timestamp or js date are both supported by the library so I don't see anything wrong with the old method imo

### DIFF
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -66,7 +66,9 @@ export default class Contact {
 
 		// if no rev set, init one
 		if (!this.vCard.hasProperty('rev')) {
-			this.vCard.addPropertyWithValue('rev', ICAL.VCardTime.now().convertToZone(ICAL.Timezone.utcTimezone))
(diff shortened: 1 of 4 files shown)


GVodyanov (approved):

SebastianKrupinski:
FYI, doing this will break contacts saving. The vcardtime  does not seem to work properly in the updated ical.js.

I would recommend fixing the check instead, since its a 5 year old patch for a bug that probably does not exist anymore

hamza221:
fixed in #4634

PR #4619, opened 2025-08-18
fix(ui): escape organization/title properties

ChristophWurst:
To fix encoding issues


hamza221:
`npm ci` works locally 🤔

ChristophWurst:
Bodged the lodash installation without picomatch removal locally

PR #4621, opened 2025-08-18
[stable6.0] fix(ui): escape organization/title properties

ChristophWurst:
Manual backport of https://github.com/nextcloud/contacts/pull/4619

The diff is smaller because read-only contacts do not have a dedicated view.


hamza221:
test summary was introduced in https://github.com/nextcloud/contacts/pull/3757

PR #4634, opened 2025-08-21
fix: generate rev depending on card version

hamza221:
depends on https://github.com/nextcloud/server/pull/54573


SebastianKrupinski (approved):

src/services/checks/invalidREV.js
```diff
@@ -40,11 +40,18 @@ export default {
...
 			contact.vCard.removeProperty('rev')
 
-			// creatiing new value
-			contact.vCard.addPropertyWithValue('rev', ICAL.VCardTime.now().convertToZone(ICAL.Timezone.utcTimezone))
+			// creating new value
+			const version = contact.version
+			if (version === '4.0') {
+				contact.vCard.addPropertyWithValue('rev', ICAL.Time.fromJSDate(new Date(), true))
+			}
+			if (version === '3.0') {
```
SebastianKrupinski, line 48: I think we should alter the condition here. I think this needs to also accommodate, prior versions, as this logic would break down if someone has a card with a lower version. Idk, if it's worth it tho, bring it up tomorrow, lets see what the team thinks

hamza221: No other versions are supported but let's leave it for the call

realDayaa:
I think there still is an issue with this merge request unfortunatly.
When changing a contact the rev is now displayed but the timezone is ignored. So when changing a value instead of displaying "now" it shows "in 2 hours".
I have set the timezone of the server and client to Europe/Berlin.

PR #4643, opened 2025-08-25
Implement addressbook sorting

GVodyanov:
Introduces sorting for addressbooks. This sorting is applied wherever addressbooks have to be listed, so in contacts settings, contact details editor, and the contact cloning modal. 

The sorting rules are the following:

 ```
Sorts addressbooks into groups with the following priority:
 1. First group: My addressbooks
 2. Second group: Shared with me and writeable
 3. Third group: Read-only
 4. Fourth group: Disabled

 Withing each group, addressbooks are sorted by the number of contacts, from highest to lowest
```

In `ContactsDetails.vue` the pre-selected addressbook is the last used one, if there is a last used one.

> EDIT THIS IS THE PR FOR THIS NOW https://github.com/nextcloud/contacts/pull/4659 
> It also fixes a bug where when modifying the addressbook of a contact caused the app to break by introducing an addressbook with an id of `undefined`

### DIFF
--- a/src/store/addressbooks.js
+++ b/src/store/addressbooks.js
@@ -7,6 +7,7 @@ import { showError } from '@nextcloud/dialogs'
 import pLimit from 'p-limit'
 
 import Contact, { MinimalContactProperties } from '../models/contact.js'
+import { sortAddressbooks } from '../utils/addressbookUtils.js'
 
 import client from '../services/cdav.js'
 import parseVcf from '../services/parseVcf.js'
@@ -226,6 +227,14 @@ const mutations = {
 		sharee.writeable = !sharee.writeable
 	},
 
+	/**
+	 * Needed to track indirect state changes for addressbook sorting
+	 *
+	 * @param state
+	 */
+	resortAddressbooks(state) {
+		state.addressbooks = sortAddressbooks(state.addressbooks)
+	},
 }
 
 const getters = {
@@ -257,6 +266,8 @@ const actions = {
 		addressbooks.forEach(addressbook => {
 			context.commit('addAddressbook', addressbook)
 		})
+
+		context.commit('resortAddressbooks')
 
 		context.state.addressbooksFetched = true
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@mdi/svg": "^7.4.47",
         "@nextcloud/auth": "^2.5.2",
         "@nextcloud/axios": "^2.5.2",
+        "@nextcloud/browser-storage": "^0.4.0",
         "@nextcloud/capabilities": "^1.2.0",
         "@nextcloud/cdav-library": "^2.1.1",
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by hamza221** [comment]
Can you split into 2 commits please ? so we can backport the fix

**Comment by GVodyanov** [comment]
> Can you split into 2 commits please ? so we can backport the fix

Good idea! Here's another PR https://github.com/nextcloud/contacts/pull/4659

File: src/store/addressbooks.js [hunk 1]
```diff
@@ -70,6 +70,72 @@ export function mapDavShareeToSharee(sharee) {
...
+		.sort((a, b) => {
+			const getContactCount = (ab) => Object.keys(ab.contacts || {}).length
+
+			const getPriorityGroup = (ab) => {
+				if (ab.id === 'contacts') return 0
+				if (ab.enabled === false) return 4
+				if (lastUsed.includes(ab.id)) return 1
+				if (ab.readOnly) return 3
+				return 2
+			}
```
ChristophWurst, line 98: coding style:

1. move this helper function out of `sortAddressbooks` so it's not redefined for every comparison
2. don't use one-line ifs. always use a { block }

src/store/addressbooks.js
```diff
@@ -70,6 +70,72 @@ export function mapDavShareeToSharee(sharee) {
...
+
+	return addressbooks
+		.slice()
+		.sort((a, b) => {
+			const getContactCount = (ab) => Object.keys(ab.contacts || {}).length
+
+			const getPriorityGroup = (ab) => {
+				if (ab.id === 'contacts') return 0
+				if (ab.enabled === false) return 4
+				if (lastUsed.includes(ab.id)) return 1
```
ChristophWurst, line 95: last use should not change the sort order but what is preselected

imagine I have this
* Personal
* Football club shared by Alice
* ACME shared by Bob

Now I add a contact for the football club

When I open the form it will already have it enabled, and the sort order is still like ^

ChristophWurst (commented):
I'm a bit skeptical about the complexity of the sorting. The original idea was to put the personal addressbook(s) first and the shared ones last. That sorting should be simple enough to be understood. We do the same in Calendar (with a divider).

The complex approach here might be a bit challenging for users.

src/store/addressbooks.js
GVodyanov: Oke

GVodyanov:
Here's how it works now, I updated the description
Screenshot From 2025-09-28 14-26-04

An issue is that apparently it looks like u can reshare addressbooks shared with u, I'll fix it (https://github.com/nextcloud/contacts/pull/4746)

ChristophWurst (commented):
Looks good :+1: 

Let's have simple unit tests for the sorting and use the `@nextcloud/browser-storage` abstraction to save data to local storage in a scoped way

PR #4658, opened 2025-08-28
feat: show a user's groups as teams

leftybournes:
Allow a user to view their groups through the teams menu. Groups can only be viewed by its members.
Reference issue here https://github.com/nextcloud/circles/issues/1941

Requires https://github.com/nextcloud/server/pull/54691 to be merged first

image

### DIFF
--- a/src/components/AppContent/CircleContent.vue
+++ b/src/components/AppContent/CircleContent.vue
@@ -5,7 +5,7 @@
 
 <template>
 	<AppContent>
-		<EmptyContent v-if="!circle" :name="t('contacts', 'Please select a team')">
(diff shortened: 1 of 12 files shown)


hamza221 (commented):
LGTM 
Only thing is that I get a 403 when accessing the group as a non admin 
`ocs/v2.php/cloud/groups/{name}/users`

Edit: was missing server update

hamza221:
Also I'm not sure if admins are supposed to see all the groups or just the user groups where they're admin 

Edit: checked with Andy and should be okay for a first iteration

AndyScherzinger:
> Also I'm not sure if admin are supposed to see all the groups or just the user groups where they're admin

Nextcloud admin, sees all LDAP groups, hence fine to see all "read-only-exposed teams equivalent LDAP groups"

Issue #4614, opened 2025-08-13
Notes field needs to be manually resized


jancborchardt:
### Is your feature request related to a problem? Please describe.

Instead of automatically resizing to its content like in other apps (and if I recall correctly, how we did before in Contacts too), you need to manually resize the Notes field to fit its content. As a result it can also be resized all over the place with width and height.

<img width="671" height="210" alt="Image" src="https://github.com/user-attachments/assets/6702a2b2-bce4-48c5-885a-c7015feaaca8" />

### Describe the solution you'd like

Using autosize, the height of the Notes field should adjust to the content while you type. The drag handle can be hidden.
This should be the case both in View and Edit mode.

marneu:
This is a BUG. This not a feature request, as this has existed for years (since the beginning). 
In the latest version 7.2.5 this field is not a text area any more thus making the texts within nearly unreadable.

I'd also second #4613 as some like our friends & family f.e. makes extensive use of information sharing within the notes field.

First requests seems to be in #4605

PR #4679, opened 2025-09-05
fix(PropertyText): textarea auto resizing

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4614

This was caused by using `ref` on `NcTextArea` instead of the actual html `textarea` element


SebastianKrupinski:
Tested. Works

image

jancborchardt:
Noice, thanks @GVodyanov ! :)

PR #4716, opened 2025-09-16
fix: duplication of address books requests

GretaD:
The problem:
The recipient info feature is causing many requests to the SAB.
It was tested with users who are actually in the SAB and artificial email addresses. Both always cause multiple requests.

When adding/removing recipients, then the component from contacts (in mail) is destroyed/mounted/destroyed/mounted and therefore the beforeMount method is called for each of re-rendering. This will dispatch getAddressbooks with the outcome of adding the same addressbook over and over again to the state.

Solution: Needs a bit of discussion because its not that easy to fix it

### DIFF
--- a/src/store/addressbooks.js
+++ b/src/store/addressbooks.js
@@ -25,6 +25,7 @@ const addressbookModel = {
 
 const state = {
 	addressbooks: [],
+	addressbooksFetched: false,
 }
 
 /**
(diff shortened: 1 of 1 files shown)


GretaD:
@kesselb you had some ideas of possible solutions. Should we add them here and discuss with the team?

kesselb:
Yeah, that's a tricky one :fearful: 

As mentioned by Greta, if you pick a recipient in Mail, that destroys and remounts the component. Since the component uses a store, the state (in our case the fetched address books) survives. In `beforeMount` there's a dispatch for `getAddressbooks`. The result is that every time the component is remounted (every time a contact is added/removed), we append the address books again, and with each contact we end up sending more requests 🙈

I think we can work around this by filtering out address books with the same URL, checking in `beforeMount` if the list of address books is not empty, keeping a state if we already fetched them, exposing a method in the store to clear it, etc. It's probably quite rare because of this remounting, but isn't it odd that `addAddressbook` has no validation to check if an address book with the same URL/ID already exists?

PR #4728, opened 2025-09-22
fix: detect addressbook ACL's properly

SebastianKrupinski:
### Summary
- Added proper detection of address book ACLs
- Adjusted Contact details edit and delete

Required for #https://github.com/nextcloud/server/issues/55169

### DIFF
--- a/src/components/ContactDetails.vue
+++ b/src/components/ContactDetails.vue
@@ -132,7 +132,7 @@
 						@click="updateContact" />
 
 					<!-- edit and save buttons -->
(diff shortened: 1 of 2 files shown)


SebastianKrupinski:
> Looks good!
> 
> Two small nag comments:
> 
>     1. make this a `fix` commit. You are not adding a feature. you are fixing the handling of ACLs
> 
>     2. relabel the commit message accordingly. the fix is not specific to recently contacted but any addressbook with fine-granular ACLs
> 
> 
> This should be backported, as a fix

Done

PR #4732, opened 2025-09-22
fix(Issue template): rename feature circles to feature teams

GVodyanov:


SebastianKrupinski (approved):

SebastianKrupinski:
@GVodyanov this probably needs back ports, but which versions?

GVodyanov:
> @GVodyanov this probably needs back ports, but which versions?

Well this is just a github thing and not in our release process so no backports necessary

Issue #4601, opened 2025-08-04
Batch actions for adding to group and moving from one address book to another


GVodyanov:
### Is your feature request related to a problem? Please describe.

Having to move large amounts of contacts manually between address books and

Having to add large amounts of contacts to groups manually

### Describe the solution you'd like

Two new actions in the multiselect menu

### Additional context

Feature request by @mejo-

PR #4761, opened 2025-10-06
feat(ContactsList): add multiaction for batch adding to group and changing addressbook

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4601

### DIFF
--- a/src/components/ContactsList.vue
+++ b/src/components/ContactsList.vue
@@ -35,6 +35,22 @@
 			<Merging :contacts="multiSelectedContacts" @finished="finishContactMerging" />
 		</NcModal>
 
+		<NcModal
+			v-if="isGrouping"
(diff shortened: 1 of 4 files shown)


SebastianKrupinski:
Tested works. But I found some glitches

A) it seems you are checking the source address book for canModify, but should be checking it for canDestory, as a move operation deletes contacts from the source and canCreate in the destiantion

Moving contacts from recents
image


B) after clicking "move" the contacts app remains in multi-select mode

After moving contacts
image

SebastianKrupinski:
Not sure if you have seen the new acls

https://github.com/nextcloud/contacts/pull/4728

SebastianKrupinski:
Tested woks good!

GVodyanov:
Let's get this out of PR purgatory

Issue #4745, opened 2025-09-27
Contact list not shown on narrow screens


stmichalke:
### Describe the bug

On my smartphone the contact list isn't shown. It works on a tablet in landscape mode.

According to the HTML source the contacts are in the list but the visibility property of the containing element ".contacts-list > div" is set to "hidden".

Everything is fine if the three-pane-layout is displayed.

### Steps to reproduce

1. Open the contacts app.
2. If necessary narrow the browser window.

### Expected behavior

The list should be visible.

### Actual behavior

It is hidden.

### Contact version

8.0.2

### Operating system

Debian 13

### PHP engine version

PHP 8.4

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

NC version is 32.0.0

maishivamhoo123:
@team can i work on this issue?

GVodyanov:
> @team can i work on this issue?

Of course! We would be really happy if you were to give us a hand. Just please make a draft PR as soon as possible and link it to this issue so that no one does repeat work. If you need any help, feel free to reach out to us through our public dev rooms.

arnowelzel:
Now the list shows, but is not scrollable.

PR #4783, opened 2025-10-09
fix(ContactsList): not being visible on small screens

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4745

| Before | After |
| - | - | 
| image  | image |


SebastianKrupinski:
@GVodyanov do you know if 7.3 is affected by this?

GVodyanov:
> @GVodyanov do you know if 7.3 is affected by this?

Nope, not affected, we good

PR #4804, opened 2025-10-19
fix(SettingsSortContacts): update it to latest NcSelect

GVodyanov:
The selecting wasn't working


SebastianKrupinski:
Tested. But there is a bug.

image

- Select a sort order
- Then save and close the settings modal
- Re-open the settings modal

The key is show and not the option label

GVodyanov:
> Tested. But there is a bug.
> image
> 
>     * Select a sort order
> 
>     * Then save and close the settings modal
> 
>     * Re-open the settings modal
> 
> 
> The key is show and not the option label

Fixed!

PR #4821, opened 2025-10-26
fix: search should have a way to clear contents

SebastianKrupinski:
### Summary
- Resolves: https://github.com/nextcloud/contacts/issues/3643


GVodyanov (approved):
Looking good!

PR #4831, opened 2025-10-28
fix: decode HTML entities in team display names

acress1:
Added a method to decode HTML entities for display names in the picker labels.
Currently, characters are displayed as HTML codes.


hamza221:
Hello, can you please amend the commit to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

acress1:
@hamza221 Done, I believe...

PR #4856, opened 2025-11-09
[stable8.0] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (1 unfixable, 21 only fixable manually using --force)


hamza221:
8.0 is not gonna get any new realases

PR #4875, opened 2025-11-23
fix(ContactDetails): wrong box shadow on inputs

GVodyanov:
| Before | After |
| - | - |
| image | image |


SebastianKrupinski (approved):

GVodyanov:
Just noticed this makes inputs disappear when editing

PR #4920, opened 2025-12-04
[stable7.3] fix(deps): Upgrade @nextcloud/vue to fix focus-trap compatibility

nfebe:
Upgrade @nextcloud/vue from ^8.27.0 to ^8.34.0 to resolve focus-trap library version mismatch with the guests app. The mismatch caused "_setPausedState is not a function" error when opening the guest invite dialog from the Teams member picker.

Without this the "Invite guest" modal does not work over the "Add team members" modal.

### DIFF
--- a/tests/setup.js
+++ b/tests/setup.js
@@ -4,6 +4,10 @@
  */
 // eslint-disable-next-line node/no-extraneous-import
 import 'regenerator-runtime/runtime'
+
+if (typeof structuredClone === 'undefined') {
+	global.structuredClone = (val) => JSON.parse(JSON.stringify(val))
+}
 import Vue from 'vue'
 
 jest.mock('@nextcloud/l10n', () => ({


DaphneMuller:
@nfebe is it a problem that the tests are failing and who would be a good reviewer for this?
@janepie will also give a review

SebastianKrupinski:
Hi @nfebe 

This probably needs a transformation exclusion like in calendar. 

https://github.com/nextcloud/calendar/blob/5f1e875a3d36291bbd394c628efdb0e0c97d8c88/package.json#L150

nfebe:
> @nfebe is it a problem that the tests are failing and who would be a good reviewer for this?

Yes, I made a fix but I will review the recently suggested workaround as well.

Should be done today.

nfebe:
> @nfebe is it a problem that the tests are failing and who would be a good reviewer for this?

Yes, I made a fix but I will review the recently suggested workaround as well.

Should be done today.

SebastianKrupinski (approved):

tests/setup.js
```diff
@@ -4,6 +4,10 @@
  */
 // eslint-disable-next-line node/no-extraneous-import
 import 'regenerator-runtime/runtime'
+
+if (typeof structuredClone === 'undefined') {
+	global.structuredClone = (val) => JSON.parse(JSON.stringify(val))
+}
```
susnux: Available since Node 17 so should not be needed?

nfebe: Tests fail without it for some reason

SebastianKrupinski:
@nfebe we still merging this?

hamza221:
@nfebe I'm closing this as 7.3 hit eol, lmk if you think it's still needed

Issue #4987, opened 2026-01-08
Design issue contacts/mail


MicKress:
### Steps to reproduce

1. Open Mail app
2. Choose contact
3. Look at the contact information in the sidebar

### Expected behavior

In the sidebar the information fields have no borders

### Actual behavior

In the sidebar the information fields have no borders

<img width="1828" height="1284" alt="Image" src="https://github.com/user-attachments/assets/9dc0c67f-bd0e-40f3-a7c0-8892e267e0c8" />

### Server operating system

Linux Debian Trixie

### Web server

Apache

### Database engine version

None

### PHP engine version

None

### Nextcloud version

32.03, Contacts App 8.20

### Updated from an older installed version or fresh install

None

### Signing status

```Text

```

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Client operating system

MacOS

### Web server error log

```Text

```

### Nextcloud log

```Text

```

### Browser log

```Text

```

### Browser log

```Text

```

hamza221:
Hello, 
Thank you for opening the issue, I can reproduce

PR #5003, opened 2026-01-16
fix(ReadOnlyContactDetails): remove box shadow

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/4987

Removes the shadow here (in mail):
image

### DIFF
--- a/src/views/ReadOnlyContactDetails.vue
+++ b/src/views/ReadOnlyContactDetails.vue
@@ -316,4 +316,8 @@ export default {
 .recipient-details-loading {
 	margin-top: 64px;
 }
+
+:deep(input) {
+	box-shadow: none !important;
+}
 </style>


SebastianKrupinski:
@GVodyanov back port to?? Stable8.3 and...?

SebastianKrupinski:
@GVodyanov 8.3 supersedes 8.2 and 8.1 and is compatible with the same NC versions... so there is no need to backport to all 3, the only other branch would 7.3 which supports NC31 and down. You can delete the 8.1 and 8.2 branches

GVodyanov:
> @GVodyanov 8.3 supersedes 8.2 and 8.1 and is compatible with the same NC versions... so there is no need to backport to all 3, the only other branch would 7.3 which supports NC31 and down. You can delete the 8.1 and 8.2 branches

I'm slow. Anyways, no 7.3 doesn't need it

PR #5010, opened 2026-01-17
Fix: improve reactivity by reducing debounce time for value changes

### LINKED ISSUE #4994: Hitting the save button too quickly doesn't save the changes
### Describe the bug

When editing a contact, changes are not saved when the save button is pushed too quickly.

### Steps to reproduce

1. Open an existing contact.
2. Edit a field (e.g. phone number).
3. Hit "Save" instantly (be fast!).
4. Reload the page.

### Expected behavior

The field reflects the edit.

### Actual behavior

The edited field has the original state before the edits.

https://github.com/user-attachments/assets/8eff0695-65e8-4834-8c9c-7e90901e06e4

### Contact version

7.3.10 and latest master

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

max65482:
Fixes #4994 by decreasing debounce time from 500 ms to 250 ms.

### DIFF
--- a/src/mixins/PropertyMixin.js
+++ b/src/mixins/PropertyMixin.js
@@ -114,7 +114,7 @@ export default {
 		updateValue: debounce(function(e) {
 			// https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier
 			this.$emit('update:value', this.localValue)
-		}, 500),
+		}, 250),
 
 		updateType: debounce(function(e) {
 			// https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier


SebastianKrupinski:
@hamza221 any objections?

hamza221:
We might need to fully remove debounce completely for button click, I don't think it makes sense debouncing at all here.
The method is used by other components (so debouce should stay for them and removed only for button click) I can update the PR if needed

SebastianKrupinski:
I think the de-bounce is on the input not the button click. It slows the update of the contact object from my reading of the code. Its part of the property mixins, it was added 8 years ago so hard to know what the purpose was

hamza221 (approved):
I took a second look at the code, It weird that most components overwrite it,  I think we actually need to clean the code here. 
Approving as a temporary fix, but not sure if we should merge. I can take a deep five next week

SebastianKrupinski:
Okay, Look in to it... then merge it if it makes sense...

hamza221:
Set to draft to avoid accidental merge until next week, @max65482 I hope you're okay with that 😊

PR #5013, opened 2026-01-19
chore(ci): Switch to setup-php@verbose version due to stuck runs

DerDreschner:
Please see https://github.com/nextcloud/mail/pull/12266 for more information about the change.

PR #5017, opened 2026-01-20
Update workflows

kesselb:
Please review carefully, as I'm not so familiar with the CI over here :warning: 

https://github.com/nextcloud/contacts/pull/5017/commits/a214571569c429f870fd8ae07b60c360f7fb2298 is **optional**. We are doing the same in mail to avoid unnecessary action bumps since https://github.com/nextcloud/mail/pull/12137

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -106,6 +106,7 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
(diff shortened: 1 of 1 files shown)


SebastianKrupinski:
Duplicate of https://github.com/nextcloud/contacts/pull/5013

SebastianKrupinski:
Also can we make the appstore-build-publish.yml a separate PR this has to be independently tested and if it needs to be reversed it will be easier to reverse

kesselb:
It appears that David missed a couple of changes in https://github.com/nextcloud/contacts/pull/5013?

SebastianKrupinski:
The new node.yml is failing... why do we need node.yml and node-test.yml?

Looks like node.yml just does a build test... and so does node-test.yml

kesselb:
> The new node.yml is failing... 

Looking at https://github.com/nextcloud/.github/pull/620 and https://github.com/nextcloud/photos/pull/3178 it seems that the lockfile is not up-to-date. I'm running `npx npm-package-lock-add-resolved`  right now and will commit the result when it's done. 

> why do we need node.yml and node-test.yml?

I don't know, probably there are there for historical reasons. That validation step, package lock ok/nok, is currently only available in node.yml though.

PR #5029, opened 2026-01-22
chore(CI): Adjust testing matrix for Nextcloud 33 on main

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


SebastianKrupinski:
@hamza221  

In my opinion, I would only run these tests on the min and max version of PHP, I can't see a case where if min passes and max passes that a in between version would fail... the config currently runs 4 versions against each database... thats 12 tests for very little benefit... 

But the choice is yours

Issue #4181, opened 2024-10-09
Group "Birthday" and "Date of death" entries together


provokateurin:
### Steps to reproduce

1. Create a contact with both birthday and date of death

### Expected behavior

These dates belong together so they should be grouped together.

### Actual behavior

The birthday is it's own group while the date of death is grouped together with other properties at the bottom.

### Contact version

5.5.3

### Operating system

Alpine

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database

PostgreSQL

hamza221:
fixed in #4377

PR #5038, opened 2026-01-24
fix(contactsDetails): remove inaccessible code

hamza221:
Looks like `name` is undefined so the component will never be rendered, 
@GVodyanov  is this needed ? I see that we already group them in `groupedPropreties`


GVodyanov:
There is an intentional exception here https://github.com/nextcloud/contacts/blob/1bd40ebc0a196a957c604653fb5648f0b0903653/src/components/ContactDetails.vue#L676

Related to https://github.com/nextcloud/contacts/issues/4181

I don't think this PR is necessary

PR #5068, opened 2026-02-03
fix(Batch): incorrect translations import

GVodyanov:
The modal for batch grouping didn't use to render (with multiple contacts selected)


SebastianKrupinski:
Which versions are affected?

Issue #5048, opened 2026-01-27
Missing accessible labels for contact detail properties


GVodyanov:
### Steps to reproduce

1. Go to contact details
2. Find any fields

### Expected behavior

They should have a label connected to the input field

### Actual behavior

The label is a heading, disconnected from the input field.

Adding placeholders is not enough to make this accessible

### PHP engine version

None

### Web server

None

### Database

None

PR #5086, opened 2026-02-08
fix: make property labels accessible

GVodyanov:
Fix https://github.com/nextcloud/contacts/issues/5048

Final result:
Screenshot From 2026-02-17 11-48-53
Screenshot From 2026-02-17 11-48-36

Changes:
- Read only mode: gray label + `p` tags for text
- Edit mode: `NcInputField` with native label, aligned `NcSelect`

Getting `NcSelect` to be the same height and aligned with the input field took some nasty custom CSS seeing as one is 34px and the other 30px, but I'm hoping this will eventually be fixed by nextcloud vue


GVodyanov:
image

GVodyanov:
image

PR #5163, opened 2026-03-06
feat: Improve circle mobile experience

pymnh:
# Changes

The circles integration in the contacts app is currently not very mobile friendly and the detail view a bit cluttered, especially on mobile. This PR tries to improve the mobile experience by ... 

- streamlining the different means to update the settings (edit title/description vs editing invite/privacy settings): 
  - Rewrite team settings from NcPopover to NcAppSettingsDialog
  - Move editable title/description to new NcAppSettingsDialog and simplify CircleDetails.vue
  - Use toggles instead of checkboxes
  - Add a "Manage settings" action to circles item in sidebar (and rename "Manage team" action to "Add members", which is more appropriate anyways)
- Some CSS fixes:
  - wrap action items
  - Emit "Team profile picture" from mobile view (circles do not have profiles pictures anyways so currently this really just wastes space - should probably be reintroduced once circles have configurable profile pictures?)
  
I changed PUBLIC_CIRCLES_CONFIG so that one item is an object ({ label: string, description?: string }) instead of a string to fix the awkward overflowing checkbox for CIRCLE_CONFIG_REQUEST. Not sure if this is the cleanliest way, but it seemed straightforward to me. Another option would be to just split the strings at newlines and use everything behind `\n` as description. But that is maybe not very stable.

# Screenshots before

- Membership/privacy settings:
  image

- Editing name and description:
  image

- Sidebar actions button:
  image

-  Mobile view:
  image

# Screenshots afterwards

- Desktop circles view:
  image

- Desktop settings modal: 
  image

- Sidebar actions menu:
  image

- Mobile view:
  image image


GVodyanov:
Also @kra-mo what do you think? Are you responsible for circles too? Thanks!

kra-mo:
Thanks for the consideration, but the editing was made to be in a popover and inline on purpose. See for example https://modalzmodalzmodalz.com

Also see https://github.com/nextcloud/server/issues/56841#issuecomment-3611903918 for the actual wording that was supposed to be used inside the popover but that was never implemented… (And see that issue in general.)

The wrapping for the buttons is a nice fix, although the profile picture I'd just move to above the title so that when/if we do add pictures, it won't have to be reverted.

Toggles instead of checkboxes is also good at this point, but again, I'd keep them in the popover. It's not a particularly common pattern we have but along with the inline editing, it presents a much nicer spacial model.

PR #5170, opened 2026-03-10
Fix/lazy load team members

GVodyanov:
Previously team members were just loaded through a `v-for`, this implements a virtual list like in the contacts list

This also removes the grid and makes it a list

Side note, why does this need to be this weird? The virtual list requires a fixed area boundary so essentially this PR calculates it through DOM elements


Antreesy:
Can we backport it to stable8.3, so it lands on Nextcloud 32/33?

PR #5251, opened 2026-04-10
feat: sort favorite contact first in list

### LINKED ISSUE #105: Sort favorite contacts first in list
Just like we do in Files: https://github.com/nextcloud/server/issues/1802

This will make Favorites more integrated and more than just a filter. Whether you are in all contacts, or a specific group, favorites will always be sorted at the top (itself sorted alphabetically).

cc @nextcloud/contacts @daniellandau

GretaD:
Fixes #105 

- [x] Requires https://github.com/nextcloud/cdav-library/pull/1029 merged, and released
- [x] Requires cdav lib update

Screenshot from 2026-04-22 11-25-30

## How to test

You need to link cdav-library with contacts while in this pr: https://github.com/nextcloud/cdav-library/pull/1029

### DIFF
--- a/src/components/ContactsList/ContactsListItem.vue
+++ b/src/components/ContactsList/ContactsListItem.vue
@@ -36,6 +36,19 @@
 						v-if="(source.isMultiSelected || hoveringAvatar) && !isStatic"
 						:size="28"
 						:class="{ 'contacts-list__item-avatar-selected': source.isMultiSelected, 'contacts-list__item-avatar-hovered': !source.isMultiSelected }" />
+					<div
+						class="favorite-star"
+						:class="{ favorite: isFavorite }"
+						@click.stop="toggleFavorite">
+						<StarIcon
+							v-if="isFavorite"
+							:size="16"
+							class="favorite-icon" />
+						<StarOutlineIcon
+							v-else
+							:size="16"
+							class="favorite-icon" />
+					</div>
 				</div>
 			</template>
 			<template #subname>
@@ -50,16 +63,36 @@
 					</span>
 				</div>
 			</template>
+			<template #actions>
+				<NcActionButton
+					v-if="!isStatic"
+					@click="toggleFavorite">
+					<template #icon>
+						<StarIcon
+							v-if="isFavorite"
+							:size="20"
+							class="favorite-icon" />
+						<StarOutlineIcon
+							v-else
+							:size="20" />
+					</template>
+					{{ isFavorite ? t('contacts', 'Remove from favorites') : t('contacts', 'Add to favorites') }}
+				</NcActionButton>
+			</template>
 		</ListItem>
 	</div>
 </template>
 
 <script>
+import { showError } from '@nextcloud/dialogs'
 import {
 	NcListItem as ListItem,
+	NcActionButton,
 	NcAvatar,
 } from '@nextcloud/vue'
 import CheckIcon from 'vue-material-design-icons/Check.vue'
+import StarIcon from 'vue-material-design-icons/Star.vue'
+import StarOutlineIcon from 'vue-material-design-icons/StarOutline.vue'
 import RouterMixin from '../../mixins/RouterMixin.js'
 
 export default {
@@ -69,6 +102,9 @@ export default {
 		ListItem,
 		NcAvatar,
 		CheckIcon,
+		NcActionButton,
+		StarIcon,
+		StarOutlineIcon,
 	},
 
 	mixins: [
@@ -117,6 +153,10 @@ export default {
 	},
 
 	computed: {
+		isFavorite() {
+			return this.source.favorite
+		},
+
 		// contact is not draggable when it has not been saved on server as it can't be added to groups/circles before
 		isDraggable() {
 			return !!this.source.dav && this.source.addressbook.id !== 'z-server-generated--system' && !this.isStatic
@@ -151,6 +191,26 @@ export default {
 	},
 
 	methods: {
+		async toggleFavorite() {
+			const contact = this.$store.getters.getContact(this.source.key)
+			if (!contact) {
+				this.logger.error('Could not find contact in store', this.source.key)
+				showError(t('contacts', 'Could not update favorite status'))
+				return
+			}
+			if (!contact.dav) {
+				this.logger.error('Missing DAV object for contact', { contactKey: contact.key })
+				showError(t('contacts', 'Could not update favorite status'))
+				return
+			}
+			try {
+				await this.$store.dispatch('toggleFavorite', contact)
+			} catch (error) {
+				logger.error('Could not toggle favorite state', error)
+				showError(t('contacts', 'Could not update favorite status'))
+			}
+		},
+
 		startDrag(evt, item) {
 			evt.dataTransfer.dropEffect = 'move'
 			evt.dataTransfer.effectAllowed = 'move'
@@ -286,6 +346,7 @@ export default {
 
 .contacts-list__item-icon {
 	cursor: pointer !important;
+	position: relative;
 }
 
 .contacts-list__item-avatar {
@@ -308,5 +369,26 @@ export default {
 		color: var(--color-primary-hover);
 		background-color: var(--color-primary-light-hover);
 	}
+}
+
+.favorite-star {
+	position: absolute;
+	top: calc(var(--default-grid-baseline) * -1);
+	inset-inline-end: calc(var(--default-grid-baseline) * -1);
+	opacity: 0;
+	cursor: pointer;
+}
+
+.contacts-list__item-icon:hover .favorite-star {
+	opacity: 1;
+}
+
+.favorite-star.favorite {
+	opacity: 1;
+	color: var(--color-favorite)
+}
+
+.favorite-icon {
+	color: var(--color-favorite);
 }
 </style>
--- a/src/store/contacts.js
+++ b/src/store/contacts.js
@@ -48,9 +48,38 @@ const state = {
 	contacts: {},
 	sortedContacts: [],
 	orderKey: 'displayName',
+	// Stores contact keys that the current user has marked as favorites.
+	// TODO [backend_value]: replace initial value with data fetched from
+	// GET /apps/contacts/api/v1/favorites
+	favorites: new Set(),
 }
 
 const mutations = {
+
+	/**
+	 * Set or unset a contact as a favorite.
+	 * Called by the toggleFavorite action.
+	 *
+	 * @param {object} state the store data
+	 * @param {object} contact object
+	 * @param {string} contact key
+	 * @param {boolean} boolean state true = favorite, false = not favorite
+	 */
+	setFavorite(state, { key, value }) {
+		if (value) {
+			state.favorites.add(key)
+		} else {
+			state.favorites.delete(key)
+		}
+		state.sortedContacts.sort((a, b) => {
+			const af = state.favorites.has(a.key)
+			const bf = state.favorites.has(b.key)
+			if (af !== bf) {
+				return af ? -1 : 1
+			}
+			return sortData(a, b)
+		})
+	},
 
 	/**
 	 * Store raw contacts into state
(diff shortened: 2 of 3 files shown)


GretaD:
need to fix the storing: 
Screenshot from 2026-04-22 11-25-01

src/store/contacts.js
```diff
@@ -93,33 +122,35 @@ const mutations = {
...
 
 			// Not using sort, splice has far better performances
 			// https://jsperf.com/sort-vs-splice-in-array
 			for (let i = 0, len = state.sortedContacts.length; i < len; i++) {
-				if (sortData(state.sortedContacts[i], sortedContact) >= 0) {
+				if (
+					(state.sortedContacts[i].favorite !== sortedContact.favorite
+						? state.sortedContacts[i].favorite < sortedContact.favorite
+						: sortData(state.sortedContacts[i], sortedContact) >= 0)
+				) {
```
ChristophWurst: nit: really hard to follow this condition. can it be broken up?

src/store/contacts.js
```diff
@@ -70,6 +69,36 @@ const mutations = {
 		}, state.contacts)
 	},
 
+	/**
+	 * Store favorite state into store
+	 *
+	 * @param {object} state Default state
+	 * @param {Contact} contact Contact
+	 * @param {boolean} newStatus
```
ChristophWurst: does not exist

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -36,6 +36,15 @@
 						v-if="(source.isMultiSelected || hoveringAvatar) && !isStatic"
 						:size="28"
 						:class="{ 'contacts-list__item-avatar-selected': source.isMultiSelected, 'contacts-list__item-avatar-hovered': !source.isMultiSelected }" />
+					<div
+						class="favorite-star"
+						:class="{ favorite: isFavorite }"
+						@click="toggleFavorite">
```
ChristophWurst: check if this click could bubble up to onSelectMultiple of the parent. in that case it needs a click.stop

src/store/contacts.js
```diff
@@ -134,20 +165,29 @@ const mutations = {
...
+
+				state.sortedContacts.sort((a, b) => {
+					if (a.favorite !== b.favorite) {
+						return b.favorite - a.favorite
+					}
+					return sortData(a, b)
+				})
 			}
-		} else {
-			console.error('Error while replacing the following contact', contact)
```
ChristophWurst, line 150: keep?

ChristophWurst:
Cleaned-up the PR description so it's clear that we can not merge this before the cdav merge, release and update here

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -50,16 +59,35 @@
...
 			</template>
+			<template #actions>
+				<NcActionButton
+					v-if="!isStatic"
+					@click="toggleFavorite">
+					<template #icon>
+						<StarIcon
+							v-if="isFavorite"
+							:size="20"
+							fill-color="#FFCC00" />
```
GVodyanov: There is a specific CSS variable for favourite icons:

> `--color-favorite` #a37200 Color to mark favorites, can be used to color e.g. a star icon for favorites

https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/css.html

I feel like this would be the right place to use it if possible

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -309,4 +362,20 @@ export default {
 		background-color: var(--color-primary-light-hover);
 	}
 }
+
+.favorite-star {
+	position: absolute;
+	top: -4px;
```
GVodyanov:
```suggestion
	top: calc(var(--default-grid-baseline) * -1);
```

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -309,4 +362,20 @@ export default {
...
+	inset-inline-end: -4px;
```
GVodyanov:
```suggestion
	inset-inline-end:  calc(var(--default-grid-baseline) * -1);
```

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -36,6 +36,15 @@
...
 						:size="28"
 						:class="{ 'contacts-list__item-avatar-selected': source.isMultiSelected, 'contacts-list__item-avatar-hovered': !source.isMultiSelected }" />
+					<div
+						class="favorite-star"
+						:class="{ favorite: isFavorite }"
+						@click.stop="toggleFavorite">
+						<StarIcon
+							v-if="isFavorite"
+							:size="16"
+							fill-color="#FFCC00" />
```
GVodyanov: Same as https://github.com/nextcloud/contacts/pull/5251/changes#r3195530871

GVodyanov (approved):
Tested it and works great!

Just a few styling comments, will approve to not block.

I also noted that there is the age old issue representing itself (at least to me in firefox)

Image

Right after favouriting u get the star stuck on the side of the component until you hover over it, but that's almost more of a vue or browser bug probably, I had this on other PRs too and wasn't able to fix it.

DerDreschner (changes requested):
For me, the sorting is... strange?

https://github.com/user-attachments/assets/8ff82f9f-2f43-4724-8687-942858788059

And when I open a contact, the favorite status is being lost as it's only included in the `PROPFIND` for the address book by the cdav-library right now, not the contact itself. Could you have a look?

If necessary, I can modify the cdav-library to retrieve the favorite property as well. But as there is no star indicator on the contact itself, only the contacts list, I see no reason for retrieving it there as well.

DerDreschner (approved):
Retested and the sorting is now correct! :smile: Only two things that can be fixed in follow-ups:

- When selecting a contact marked as favorite, the contact list jumps unexpectedly so the selected contact is on the top for some reason:

https://github.com/user-attachments/assets/ade73380-13c0-4b6f-a5ec-2e0ec917366e

- Hiding the star for address books that don't have the `{DAV:}write-properties` right (see https://github.com/nextcloud/server/pull/60183#issuecomment-4388565340)

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -151,6 +186,23 @@ export default {
...
+				return
+			}
+			if (!contact.dav) {
+				console.error('Missing DAV object for contact', contact.key)
+				return
+			}
+			try {
+				await this.$store.dispatch('toggleFavorite', contact)
+			} catch (error) {
+				console.error('Could not toggle favorite state', error)
```
hamza221: Should have user feedback too, `showError`

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -151,6 +186,23 @@ export default {
...
 
 	methods: {
+		async toggleFavorite() {
+			const contact = this.$store.getters.getContact(this.source.key)
+			if (!contact) {
+				console.error('Could not find contact in store', this.source.key)
+				return
+			}
+			if (!contact.dav) {
+				console.error('Missing DAV object for contact', contact.key)
```
hamza221: use logger instead of console

src/store/contacts.js
```diff
@@ -274,6 +373,34 @@ const getters = {
...
+	/**
+	 * Toggle the favorite state of a contact.
+	 * Updates the store
+	 *
+	 * @param {object} context the store mutations
+	 * @param {string} contact the contact key to toggle
+	 */
+	async toggleFavorite(context, contact) {
+		if (!contact.dav) {
+			console.error('Missing DAV object for contact', contact.key)
```
hamza221: Maybe throw exceptions instead, so the user is aware

src/store/contacts.js
```diff
@@ -274,6 +373,34 @@ const getters = {
...
+			return
+		}
+
+		const oldValue = contact.dav.favorite || false
+		const newValue = !oldValue
+
+		try {
+			contact.dav.favorite = newValue
+			context.commit('updateContactFavorite', contact)
+			await contact.dav.updateProperties()
```
hamza221: These 2 have to be flipped, so the star is not added when the change is not saved

hamza221 (changes requested):

hamza221 (changes requested):
Image
IMO the button shouldn't show for readonly contacts

ChristophWurst:
Read-only is fine. We want the feature to be usable for the SAB and for read-only shared ABs. It should only not be usable for the recently contacted, because that data is volatile.

src/store/contacts.js
```diff
@@ -70,6 +69,50 @@ const mutations = {
...
+					return 0
+				}
+				if (!a.value) {
+					return 1
+				}
+				if (!b.value) {
+					return -1
+				}
+				return a.value.localeCompare(b.value)
+			})
```
ChristophWurst: this …

src/store/contacts.js
```diff
@@ -134,17 +186,44 @@ const mutations = {
...
+						return 0
+					}
+					if (!a.value) {
+						return 1
+					}
+					if (!b.value) {
+						return -1
+					}
+					return a.value.localeCompare(b.value)
+				})
```
ChristophWurst: … and this …

src/store/contacts.js
```diff
@@ -215,10 +294,30 @@ const mutations = {
...
+					return 0
+				}
+				if (!a.value) {
+					return 1
+				}
+				if (!b.value) {
+					return -1
+				}
+				return a.value.localeCompare(b.value)
+			})
```
ChristophWurst: … are duplicates of this, no? extract it to a shared function so it doesn't go out of sync

hamza221:
> Read-only is fine. We want the feature to be usable for the SAB and for read-only shared ABs. It should only not be usable for the recently contacted, because that data is volatile.

So it's supposed to work for SAB cards ? 
I'll test again cause it didn't work when testing

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -151,6 +186,24 @@ export default {
 	},
 
 	methods: {
+		async toggleFavorite() {
+			const contact = this.$store.getters.getContact(this.source.key)
+			if (!contact) {
+				console.error('Could not find contact in store', this.source.key)
```
hamza221: all `console.error` should be `logger.error` instead

hamza221:
> > Read-only is fine. We want the feature to be usable for the SAB and for read-only shared ABs. It should only not be usable for the recently contacted, because that data is volatile.
> 
> So it's supposed to work for SAB cards ? I'll test again cause it didn't work when testing

Works fine btw 👏🏼

DerDreschner:
> Works fine btw 👏🏼

Thanks for the feedback! Makes my blood pressure going down a lot :smile:

hamza221:
requires #5310

src/store/contacts.js
```diff
@@ -274,6 +345,33 @@ const getters = {
...
+	 * Toggle the favorite state of a contact.
+	 * Updates the store
+	 *
+	 * @param {object} context the store mutations
+	 * @param {string} contact the contact key to toggle
+	 */
+	async toggleFavorite(context, contact) {
+		if (!contact.dav) {
+			throw new Error(`Missing DAV object for contact ${contact.key}`)
+		}
```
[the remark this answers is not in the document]
hamza221: valid

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -50,16 +59,35 @@
...
 <script>
 import {
 	NcListItem as ListItem,
+	NcActionButton,
 	NcAvatar,
 } from '@nextcloud/vue'
 import CheckIcon from 'vue-material-design-icons/Check.vue'
+import StarIcon from 'vue-material-design-icons/Star.vue'
+import StarOutlineIcon from 'vue-material-design-icons/StarOutline.vue'
 import RouterMixin from '../../mixins/RouterMixin.js'
```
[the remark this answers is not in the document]
hamza221: also valid

src/components/ContactsList/ContactsListItem.vue
```diff
@@ -36,6 +36,15 @@
...
 						:class="{ 'contacts-list__item-avatar-selected': source.isMultiSelected, 'contacts-list__item-avatar-hovered': !source.isMultiSelected }" />
+					<div
+						class="favorite-star"
+						:class="{ favorite: isFavorite }"
+						@click.stop="toggleFavorite">
+						<StarIcon
+							v-if="isFavorite"
+							:size="16"
+							class="favorite-icon" />
+					</div>
```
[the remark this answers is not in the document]
hamza221: fyi @GVodyanov

ChristophWurst:
> requires #5310

Showtime

hamza221 (approved):

jancborchardt (commented):
Very nice, thank you! :)

PR #5265, opened 2026-04-20
fix: Sort by "last modified" (issue #5158)

### LINKED ISSUE #5158: Bug: Sort by "last modified" is not working
### Describe the bug

The Sort option "last modified" is not working.
When I change sort order to _last modified_ and switch e.g. between _All contacts_ and _Not grouped_, I get this error:
```
Unable to retrieve the contact from the server, please check your network connection.
```
If I refresh the page instead, Sort setting is reset to previous value.

### Steps to reproduce

1. Go to `Contacts` in WebUI
2. Click on `Contact settings`
3. Set sort contacts to `Last modified`
4. Close settings and:
   4a. Switch between e.g. between _All contacts_ and _Not grouped_ and you'll see this error:
   ```
   Unable to retrieve the contact from the server, please check your network connection.`
   ```
   4b. Reload the site and the Sort settings will be reset to previous setting.

### Expected behavior

Sort by _last modified_ should work

### Actual behavior

Sort by _last modified_ is producing an error message and is not saved.

### Contact version

v8.4.0

### Operating system

TrueNAS SCALE 25.10

### PHP engine version

PHP 8.4

### Web server

Nginx

### Database

PostgreSQL

### Additional info

Nextcloud: v33.0.0

v3DJG6GL:
## Summary
Fix #5158

With sort order set to **Last modified**:
- Switching between groups (e.g. *All contacts* ↔ *Not grouped*) showed an error:
  ```
  Unable to retrieve the contact from the server, please check your network connection.
  ```
- After reloading the page, the sort setting was reset to the previous value.
- This issue was described in #5158

Underlying issue: sorting crashed whenever any contact had a missing or malformed `REV` (last-modified) value:
The code called a date method on every contact's `REV`. If `REV` was missing or not a valid date, it threw an error.

## Changes
- Read each contact's REV once when contacts are first loaded from the server, and store a plain timestamp (or `null` if it's missing/unreadable).
- Contacts with no valid `REV` sort to the end instead of crashing the list.
- Tolerate malformed date strings (e.g. wrong format, missing seconds) by parsing them defensively.
- Make the `rev` getter return `null` on error instead of throwing, so one contact with a broken `REV` can't crash the component that reads it.
- Stop fabricating `REV = now` when one is missing (both in the `Contact` constructor and in a background auto-"fix"). The old behavior made contacts jump to the top every time they were touched.
- Don't re-sort the whole list when a contact is opened. Its stored `REV` doesn't change just because it was viewed.
- Replaced custom scroll math with the list library's built-in scroll-into-view, which prevents visible jumps when the list updates.

## LLM Disclosure
*Developed with assistance from Claude Code (Opus 4.7, 1M context).
The LLM helped me identify the related files, code parts and proposed possible code modifications and fixes.*
*I've manually reviewed, the code the best I could.*
***A proper code review is required in any case...***


v3DJG6GL:
@GVodyanov could you take a look at this PR please? I don't want to and don't have the capacity to keep this PR up-to-date over months...
Thanks!

src/components/ContactsList.vue
```diff
@@ -363,24 +363,7 @@ export default {
...
-				scroller.scrollToIndex(index)
-			} else if (item) {
-				const itemHeight = scroller.getItemSize(index)
-				const pos = itemBoundingRect.y + itemHeight - (this.$el.offsetHeight + 50)
-				if (pos > 0) {
-					// Item is below the current scroll window (or partly overlapping)
-					scroller.scrollTo(scroller.scrollOffset + pos)
-				}
-			}
+			this.$refs.scroller.scrollToIndex(index, { align: 'nearest' })
```
hamza221, line 368: How is this related to the fix

src/models/contact.js
```diff
@@ -204,6 +197,41 @@ export default class Contact {
...
 
+	/**
+	 * REV as a unix timestamp (seconds), or null if missing/unparseable.
+	 * Bypasses ICAL.Time parsing so malformed-but-salvageable basic-ISO
+	 * strings ("20260312T192500Z", "T1925Z", trailing "T19:25Z") still sort.
+	 *
+	 * @readonly
+	 * @memberof Contact
+	 */
+	get revTimestamp() {
```
hamza221, line 213: This would be really hard to maintain

hamza221 (commented):
Thank you for the PR, The fix looks really promising

hamza221:
Could you also follow our AI-policy for commit messages https://github.com/nextcloud/contacts/blob/main/AGENTS.md#commit-message-format . 
And squash your commits into one => this could be done at the end before merge.

src/components/ContactsList.vue
v3DJG6GL: With **Last modified** sorting, the list re-orders whenever a REV of a contact changes, so the selected contact can move within the list.
The previous scroll logic measured DOM rects to decide whether and where to scroll. This caused visible jumps when it ran against the virtual list mid-update, since the measurements were stale.

```scrollToIndex(index, { align: 'nearest' })``` implements exactly what the manual code was trying: scroll only if the item isn't fully visible, minimal movement - but based on the library internal state, so it should stay correct while the list updates.

File: src/models/contact.js (hunk 2 above)
Comment (v3DJG6GL) [reply src/models/contact.js:213]: I've adjusted it - it's now a single normalization instead of chained regex replacements, and a comment:
**ical.js** reformats REV values by fixed-position slicing, so values missing seconds arrive with misplaced colons (e.g. _2026-03-12T19::25Z_) 
Additionally, I've added unit tests covering every supported format and the **ical.js** artifacts.

File: src/models/contact.js [hunk 3]
```diff
@@ -204,6 +197,35 @@ export default class Contact {
...
+		if (typeof raw !== 'string') {
+			return null
+		}
+		// The jCal value may be basic ISO 8601 ("20260312T192500Z"), extended
+		// ("2026-03-12T19:25:00Z"), or carry misplaced separators from ical.js
+		// reformatting a REV with missing seconds ("2026-03-12T19::25Z").
+		// Strip all separators and rebuild the extended format for Date.parse.
+		// Seconds default to 00, a missing zone is treated as UTC.
+		const match = raw.replace(/[-:]/g, '')
+			.match(/^(\d{4})(\d{2})(\d{2})T(\d{2})(\d{2})(\d{2})?(Z|[+-]\d{4})?$/)
```
GVodyanov: raw.replace(/[-:]/g, '') strips every hyphen, including the sign of a negative UTC offset. So a REV with a negative offset loses its -, fails the match, and returns null (the contact then sorts as if it had no REV). Positive offsets are fine because + isn't stripped, which is why the +0100 test      
  passes and this slips through.                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                
  "20260312T202500+0100"      -> 1773343500   ✅ (tested)                                                                                                                                                                                                                                                       
  "20260312T142500-0500"      -> null          ❌                                                                                                                                                                                                                                                               
  "2026-03-12T14:25:00-05:00" -> null          ❌                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                
  All three are the same instant; the negative ones just lose the offset sign during the strip.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                
  In practice REV is stored in UTC (Z), so real data rarely hits this, but the regex clearly intends to support offsets ([+-]\d{4}) and only the negative half silently fails.

SebastianKrupinski: Good catch!

v3DJG6GL: Thanks, I've adjusted the regex:
The zone suffix is now split off before the separator strip. Added both failing cases as unit tests.

hamza221:
Needs conflict resolution

v3DJG6GL:
> Needs conflict resolution

conflicts resolved :)

v3DJG6GL:
any update on this?

hamza221:
Could you please check if this fixes the bug mentioned #5708

PR #5266, opened 2026-04-20
Security: Potential denial of service from unbounded image decoding

tomaioo:
## Summary

Security: Potential denial of service from unbounded image decoding

## Problem

**Severity**: `Medium` | **File**: `lib/Service/ImageResizer.php:L22`

`resizeImage()` loads arbitrary image bytes into memory and decodes them without checking input size or dimensions first. If attacker-controlled or remote-fetched image data is very large or crafted, this may cause excessive memory/CPU consumption and degrade service availability.

## Solution

Validate input size before decoding (e.g., maximum bytes), enforce pixel/dimension limits, and fail fast on oversized payloads. Prefer safe decoding paths with resource limits and consider rejecting unsupported/complex image formats.

## Changes

- `lib/Service/ImageResizer.php` (modified)

### DIFF
--- a/lib/Service/ImageResizer.php
+++ b/lib/Service/ImageResizer.php
@@ -14,14 +14,28 @@ use OCP\Image;
 class ImageResizer {
 	public const RESIZE_MAX_X = 256;
 	public const RESIZE_MAX_Y = 256;
+	public const MAX_INPUT_BYTES = 5 * 1024 * 1024;
+	public const MAX_INPUT_PIXELS = 4096 * 4096;
 
 	/**
 	 * @param string $socialData
 	 * @return null|string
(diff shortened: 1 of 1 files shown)


hamza221 (changes requested):
Hello, 
I'm not sure if this is needed : 
https://github.com/nextcloud/contacts/pull/5124 Whitelisted image formates were added in this PR https://github.com/nextcloud/contacts/pull/5124 
And I think we can trust Instagram, Mastodon, Tumblr, Diaspora, Xing, Telegram, Gravatar Not send us Huge images  
But even if we want to add the limit I would rather have a configurable limit than a hardcoded one

Issue #3254, opened 2023-02-28
Show users circle name validation errors


susnux:
### Steps to reproduce

1. Go to contacts app
2. Click on create new circle
3. Enter a name (length 1 or 2)

### Expected behavior

Show a meaningful error or better: Validate the input before sending it so users know that there is a minimum length

### Actual behavior

Obscure error message "Error while creating circle A"

### Server operating system

Docker

### Web server

Apache

### Database engine version

MySQL

### PHP engine version

PHP 8.0

### Nextcloud version

master branch

### Updated from an older installed version or fresh install

None

PR #5292, opened 2026-04-29
Fix: enforce minimum team name length (>=3) in UI

JSchut96:
**Issue**
Based on: https://github.com/nextcloud/contacts/issues/3254

The backend enforces a minimum team name length of 3 characters and returns a 400 Bad Request when this requirement is not met (e.g. "Circle name is too short").

However, the UI does not properly handle or display the error message from the server, leaving the user without a clear indication of why their submission failed.

**Proposed Solution**
Instead of focusing on improving error message display, this fix prevents invalid team names from being submitted altogether.

The solution:

- Ensures the "Create team" button is only enabled when the team name meets the length requirement.
- Prevents the form from being submitted via the Enter key if the name is invalid.
- Communicates the length requirement within the input placeholder and shows a helpful error/hint when the name does not meet the minimum character count.

### DIFF
--- a/src/components/EntityPicker/NewCircleIntro.vue
+++ b/src/components/EntityPicker/NewCircleIntro.vue
@@ -14,10 +14,16 @@
 				<input
 					ref="input"
 					v-model="circleName"
-					:placeholder="t('contacts', 'New team name')"
+					:placeholder="t('contacts', 'New team name (min. 3 characters)')"
 					class="entity-picker__new-input"
 					type="text"
 					@keypress.enter="onSubmit">
+
+				<p
+					v-if="circleName && isInvalidName"
+					class="entity-picker__hint">
+					{{ t('contacts', 'Name must be at least 3 characters') }}
+				</p>
 			</div>
 
 			<div class="entity-picker__content">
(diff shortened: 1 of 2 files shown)


DerDreschner (changes requested):
Hey, thank you very much for your pull request! Could you please change the code so the CI runs are green?

src/components/EntityPicker/NewCircleIntro.vue
```diff
@@ -181,6 +191,15 @@ $icon-margin: math.div($clickable-area - $icon-size, 2);
 		}
 	}
 
+	&__hint {
+		position: absolute;
+		top: 100%;
+		inset-inline-start: 0;
+		width: 100%;
+		font-size: 0.75rem;
```
GVodyanov: Hey! Thanks for your PR

Please use one of our CSS variables for the font size https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/css.html

JSchut96: Hi,
Should be fixed now.
Kind regards,
Jeroen

DerDreschner (approved):

DerDreschner:
Hey @JSchut96, could you please rebase your branch to the latest commits on main, so we can run the CI? Thanks!

PR #5293, opened 2026-04-30
feat(circle-details): allow setting avatar image for circle

cristianscheid:
* Resolves: https://github.com/nextcloud/circles/issues/2433
* Related to: https://github.com/nextcloud/circles/pull/2438

## Summary

- Adds support for setting avatar image for circle

## Screenshots
| | |
|-|-|
|Screenshot from 2026-05-12 07-15-12|Screenshot from 2026-05-12 07-15-21|
|Screenshot from 2026-05-12 07-15-44 Screenshot from 2026-05-18 13-46-48|Screenshot from 2026-05-18 13-47-06|

### DIFF
--- a/src/components/CircleDetails.vue
+++ b/src/components/CircleDetails.vue
@@ -45,7 +45,7 @@
 							label="Description"
 							:maxlength="1024" />
 					</div>
-					<div class="circle-avatar-buttons-wrapper">
+					<div v-if="avatarSupported" class="circle-avatar-buttons-wrapper">
 						<NcButton
 							v-if="isEditing"
 							:disabled="loadingAvatar"
@@ -385,7 +385,9 @@ export default {
 		const avatarList = ref()
 		const { width } = useElementSize(avatarList)
 		const avatarAccept = VALID_MIME_TYPES.join(',')
-		return { avatarList, width, avatarAccept }
+		const nextcloudMajorVersion = parseInt(window.OC.config.version.split('.')[0])
+		const avatarSupported = nextcloudMajorVersion >= 34
+		return { avatarList, width, avatarAccept, avatarSupported }
 	},
 
 	data() {
@@ -460,6 +462,9 @@ export default {
 		},
 
 		displayAvatarUrl() {
+			if (!this.avatarSupported) {
+				return undefined
+			}
 			if (this.pendingAvatarAction === AVATAR_ACTIONS.DELETE) {
 				return undefined
 			}
@@ -576,11 +581,13 @@ export default {
 	watch: {
 		'circle.id': {
 			handler() {
-				this.cancelSetAvatar()
-				this.clearPendingAvatar()
 				this.isEditing = false
 				this.fetchTeamResources()
-				this.loadAvatarUrl()
+				if (this.avatarSupported) {
+					this.cancelSetAvatar()
+					this.clearPendingAvatar()
+					this.loadAvatarUrl()
+				}
 			},
 
 			immediate: true,
@@ -907,6 +914,9 @@ export default {
 		},
 
 		async loadAvatarUrl() {
+			if (!this.avatarSupported) {
+				return
+			}
 			try {
 				const response = await axios.get(
 					generateOcsUrl(`/apps/circles/circles/${this.circle.id}/avatar`),
@@ -974,7 +984,7 @@ export default {
 			}
 
 			// Upload pending avatar if set
-			if (this.pendingAvatarAction === AVATAR_ACTIONS.SET && this.pendingAvatarBlob) {
+			if (this.avatarSupported && this.pendingAvatarAction === AVATAR_ACTIONS.SET && this.pendingAvatarBlob) {
 				this.loadingAvatar = true
 				const formData = new FormData()
 				formData.append('file', this.pendingAvatarBlob)
@@ -991,7 +1001,7 @@ export default {
 			}
 
 			// Delete avatar if marked for deletion
-			if (this.pendingAvatarAction === AVATAR_ACTIONS.DELETE) {
+			if (this.avatarSupported && this.pendingAvatarAction === AVATAR_ACTIONS.DELETE) {
 				this.loadingAvatar = true
 				try {
 					await axios.delete(generateOcsUrl(`/apps/circles/circles/${this.circle.id}/avatar`))
--- a/src/components/CircleDetails.vue
+++ b/src/components/CircleDetails.vue
@@ -8,7 +8,9 @@
 		<div class="circle-details-grid" :class="{ 'is-editing': isEditing }">
 			<div class="circle-details__header-wrapper">
 				<div class="circle-details-grid__avatar">
+					<NcLoadingIcon v-if="loadingAvatar" :size="75" />
 					<Avatar
+						v-else
 						:disable-tooltip="true"
 						:display-name="circle.displayName"
 						:is-no-user="true"
@@ -46,6 +48,7 @@
 					<div class="circle-avatar-buttons-wrapper">
 						<NcButton
 							v-if="isEditing"
+							:disabled="loadingAvatar"
 							@click="openLocalFilePicker">
 							<template #icon>
 								<TrayArrowUpIcon :size="20" />
@@ -54,6 +57,7 @@
 						</NcButton>
 						<NcButton
 							v-if="isEditing"
+							:disabled="loadingAvatar"
 							@click="openFilePicker">
 							<template #icon>
 								<FolderIcon :size="20" />
@@ -62,6 +66,7 @@
 						</NcButton>
 						<NcButton
 							v-if="isEditing && displayAvatarUrl !== undefined"
+							:disabled="loadingAvatar"
 							@click="removeAvatar">
 							<template #icon>
 								<TrashCanOutlineIcon :size="20" />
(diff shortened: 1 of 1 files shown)


cristianscheid:
@kra-mo following [your suggestion](https://github.com/nextcloud/circles/pull/2438#issuecomment-4221875497):

I positioned the avatar buttons beneath the name and description in edit mode and styled them closer to the reference in [this issue](https://github.com/nextcloud/server/issues/57008). The PR description has been updated with screenshots and a short video showing the current state.

Two questions I'd appreciate your input on:

1. How should we handle the button widths?
2. Right now, selecting an image opens the cropper and clicking "Set as team picture" saves immediately to the backend. Should it instead only save when the main edit mode "Save" button is clicked?

Besides that, I'm also aware that probably a spacing between the buttons is needed and also some margin from the main edit mode "Cancel" and "Save" buttons. Just waiting for another round of input before moving forward.

Any other thoughts are welcome :)

kra-mo:
> How should we handle the button widths?


The widths should definitely be homogeneous.

They should also have margins between them.

> Right now, selecting an image opens the cropper and clicking "Set as team picture" saves immediately to the backend. Should it instead only save when the main edit mode "Save" button is clicked?

Since there is a "Cancel" button, the second behavior makes more sense.

cristianscheid:
@kra-mo thanks for the input, really appreciated! I've adopted your suggestions and updated the implementation. Also updated the main comment of this PR with new screenshots and a short video demoing the current state.

src/components/CircleDetails.vue
```diff
@@ -462,12 +563,22 @@ export default {
 		'circle.id': {
 			handler() {
 				this.fetchTeamResources()
+				this.loadAvatarUrl()
```
pringelmann: If I start editing one team, pick a picture, then click another team in the sidebar before saving, the pending blob carries over and saves to the wrong team. Watcher should call `clearPendingAvatar()` (and probably reset isEditing too) when the id actually changes.

src/components/CircleDetails.vue
```diff
@@ -42,6 +43,54 @@
...
+					</div>
+					<!-- Use v-show to ensure early cropper ref availability -->
+					<div v-show="showCropper" class="circle-avatar-cropper-wrapper">
+						<VueCropper
+							ref="cropper"
+							class="circle-avatar-cropper"
+							v-bind="cropperOptions" />
+						<div class="circle-avatar-cropper-buttons">
+							<NcButton @click="cancelSetAvatar">
+								{{ t('settings', 'Cancel') }}
```
pringelmann: These three (Cancel, Apply, Error cropping avatar picture) are still in the settings domain. I might be wrong but I suspect won't be picked up by the contacts gettext catalog, so they stay English regardless of locale. Should be `t('contacts', ...)` maybe?

src/components/CircleDetails.vue
```diff
@@ -647,6 +757,98 @@ export default {
...
 
+		openLocalFilePicker() {
+			this.$refs.avatarInput.value = null
+			this.$refs.avatarInput.click()
+		},
+
+		async onAvatarInputChange(e) {
+			try {
+				const file = e.target.files[0]
+				if (!['image/png', 'image/jpeg'].includes(file.type)) {
```
pringelmann: You already have `VALID_MIME_TYPES` at the top of the file, can we use it here too? Otherwise the picker filter and this check can drift.

src/components/CircleDetails.vue
```diff
@@ -647,6 +757,98 @@ export default {
...
+				showError(t('contacts', 'Error picking team picture'))
+			}
+		},
+
+		setAvatar() {
+			this.showCropper = false
+			const canvasData = this.$refs.cropper.getCroppedCanvas()
+			const scaleFactor = canvasData.width > 512 ? 512 / canvasData.width : 1
+
+			this.$refs.cropper.scale(scaleFactor, scaleFactor).getCroppedCanvas().toBlob(async (blob) => {
```
pringelmann: Cropper has `getCroppedCanvas({ maxWidth: 512, maxHeight: 512 })` for exactly this, no need to mutate the cropper state. Same approach is used in `ContactDetailsAvatar.vue` around line 425.

src/components/CircleDetails.vue
```diff
@@ -42,6 +43,54 @@
...
+							v-if="isEditing && displayAvatarUrl !== undefined"
+							@click="removeAvatar">
+							<template #icon>
+								<TrashCanOutlineIcon :size="20" />
+							</template>
+							{{ t('contacts', 'Delete picture') }}
+						</NcButton>
+						<input
+							ref="avatarInput"
+							type="file"
```
pringelmann, line 78: Worth setting `accept="image/png,image/jpeg"` here so the native OS dialog only shows valid files. Cheap UX improvement, and the MIME check stays as the actual gate. Not a blocker for this PR though :)

One way to do it could be in setup:

```ts
setup() {
    const avatarList = ref()
    const { width } = useElementSize(avatarList)
    return { avatarList, width, avatarAccept: VALID_MIME_TYPES.join(',') }
}
```

and then

```vue
<input
    ref="avatarInput"
    type="file"
    :accept="avatarAccept"
    class="hidden-visually"
    @change="onAvatarInputChange">
```

src/components/CircleDetails.vue
```diff
@@ -725,6 +947,32 @@ export default {
...
+			if (this.pendingAvatarAction === AVATAR_ACTIONS.SET && this.pendingAvatarBlob) {
+				const formData = new FormData()
+				formData.append('file', this.pendingAvatarBlob)
+				try {
+					await axios.post(generateOcsUrl(`/apps/circles/circles/${this.circle.id}/avatar`), formData)
+					this.clearPendingAvatar()
+					this.loadAvatarUrl()
+				} catch {
+					console.error('Unable to save avatar picture')
+					errors.push('avatar')
```
pringelmann, line 997: This is untranslated, but so are the existing `name` and `description` fields. If you have time it could be good to fix them, but not blocking for PR

src/components/CircleDetails.vue
```diff
@@ -258,6 +312,22 @@ import MemberList from './MemberList/MemberList.vue'
...
+
+const AVATAR_ACTIONS = Object.freeze({
+	SET: 'set',
+	DELETE: 'delete',
+})
+
+const picker = getFilePickerBuilder(t('contacts', 'Choose a team picture'))
+	.setMultiSelect(false)
+	.setMimeTypeFilter(VALID_MIME_TYPES)
+	.setType(1)
```
pringelmann: afaik `FilePickerType.Choose` is exported from `@nextcloud/dialogs`. Reads better and won't silently break if the underlying values ever shift

src/components/CircleDetails.vue
cristianscheid: looking through contacts app codebase, I couldn't spot any translated string inside console errors, so I kept the existing pattern

src/components/CircleDetails.vue
cristianscheid: great catch! adjusted.

src/components/CircleDetails.vue
cristianscheid: yes you are right, thanks for catching this one.

src/components/CircleDetails.vue
cristianscheid: great tip, wasn't aware of that. Updated the code using the approach like in `ContactDetailsAvatar.vue`.

src/components/CircleDetails.vue
cristianscheid: adjusted.

src/components/CircleDetails.vue
cristianscheid: great tip! Adjusted as well.

src/components/CircleDetails.vue
cristianscheid: adjusted

pringelmann (approved):
🚀 nice work!

src/components/CircleDetails.vue
pringelmann: @cristianscheid a simpler and more idiomatic alternative is to just set the ID as a `key` on the component:

`<CircleDetails v-else :key="circle.id" :circle="circle" />`

This will force each details component instance to re-mount based on the ID. This introduces a re-mount cost but that is trivial in this case. (only really an issue for heavy components that re-render very often)

One benefit of this approach is that it prevents any future internal state from leaking because it was not correctly cleaned up in the watcher.

kra-mo (changes requested):
Having the Cancel/Save buttons in such close proximity is quite confusing, especially here:

Image

Could this be in a dialog instead to make it a bit more focused?

cristianscheid:
> Could this be in a dialog instead to make it a bit more focused?

@kra-mo good idea! I've adjusted the PR accordingly, here are some screenshots:

| | |
|-|-|
|Screenshot from 2026-05-18 13-46-48|Screenshot from 2026-05-18 13-47-06|

kra-mo (approved):

kra-mo:
@cristianscheid I was talking about a layout similar to this for the dialog:

image

And I'd also like if the input fields for the title and description could have a reduced maximum width, and the buttons there could have the same width:

image

Using [NcFormGroup](https://nextcloud-vue-components.netlify.app/#/Components/NcFormGroup)/[NcFormBox](https://nextcloud-vue-components.netlify.app/#/Components/NcFormBox).

Neither are a blocker, but it would be nice to have these in a follow-up.

Issue #4454, opened 2025-05-08
Adding line manager property does not work


kesselb:
### Describe the bug

Adding a new line manager property does not work

### Steps to reproduce

1. Open the browser's console
2. Add a contact
3. Click "Add more info"
4. Select "Line Manager"
5. No input field for line manager and error in console

### Expected behavior

Should work ;)

### Actual behavior

It does not 🙈 

### Contact version

main

### PHP engine version

None

### Web server

None

### Database

None

invisiblehannes-stack:
Any progress in this case? Still not working.

Steps to reproduce
Open the browser's console
Add a contact
Click "Add more info"
Select "Line Manager"
No input field for line manager and error in console

When copy an existing contact, its not possible to use the "manager field"

Is there any other way to set up an organizational chart?

0KayKay:
Can verify this on contacts app version 8.3.6, checked on edge and firefox.

It's adding an empty div tag without an input field in it and creating a warning from ContactDetailsProperty.vue: "No focusable element found for property x-managersname". Existing contacts with the manager property filled out can be edited, so I think it's probably not a data problem but a frontend bug when adding the UI element.

This started with an automatic update of our managed Hetzner instance to Nextcloud Hub 25 Autumn (32.0.6). Sadly I have no idea whether the contacts app was also automatically updated (which is plausible) and can't provide any information on the last working version before that.

Since this impacting some some of our day to day processes, I'd be happy to hear a word from the devs. Feel free to reach out to me for further info for debugging and replication of the bug.

nfk:
@hamza221 fixed and released. Thanks! You can close this issue.

hamza221:
https://github.com/nextcloud/contacts/pull/5304

madbob:
The PR has been merged in `stable8.3` but not in `main`: the issue still exists.

PR #5304, opened 2026-05-07
fix(contacts): handle line manager property creation

nfk:
Adding the Line manager field created an empty x-managersname property and then immediately tried to focus the newly rendered editor. When the property had no matching contact value, the select received an undefined value and no focusable field was available.

This produced the following JavaScript errors:

- No focusable element found for property x-managersname
- TypeError: can't access property "toLowerCase", this.localValue is undefined

Keep empty manager properties represented as an empty string, guard select matching before lowercasing the local value, and store the only available manager using the vCard display name plus UID parameter instead of the internal contact key.

Fixes nextcloud/contacts#4454

### DIFF
--- a/src/components/ContactDetails/ContactDetailsAddNewProp.vue
+++ b/src/components/ContactDetails/ContactDetailsAddNewProp.vue
@@ -74,6 +74,7 @@
 </template>
 
 <script>
+import { showWarning } from '@nextcloud/dialogs'
 import { NcActionButton as ActionButton, NcActions as Actions } from '@nextcloud/vue'
 import ICAL from 'ical.js'
 import ChevronLeft from 'vue-material-design-icons/ChevronLeft.vue'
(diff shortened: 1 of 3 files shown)


src/components/ContactDetails/ContactDetailsAddNewProp.vue
```diff
@@ -207,12 +207,20 @@ export default {
...
 			}
 
+			let shouldFocusProp = true
 			if (id === 'x-managersname') {
 				const others = this.otherContacts(this.contact)
+				if (others.length === 0) {
+					this.moreActionsOpen = false
+					return
+				}
 				if (others.length === 1) {
```
hamza221, line 219: We shouldn't fail silently here 
Maybe add a `showError` or a `showWarning`

nfk: Thanks for the input, added

nfk:
@hamza221 @GVodyanov could you consider this MR? or let me know if some other changes are required. Thanks.

hamza221:
> @hamza221 @GVodyanov could you consider this MR? or let me know if some other changes are required. Thanks.

Hello, Really sorry for the delay, I will give it another review this week

hamza221 (approved):

PR #5368, opened 2026-06-03
fix: drop stable32 from tests and adjust php version

hamza221:


SebastianKrupinski (approved):

appinfo/info.xml
```diff
@@ -43,7 +43,7 @@
 	<screenshot>https://raw.githubusercontent.com/nextcloud/screenshots/master/apps/Contacts/contacts.png</screenshot>
 
 	<dependencies>
-		<php min-version="8.1" max-version="8.5" />
+		<php min-version="8.2" max-version="8.5" />
```
ChristophWurst, line 46: This kind of change in a *chore* feels sneaky. Make it a fix.

ChristophWurst: Reasoning: chore would be hidden from the changelog. Fix is listed. This is a change that should be communicated.

GVodyanov (approved):

appinfo/info.xml
hamza221: changed to fix

PR #5370, opened 2026-06-04
Bug: Federated Cloud ID shown as array object

T0mWz:
# Summary
When an empty Federated Cloud ID is added, the record is shown as empty array.
Also returns a JSON error in de browser console.

# Fix
Make not an array but string object.

FedCloudID

### DIFF
--- a/src/models/rfcProps.js
+++ b/src/models/rfcProps.js
@@ -90,7 +90,7 @@ const properties = {
 		readableName: t('contacts', 'Federated Cloud ID'),
 		force: 'text',
 		defaultValue: {
-			value: [''],
+			value: '',
 			type: [defaultProfileState],
 		},
 		options: [


T0mWz:
Related javascript error;
```
[Vue warn]: Invalid prop: type check failed for prop "value". Expected String with value "", got Array  
  at <PropertyText ref="component" select-type=null onUpdate:selectType=fn  ... > 
  at <ContactDetailsProperty key="0-NmMyMmE2ODktODlkOS00Y2U0LTgxODktOThiNTMxNjZkZWU5fmNvbnRhY3Rz-cloud" is-first-property=true is-last-property=false  ... > 
  at <NcAppContentDetails> 
  at <ContactDetails contact-key="NmMyMmE2ODktODlkOS00Y2U0LTgxODktOThiNTMxNjZkZWU5fmNvbnRhY3Rz" contacts= 
Array(8) [ {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…} ]
```

T0mWz:
Hi @hamza221 can you review this for me? :-)

PR #5371, opened 2026-06-04
Feat: Make Federated Cloud ID by default visible in Contacts Card

T0mWz:
# Summary

Make Federated Cloud ID by default visible in Contacts Card.
This simplifies federated sharing, as the field is visible by default, making it easier to add contacts along with their federated cloud ID.

FedCloudID by default visible

### DIFF
--- a/src/models/rfcProps.js
+++ b/src/models/rfcProps.js
@@ -91,7 +91,7 @@ const properties = {
 		force: 'text',
 		default: true,
 		defaultValue: {
-			value: [''],
+			value: '',
 			type: [defaultProfileState],
 		},
 		options: [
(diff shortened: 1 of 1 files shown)


jancborchardt (commented):
I was thinking it would already be great (and a good middle ground) to move the Federated Cloud ID one level up, from "More fields" directly to the short list of "Add more info" elements.

I think I would still prefer that instead of having the Federated Cloud ID directly shown, as we really reserve that for only the most essential of information.

What do you think @nimishavijay?

nimishavijay:
Agreed, if when someone imports a contact it already has the federated cloud ID field, we should of course show it, but when creating a new contact this information seems quite technical.

PR #5378, opened 2026-06-09
fix: handle ical.js group parameter form to prevent custom label corruption on save/re-fetch cycle

### LINKED ISSUE #5302: Custom email/phone labels not working
### Describe the bug

Setting custom labels for a phone number and email address does not work currently. While it is possible (after retries) to type in and select the label, after saving and reloading the default "Private" appears in the UI. Checking with a different app (DAVx5 -> Fossify Contatcs) those custom labels are set. Editing the contact again and selecting "work" instead of "private" (which already is wrong at this place) removes the number/email  completely.

### Steps to reproduce

1. Create phone number/email with custom label
2. save contact
3. reload page

Plus for the real mess:
1. edit contact again
2. select work instead of private as label
--> the phone number has been removed completely.

### Expected behavior

Contacts with custom labels are displayed correctly.

### Actual behavior

See bug description

### Contact version

8.4.5

### Operating system

AIO container

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

pallavibakale:
- Fix custom label (`X-ABLABEL`) handling to correctly work with properties loaded from the server
- Prevent prefix corruption (e.g `NEXTCLOUD1.TEL`) when re-applying a custom label after a save/fetch cycle
- Fix `propGroup` and `propLabel` to detect the `group` parameter form that ical.js uses after parsing from server
image

Fixes: #5302


hamza221 (changes requested):
Hello, 
Thank you for opening this PR, 
The diff is really big and it looks like it's mostly spacing related fixes. 
Can you please run `npm ci` then `npm run lint:fix` then push again

pallavibakale:
> Hello, Thank you for opening this PR, The diff is really big and it looks like it's mostly spacing related fixes. Can you please run `npm ci` then `npm run lint:fix` then push again

Hello Hamza, I have fixed the linting issues, the PR looks cleaner than before. 
The DCO also failed, I ran the below to fix the issue:
```
git rebase HEAD~5 --signoff  
git push --force-with-lease origin fix/custom-label-for-contact-details-5302
```

Will do the same for the other PR as well.

hamza221:
while reviewing this I went down a rabbit hole and tried to use Ical.js 's native grouped properties handling 
https://github.com/nextcloud/contacts/pull/5560 would you mind giving it a look ?

pallavibakale:
> while reviewing this I went down a rabbit hole and tried to use Ical.js 's native grouped properties handling #5560 would you mind giving it a look ?

Agree with the diagnosis and the fix direction. I hit the same bug and tried fixing it by keeping both the old and new group formats around, but that just caused double-prefix issues (NEXTCLOUD1.NEXTCLOUD1.TEL). Fully switching to the group parameter like you did here is much cleaner.

Might be worth a test for properties that come in already grouped with a non-nextcloud prefix (like ITEM1. from iCloud/Google exports), just to confirm custom labels still work on those.

PR #5383, opened 2026-06-11
fix: adjust supported php versions

hamza221:


DerDreschner (approved):

ChristophWurst (commented):
33 is 8.2+ 👍 

Nit: this is not a chore change. You are effectively changing the product and what it supports. Per https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/release_process.html#minor-updates this will need a minor version bump.

PR #5384, opened 2026-06-11
chore(e2e-ci): pin node 24.15.0

hamza221:
context:  https://github.com/nextcloud/calendar/pull/8424 / https://github.com/nodejs/node/issues/63487

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -48,10 +48,10 @@ jobs:
           fallbackNode: '^24'
           fallbackNpm: '^11'
           path: nextcloud/apps/contacts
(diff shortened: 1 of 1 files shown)


DerDreschner:
Why not pinning it in `package.json` instead of manipulating the pipeline? Would that break something else?

hamza221:
> Why not pinning it in `package.json` instead of manipulating the pipeline? Would that break something else?

Could be annoying for devs with different node versions and afaik [this](https://github.com/nodejs/node/issues/63487) is our only issue with v>24.15.0 so doesn't make sense to restrict dev setups

DerDreschner (approved):

PR #5410, opened 2026-06-24
[stable8.3] chore: Don't run test pipeline for stable8.3 against master

DerDreschner:
`stable8.3` only supports Nextcloud 32. Anything else will break the pipeline and must therefore being removed to ensure backports can be merged smoothly.

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         php-versions: ['8.2']
         databases: ['sqlite']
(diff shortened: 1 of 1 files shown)


hamza221 (approved):

.github/workflows/phpunit.yml
hamza221: here

DerDreschner: Ahh, good one! Thought I've searched for all occurrences, but must have slipped through.

PR #5556, opened 2026-07-09
fix(router): redirect profile details view to internal contacts list

### LINKED ISSUE #5554: "View profile" for team members is broken
### Steps to reproduce

1. Go to any team
2. Click on any of the team members
3. Click on "View profile
4. The contacts app shows the default page and throws errors

Image

This is reproducible on c.nc.com

### Expected behavior

The address book entry for the user is being shown

### Actual behavior

Several errors are being thrown and the default address book with their first contact is being opened.

### Server operating system

_No response_

### Web server

_No response_

### Database engine version

_No response_

### PHP engine version

_No response_

### Nextcloud version

_No response_

### Updated from an older installed version or fresh install

_No response_

### Signing status

```Text

```

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Browser

_No response_

### Client operating system

_No response_

### CardDAV-clients

_No response_

### Web server error log

```Text

```

### Nextcloud log

```Text

```

### Browser log

```Text

```

### Browser log

```Text

```

silverkszlo:
Fixes https://github.com/nextcloud/contacts/issues/5554

## The bug
When inside the Contacts app one goes to  > Teams > some team > some member > "View profile", the user does not get navigated to the clicked user's profile but to the general "All contacts" list and the following errors appear:
Screenshot from 2026-07-09 08-56-10

## What seems to be the cause
- The "View profile" link is generated server-side and used by NcAvatar. And NcAvatar only checks if the link matches any route in the current app's router.
- Contacts app's router uses `:selectedGroup/:selectedContact` with no literal prefix, so it matches any two-segment path and treats `/u/alice` like an internal path (but it's not)
- To navigate to the Contact's internal contacts list it should navigate to `cloud.host.com/apps/contacts/All%20contacts/[someLongbase64String]`
Screenshot from 2026-07-09 09-30-44

## The solution
- Add a profile-redirect in Contacts app's router for `u/:userId` and build the correct url

That solution seems like a workaround though. Maybe `:selectedGroup/:selectedContact` should be less permissive or the NcAvatar needs an adjustment?

## 🤖 AI (if applicable)

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -4,9 +4,10 @@
  */
 
 import { generateUrl } from '@nextcloud/router'
+import { Buffer } from 'buffer'
 import { createRouter, createWebHistory } from 'vue-router'
 import Contacts from '../views/Contacts.vue'
-import { ROUTE_CHART, ROUTE_CIRCLE, ROUTE_USER_GROUP } from '../models/constants.ts'
+import { GROUP_ALL_CONTACTS, ROUTE_CHART, ROUTE_CIRCLE, ROUTE_USER_GROUP } from '../models/constants.ts'
 
 // if index.php is in the url AND we got this far, then it's working:
 // let's keep using index.php in the url
@@ -56,6 +57,17 @@ export default createRouter({
 					path: `${ROUTE_USER_GROUP}/:selectedUserGroup`,
 					name: 'user_group',
 					component: Contacts,
+				},
+				{
+					path: 'u/:userId',
+					name: 'profile-redirect',
+					redirect: (to) => ({
+						name: 'contact',
+						params: {
+							selectedGroup: GROUP_ALL_CONTACTS,
+							selectedContact: Buffer.from(`${to.params.userId}~z-server-generated--system`, 'utf8').toString('base64'),
+						},
+					}),
 				},
 			],
 		},
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -4,10 +4,10 @@
  */
 
 import { generateUrl } from '@nextcloud/router'
-import { Buffer } from 'buffer'
 import { createRouter, createWebHistory } from 'vue-router'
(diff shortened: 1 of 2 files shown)


src/router/index.js
```diff
@@ -57,6 +58,17 @@ export default createRouter({
...
 					component: Contacts,
 				},
+				{
+					path: 'u/:userId',
+					name: 'profile-redirect',
+					redirect: (to) => ({
+						name: 'contact',
+						params: {
+							selectedGroup: GROUP_ALL_CONTACTS,
+							selectedContact: Buffer.from(`${to.params.userId}~z-server-generated--system`, 'utf8').toString('base64'),
```
cristianscheid: I see the logic to generate the contact key is the same from [src/models/contact::key() method](https://github.com/nextcloud/contacts/blob/main/src/models/contact.js#L214).

To avoid duplication, maybe we could extract the logic into a function and use it in both places? Maybe something like:
```js
// src/models/contact.js

{...}
export function generateContactKey(uid, addressbookId) {
	return Buffer.from(uid + '~' + addressbookId, 'utf8').toString('base64')
}
{...}

export default class Contact {
	{...}
	get key() {
		return generateContactKey(this.uid, this.addressbook.id)
	}


// src/router/index.js

{...}
import { generateContactKey } from '../models/contact.js'
{...}
export default createRouter({
{...}
				{
					path: 'u/:userId',
					name: 'profile-redirect',
					redirect: (to) => ({
						name: 'contact',
						params: {
							selectedGroup: GROUP_ALL_CONTACTS,
							selectedContact: generateContactKey(to.params.userId, 'z-server-generated--system'),
						},
					}),
				},
```

silverkszlo: pushed it, thank you!

cristianscheid (approved):

PR #5565, opened 2026-07-12
fix(contacts): show photos that have no TYPE parameter

### LINKED ISSUE #5401: Contact photos with PHOTO;ENCODING=b: are not displayed in Contacts app (Nextcloud 34)
### Describe the bug

After upgrading/migrating to Nextcloud 34, most contact photos stopped appearing in the Contacts app. The contacts themselves were still present and synchronized correctly via CardDAV.

The photos continued to work in other CardDAV clients such as eM Client and iPhone Contacts, but Nextcloud Contacts only displayed initials instead of photos.

The browser console repeatedly showed:

```text
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
at getPhotoUrl()
```

Investigation showed that affected contacts contained photo entries such as:

```vcf
PHOTO;ENCODING=b:
iVBORw0KGgoAAAANSUhEUgAA...
```

without an explicit `TYPE=` parameter.

Contacts whose photo entries contained an explicit type, for example:

```vcf
PHOTO;ENCODING=b;TYPE=png:
```

or

```vcf
PHOTO;ENCODING=b;TYPE=JPEG:
```

were displayed correctly.

### Possible origin of the issue

One observation may help narrow down the root cause:

The affected contacts were synchronized for years using eM Client and CardDAV. After editing a contact photo in eM Client and synchronizing it again, the contact immediately started working in Nextcloud Contacts.

This suggests that eM Client may currently write photo entries including an explicit type, for example:

PHOTO;ENCODING=b;TYPE=PNG:

while many existing contacts in the address book were stored as:

PHOTO;ENCODING=b:

without a type.

It is unclear where the TYPE parameter was removed. Possible explanations include:

Older versions of Nextcloud Contacts / DAV may have normalized photo fields and omitted the type parameter because it was not required at the time.
A migration or import/export process may have removed the parameter.
Different CardDAV implementations may store equivalent photo data differently.

I cannot determine exactly when the TYPE parameter disappeared, but all affected contacts shared the same characteristic: the photo entry lacked an explicit type.

The fact that adding TYPE=png immediately restores photo display suggests that the current Contacts app may no longer handle such entries correctly.

### Steps to reproduce

1. Create or import a contact containing a photo entry without an explicit type:

```vcf
PHOTO;ENCODING=b:
iVBORw0KGgoAAAANSUhEUgAA...
```

2. Open the contact in the Contacts app.

3. Observe that the contact photo is not displayed.

4. Export the contact and modify the photo entry to:

```vcf
PHOTO;ENCODING=b;TYPE=png:
iVBORw0KGgoAAAANSUhEUgAA...
```

5. Delete the contact and re-import the modified vCard.

6. Reload the Contacts app.

Result: the contact photo immediately appears.

### Expected behavior

The Contacts app should correctly display contact photos stored as:

```vcf
PHOTO;ENCODING=b:
```

even when no explicit `TYPE=` parameter is present.

At minimum, the application should gracefully handle missing photo types and not throw JavaScript exceptions.

These contacts are successfully interpreted by other CardDAV clients (eM Client and iPhone Contacts), and older Nextcloud installations previously displayed the photos correctly.

### Actual behavior

Contacts containing photo entries without an explicit `TYPE=` parameter do not display a photo in the Contacts app.

Instead, only the contact initials are shown.

The browser console reports:

```text
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
at getPhotoUrl()
```

Adding an explicit photo type (for example `TYPE=png`) and re-importing the contact immediately restores the photo display.

In my case, approximately 150 contact photos were affected. After adding the missing `TYPE=png` parameter to the vCard entries and re-importing the contacts, all photos were displayed correctly again.

### Contact version

8.7.1

### Operating system

Unraid (Docker)

### PHP engine version

_No response_

### Web server

Apache (supported)

### Database

MySQL

### Additional info

_No response_

MiMoHo:
## Summary

Contact photos stored without an explicit `TYPE` parameter, e.g.

```
PHOTO;ENCODING=b:
iVBORw0KGgoAAAANSUhEUgAA...
```

were not displayed: `getPhotoUrl()` called `photoType.toLowerCase()` on the missing parameter and threw

```
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
```

so the app fell back to initials, while other CardDAV clients (eM Client, iOS Contacts) display these photos fine. The `TYPE` parameter is optional per RFC 2426/6350.

### Changes

* When the `TYPE` parameter is missing, detect the image type from the magic bytes of the base64 data (png, jpeg, gif, webp, bmp, svg). SVG detected this way still goes through the existing `sanitizeSVG` path.
* Unknown signatures fall back to `jpeg`: browsers content-sniff raster images in an `img` element, so a wrong subtype still renders — before this change the app just threw.
* Drive-by: the error logs in `getPhotoUrl()` referenced `this.contact`, which does not exist on the `Contact` class and always logged `undefined`; they now log `this`.

## Checklist

- [x] Jest tests added: photo with explicit `TYPE`, binary photo without `TYPE` (png + jpeg magic bytes), vCard 4.0 data-URI photo
- [x] Full Jest suite passes, `npm run lint` clean for the touched files, production build succeeds

Resolve #5401

### DIFF
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -40,9 +40,13 @@ function detectImageType(photoB64Data) {
 		PD94b: 'svg+xml', // '<?xml'
 	}
 	const signature = Object.keys(signatures).find((signature) => photoB64Data.startsWith(signature))
-	// browsers detect raster images in an img element from the content,
-	// so a wrong subtype still renders fine
-	return signature ? signatures[signature] : 'jpeg'
+	if (!signature) {
+		// browsers detect raster images in an img element from the content,
+		// so an assumed subtype still renders fine, but warn about the guess
+		console.warn('Could not detect the photo type from its content, assuming JPEG')
+		return 'jpeg'
+	}
+	return signatures[signature]
 }
 
 export const ContactKindProperties = ['KIND', 'X-ADDRESSBOOKSERVER-KIND']
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -22,6 +22,29 @@ function isEmpty(value) {
 	return (Array.isArray(value) && value.join('') === '') || (!Array.isArray(value) && value === '')
 }
 
+/**
+ * Detect the image type from the magic bytes of base64 encoded image data
+ *
+ * @param {string} photoB64Data the base64 encoded image data
+ * @return {string} the image type, jpeg if unknown
+ */
+function detectImageType(photoB64Data) {
+	// base64 encodings of the magic bytes of the common image formats
+	const signatures = {
+		'/9j/': 'jpeg',
+		iVBOR: 'png',
+		R0lGO: 'gif',
+		UklGR: 'webp',
+		Qk: 'bmp',
+		PHN2Z: 'svg+xml', // '<svg'
+		PD94b: 'svg+xml', // '<?xml'
+	}
+	const signature = Object.keys(signatures).find((signature) => photoB64Data.startsWith(signature))
+	// browsers detect raster images in an img element from the content,
+	// so a wrong subtype still renders fine
+	return signature ? signatures[signature] : 'jpeg'
+}
+
 export const ContactKindProperties = ['KIND', 'X-ADDRESSBOOKSERVER-KIND']
 
 export const MinimalContactProperties = [
@@ -271,13 +294,19 @@ export default class Contact {
 			photoType = photoB64.split(';')[0].split('/').pop()
 		}
 
+		// The TYPE parameter is optional (e.g. `PHOTO;ENCODING=b:…`),
+		// so fall back to the magic bytes of the image data (see issue #5401)
+		if (!photoType) {
+			photoType = detectImageType(photoB64Data)
+		}
+
 		// Verify if SVG is valid
 		if (photoType.toLowerCase().startsWith('svg')) {
 			const imageSvg = atob(photoB64Data)
 			const cleanSvg = await sanitizeSVG(imageSvg)
 
 			if (!cleanSvg) {
-				console.error('Invalid SVG for the following contact. Ignoring...', this.contact, { photoB64, photoType })
+				console.error('Invalid SVG for the following contact. Ignoring...', this, { photoB64, photoType })
 				return false
 			}
 		}
@@ -287,7 +316,7 @@ export default class Contact {
 			const blob = b64toBlob(photoB64Data, `image/${photoType}`)
 			return URL.createObjectURL(blob)
 		} catch {
-			console.error('Invalid photo for the following contact. Ignoring...', this.contact, { photoB64, photoType })
+			console.error('Invalid photo for the following contact. Ignoring...', this, { photoB64, photoType })
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/models/contact.js
```diff
@@ -22,6 +22,29 @@ function isEmpty(value) {
...
+		R0lGO: 'gif',
+		UklGR: 'webp',
+		Qk: 'bmp',
+		PHN2Z: 'svg+xml', // '<svg'
+		PD94b: 'svg+xml', // '<?xml'
+	}
+	const signature = Object.keys(signatures).find((signature) => photoB64Data.startsWith(signature))
+	// browsers detect raster images in an img element from the content,
+	// so a wrong subtype still renders fine
+	return signature ? signatures[signature] : 'jpeg'
```
DerDreschner: @SebastianKrupinski: I don't know if we should do such a fallback here. I think adding a warning or something like this would make more sense?

DerDreschner:
Hey @MiMoHo, thank you very much for your first PR here! Looks pretty good and useful to me. But additional colleagues will have a look into it as well.

src/models/contact.js
```diff
@@ -22,6 +22,33 @@ function isEmpty(value) {
...
 
+/**
+ * Detect the image type from the magic bytes of base64 encoded image data
+ *
+ * @param {string} photoB64Data the base64 encoded image data
+ * @return {string} the image type, jpeg if unknown
+ */
+function detectImageType(photoB64Data) {
+	// base64 encodings of the magic bytes of the common image formats
+	const signatures = {
```
DerDreschner, line 33: We only allow upload of PNGs, GIF, SVG, JPEG, WebP and AVIF. Could you please change the signatures accordingly? See https://github.com/nextcloud/contacts/blob/61c225e54b65c7118bc2b47c3ea5efcc8d3948da/src/components/ContactDetails/ContactDetailsAvatar.vue#L332

DerDreschner (changes requested):
Please change the image types acceptable for displaying. Once that is done, it get's a go from my side.

Thanks!!

src/models/contact.js
```diff
@@ -22,6 +22,33 @@ function isEmpty(value) {
...
+		UklGR: 'webp',
+		Qk: 'bmp',
+		PHN2Z: 'svg+xml', // '<svg'
+		PD94b: 'svg+xml', // '<?xml'
+	}
+	const signature = Object.keys(signatures).find((signature) => photoB64Data.startsWith(signature))
+	if (!signature) {
+		// browsers detect raster images in an img element from the content,
+		// so an assumed subtype still renders fine, but warn about the guess
+		console.warn('Could not detect the photo type from its content, assuming JPEG')
```
hamza221, line 46: console is not accepted by out linter anymore. Please use logger instead. 
See : https://github.com/nextcloud/contacts/pull/5624

DerDreschner:
Hey @MiMoHo, are you able to implement the requested changes? Or should we do the last steps for you? Thanks for the feedback!

PR #5566, opened 2026-07-12
fix(settings): read social sync initial state as boolean

### LINKED ISSUE #5561: Wrong comparisons
After babe45a5ac16bf877fc2fee28d2df38d094f6874

https://github.com/nextcloud/contacts/blob/961dceb58204837b2bca3f28dd62819ad616ea01/src/components/AppNavigation/ContactsSettings.vue#L79-L80

and

https://github.com/nextcloud/contacts/blob/961dceb58204837b2bca3f28dd62819ad616ea01/src/components/AdminSettings.vue#L29

values fetched with `loadState` are already bool probably and should not be compared with strings but

```
diff --git a/src/components/AdminSettings.vue b/src/components/AdminSettings.vue
index 0d79be5..0d791de 100644
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -26,7 +26,7 @@ export default {
        name: 'AdminSettings',
        data() {
                return {
-                       allowSocialSync: loadState('contacts', 'allowSocialSync') === 'yes',
+                       allowSocialSync: loadState('contacts', 'allowSocialSync'),
                }
        },
 
diff --git a/src/components/AppNavigation/ContactsSettings.vue b/src/components/AppNavigation/ContactsSettings.vue
index 8e5c651..5ce8c38 100644
--- a/src/components/AppNavigation/ContactsSettings.vue
+++ b/src/components/AppNavigation/ContactsSettings.vue
@@ -77,8 +77,8 @@ export default {
 
        data() {
                return {
-                       allowSocialSync: loadState('contacts', 'allowSocialSync') !== 'no',
-                       enableSocialSync: loadState('contacts', 'enableSocialSync') !== 'no',
+                       allowSocialSync: loadState('contacts', 'allowSocialSync'),
+                       enableSocialSync: loadState('contacts', 'enableSocialSync'),
                        enableSocialSyncLoading: false,
                        showSettings: false,
                }

```

MiMoHo:
## Summary

Since babe45a5ac16bf877fc2fee28d2df38d094f6874 the `PageController` provides `allowSocialSync` and `enableSocialSync` as **booleans** (`SocialApiService::syncAllowedByAdmin()` / `backgroundSyncEnabled()` return `bool`), but `ContactsSettings.vue` still compared the initial state with the strings `'yes'`/`'no'`:

```js
allowSocialSync: loadState('contacts', 'allowSocialSync') !== 'no',
enableSocialSync: loadState('contacts', 'enableSocialSync') !== 'no',
```

A boolean is never `=== 'no'`, so both flags were **always `true`**: the *Update avatars from social media* switch always appeared enabled, regardless of the stored setting.

### Changes

* Read both states as booleans, with fallbacks matching the server defaults (`allowSocialSync`: `true`, `enableSocialSync`: `false`).
* `allowSocialSync` was loaded into `data()` but no longer used anywhere since the settings dialog refactorings (it used to guard the switch, see 973c2882 / 7310a2a7). Put it back to work: the switch is now hidden when the administrator has disabled social sync, instead of showing users a toggle that has no effect (`backgroundSyncEnabled()` is `false` whenever `syncAllowedByAdmin()` is).

### Note on `AdminSettings.vue`

The issue also suggests changing `AdminSettings.vue`, but that would break it: its initial state comes from `lib/Settings/AdminSettings.php`, which provides the **raw string** app values (`'yes'`/`'no'`) from `Application::AVAIL_SETTINGS`, so the `=== 'yes'` comparison is correct there. Left unchanged on purpose.

## Checklist

- [x] Jest suite passes, `npm run lint` clean for the touched file (0 errors), production build succeeds
- [x] Manually traced both initial-state providers (`PageController` → booleans, `Settings/AdminSettings.php` → strings)

Resolve #5561

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -22,10 +22,6 @@ use OCP\EventDispatcher\IEventDispatcher;
 class Application extends App implements IBootstrap {
 	public const APP_ID = 'contacts';
 
-	public const AVAIL_SETTINGS = [
-		'allowSocialSync' => 'yes',
-	];
-
 	public function __construct() {
 		parent::__construct(self::APP_ID);
 	}
--- a/lib/Settings/AdminSettings.php
+++ b/lib/Settings/AdminSettings.php
@@ -8,25 +8,16 @@
 namespace OCA\Contacts\Settings;
 
 use OCA\Contacts\AppInfo\Application;
+use OCA\Contacts\Service\SocialApiService;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\AppFramework\Services\IInitialState;
-use OCP\IConfig;
 use OCP\Settings\ISettings;
 
 class AdminSettings implements ISettings {
-	protected $appName;
-
-	/**
-	 * Admin constructor.
-	 *
-	 * @param IConfig $config
-	 * @param IL10N $l
-	 */
 	public function __construct(
-		private IConfig $config,
 		private IInitialState $initialState,
+		private SocialApiService $socialApiService,
 	) {
-		$this->appName = Application::APP_ID;
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


hamza221:
I added the fix to `AdminSettings.vue` too

hamza221 (approved):
LGTM thank you

DerDreschner (approved):

DerDreschner:
Maybe change the default values mentioned on the comments for the `syncAllowedByAdmin` and `backgroundSyncEnabled` methods as well.

hamza221:
sorry for taking over you PR @MiMoHo  I hope you're okay with it. 
Your fix was actually enough but it showed me an inconsistency in our code and it made sense to fix it in the same PR/
Thank you for the fix 😊

DerDreschner:
Still looks good! :+1:

Issue #5403, opened 2026-06-22
Remove teams UI code from contacs


alimmroth:
### Use Case
For the TeamPage to get its own frontend, as described in #2543 and to move forward, it should become untangled from the contacts app.

### Task
Remove the teams functionality from contacts, as it has been moved to it's own app in https://github.com/nextcloud/circles/pull/2561

PR #5595, opened 2026-07-28
fix: panel to accept or refuse invitation

### LINKED ISSUE #2406: Invited member cannot accept/refuse the invitation
Selection_055

When invited into a Circles that have checked the option `Members have to accept invitation`, on the recipient side there is no way to accept/refuse the invitation (beside the notification).

The page have the same display as if the recipient was the one requesting a membership to the Circle:

- His/her membership is marked as 'Pending'
- A message saying that `Your request to join this circle is pending approval`

The correct behavior would be to display a better message about 'Your have been invited [...]`, and in the member list a wa to accept/refuse the invitation.

This is the interesting part of the JSON returned by the backend when retrieving the member list:

Selection_056

madbob:
sample

Fixes #2406

### DIFF
--- a/src/components/CircleDetails.vue
+++ b/src/components/CircleDetails.vue
@@ -191,9 +191,38 @@
 			<div class="circle-details__main-content">
 				<!-- not a member -->
 				<template v-if="!circle.isMember">
+					<!-- Pending invitation validation -->
+					<div v-if="circle.isInvitedMember" class="pending-invitation">
+						<h2>
+						{{ t('contacts', 'You have to accept or refuse the invitation to join this team') }}
+						</h2>
+
+						<div class="buttons">
+							<NcButton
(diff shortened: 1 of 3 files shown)


cristianscheid:
Hi @madbob, thanks for taking the time to put this PR together, really appreciate the effort.

We recently moved the Teams UI code from the Contacts app into the Teams app (`circles`), as you can see here:
- https://github.com/nextcloud/contacts/issues/5403
- https://github.com/nextcloud/contacts/pull/5635

Since those files moved, the changes introduced in this PR should now be opened against the circles repository. Feel free to open a PR there with this fix, you can find the repo here: https://github.com/nextcloud/circles/

I'll close this PR for now since it no longer applies, but thanks again for the contribution!

PR #5601, opened 2026-07-28
fix: label for read-only dates

madbob:
Totally unrequested graphical fix for read only date properties.

sample


hamza221 (approved):

hamza221:
LGTM but iirc this was added for a certain edge case by @GVodyanov or am I mistaken ?

PR #5602, opened 2026-07-28
fix: removed references to v-tooltip directive

madbob:
The `Tooltip` directive has been deprecated in `@nextcloud/vue` [in April 2025](https://github.com/nextcloud-libraries/nextcloud-vue/commit/88b1ff51aad493e79e7e0bfe20f5038766a8f42a) and actually no longer inited in Contacts since the migration to Vue3.

Still, there are a few references generating tons of errors in the JS console.

Highlights:
- removed #limit (including `v-tooltip`) and #no-options slots from the `NcSelect` of `PropertyGroups`: those were valid slots for the `NcMultiselect` components, replaced in 4205d1e with `NcSelect`
- using plain `title` in `SettingsAddressbook` for the address books sharing button

### DIFF
--- a/src/components/Properties/PropertyGroups.vue
+++ b/src/components/Properties/PropertyGroups.vue
@@ -31,15 +31,6 @@
 					tag-placeholder="create"
 					@option:deselected="updateValue"
 					@close="updateValue">
-					<!-- show how many groups are hidden and add tooltip -->
-					<template #limit>
-						<span v-tooltip.auto="formatGroupsTitle" class="multiselect__limit">
-							+{{ localValue.length - 3 }}
-						</span>
-					</template>
-					<template #no-options>
-						<span>{{ t('contacts', 'No results') }}</span>
-					</template>
 				</NcSelect>
 				<div v-else>
 					<span v-if="localValue.length === 0">{{ t('contacts', 'None') }}</span>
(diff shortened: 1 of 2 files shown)


src/components/Properties/PropertyGroups.vue
```diff
@@ -31,15 +31,6 @@
 					tag-placeholder="create"
 					@option:deselected="updateValue"
 					@close="updateValue">
-					<!-- show how many groups are hidden and add tooltip -->
-					<template #limit>
-						<span v-tooltip.auto="formatGroupsTitle" class="multiselect__limit">
-							+{{ localValue.length - 3 }}
-						</span>
-					</template>
-					<template #no-options>
```
hamza221, line 40: this still exists https://vue-select.org/api/slots.html#no-options

hamza221 (commented):
Just one comment then good to go ✅

src/components/Properties/PropertyGroups.vue
madbob: You are right, but this does the same thing than the default NcSelect implementation
https://github.com/nextcloud-libraries/nextcloud-vue/blob/main/src/components/NcSelect/NcSelect.vue#L552
Is it worth to keep?

hamza221: You're right, mb

hamza221 (approved):

PR #5631, opened 2026-08-09
fix: address field layout

madbob:
Totally unrequested fix of the unaligned Address field in Edit mode.

address


hamza221 (dismissed):
I rebased on top of main and built 

Image

PR #5632, opened 2026-08-09
chore: implement rs pack

SebastianKrupinski:
### Summary
- Implemented rs pack compiler
- implemented css extraction

### Results
Before:
Compile time: 11.5s
Compile size: 3.3mb

image

After: 
Compile time: 4.2s
Compile size: 3.09mb

image


rspack.config.js
```diff
@@ -0,0 +1,238 @@
...
+				__VUE_PROD_DEVTOOLS__: false,
+				__VUE_PROD_HYDRATION_MISMATCH_DETAILS__: false,
+			}),
+			new IgnorePlugin({
+				resourceRegExp: /^\.\/locale$/,
+				contextRegExp: /moment$/,
+			}),
+			new CssExtractRspackPlugin({
+				filename: '../css/contacts-[name].css',
+				chunkFilename: '../css/[id].chunk.css',
```
odzhychko, line 219:
```suggestion
				chunkFilename: '../css/chunks/[id].chunk.css',
```

nit(non-blocking): To be consistent with https://github.com/nextcloud/calendar/blob/fb819ae3762048fe6c5069b48dd2008bb311fd07/rspack.config.js#L233

rspack.config.js
```diff
@@ -0,0 +1,238 @@
...
+const minBrowserVersion = browsers
+	.map((str) => str.split(' '))
+	.reduce((minVersion, [browser, version]) => {
+		minVersion[browser] = minVersion[browser] ? Math.min(minVersion[browser], parseFloat(version)) : parseFloat(version)
+		return minVersion
+	}, {})
+const targets = Object.entries(minBrowserVersion).map(([browser, version]) => `${browser} >=${version}`).join(',')
+
+const transpilePackages = [
+	'p-limit',
```
odzhychko, line 28: note(non-blocking): `p-queue` and `p-try` are still transitive dependencies.

Though I do not understand why we only transpile this dependencies 🤷 

See https://github.com/nextcloud/calendar/blob/fb819ae3762048fe6c5069b48dd2008bb311fd07/rspack.config.js#L27-L33

odzhychko: I mean, shouldn't we transpile everything for `targets` (derived from `minBrowserVersion`)?

Maybe `SwcJsMinimizerRspackPlugin` transpiles it down to `targets`.

Then why do we have `transpilePackages` anyway. Seems like that would only affect Dev mode and Devs probably use recent browsers anyway.

package-lock.json
odzhychko, line 1: issue: For some reason `resolved` and `integrity` are missing.

Maybe because private registry/cache/proxy where used when installing.
Can be fixed by running `npx npm-package-lock-add-resolved`.

Anyway conflicting package-lock.json‎

odzhychko (approved):
Tested the entry points. They work as expected.

package-lock.json needs to be remade.

PR #5639, opened 2026-08-12
feat(contacts): support pronouns property

### LINKED ISSUE #5108: Support pronouns for contacts

See https://github.com/nextcloud/server/issues/40017. For users, we can already set pronouns. But for (external) contacts, I didn't find a way.

https://github.com/nextcloud/contacts/issues/4342#issuecomment-2676266739 mentioned modern alternatives to RFC6350 that support this.
So I guess implementing one of these would be a good way to resolve my issue.

 Yes, there is a 'gender' field but sometimes I know the pronouns a contact prefers, but not what gender they identify as (*e.g.* does 'she/them' imply 'female'?). Just setting those to 'other' gender might loose that information. I can add it in the notes, but when writing a batch-mail to contacts, I would like to be able to programatically pick the corresponding prefered pronouns.

### Additional context

Note that neither I want to start a debate about gender and pronouns nor do I want to engange in gender studies here. I just consider it polite to address contacts the way they prefer and would like to have that information in my address book in a unified way.

xhon-pelushi:
## Summary
- Add RFC 9554 `PRONOUNS` as a repeatable text contact property.
- Include it in lightweight CardDAV property fetches and cover serialization with Jest.

Closes #5108

## Test plan
- [x] Jest: rfcProps + contact model tests
- [x] `npm run build`
- [ ] UI: add/edit pronouns on a contact and verify round-trip save/reload

### DIFF
--- a/src/models/contact.js
+++ b/src/models/contact.js
@@ -26,7 +26,7 @@ function isEmpty(value) {
 export const ContactKindProperties = ['KIND', 'X-ADDRESSBOOKSERVER-KIND']
 
 export const MinimalContactProperties = [
-	'EMAIL', 'UID', 'TEL', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME', 'X-MANAGERSNAME', 'TITLE', 'NOTE', 'RELATED',
+	'EMAIL', 'UID', 'TEL', 'CATEGORIES', 'FN', 'ORG', 'N', 'X-PHONETIC-FIRST-NAME', 'X-PHONETIC-LAST-NAME', 'X-MANAGERSNAME', 'TITLE', 'NOTE', 'RELATED', 'PRONOUNS',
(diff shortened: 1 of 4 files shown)


GVodyanov:
@hamza221 This respects the RFC right? Because code looks good

EnbyAce:
I think it respects the RFC, based on the changes to the `tests/javascript/models/rfcProps.test.js` file. The pronouns field should match `PRONOUNS:they/them` exactly, which does respect [RFC 9554](https://datatracker.ietf.org/doc/rfc9554/).

I don't know how Nextcloud's Contacts app serializes the properties, but RFC9554 also mentions some optional parameters that don't seem to be mentioned anywhere in the commit:
- `LANGUAGE`
- `PREF`
- `TYPE`
- `ALTID`

hamza221:
It doesn't look like there's a way to make this interoperable. google contacts doesn't seem to have a pronouns field and apple doesn't save it to vcf and doesn't read it either. 
Any ideas about other clients that support this property ?

Issue #3273, opened 2023-03-13
Custom Fields


darkdecoy:
### Is your feature request related to a problem? Please describe.

We have a client that is looking at moving from Contactzilla to Nextcloud.  One thing that is missing is the ability to create custom fields for contacts.  This is something they used extensively on Contactzilla and would be nice to have on Nextcloud Contacts.  Thanks.

### Describe the solution you'd like

Right now you can add additional fields of a variety of types.  I would like to be able to create a general text field and give it a custom name.

gopher333:
You can do that by renaming the mime Types for the given fields in Nextcloud. But only so if you administrate the server yourself (if you have a hosted Server the Hoster will hardly give you permission to do so and fiddle around on Operating System Level).
And-everything is back to standard Fieldnames every time a package (here:Contacts) is updated.

christophjahn:
Please consider this comment as a vote for this enhancement/feature.

I was really surprised that custom fields are not possible. At least that is what my "research" suggests so far. In a business context this is a standard requirement in my experience. I am open to workarounds.

FelixKainz:
Yes, please. Custom fields are so useful in addition to tags.

SebastianKrupinski:
@hamza221

fberke:
I had the same need and took a stab at the config-based flavor of this: admin-defined custom field types, configured with a single `occ` command — text and select fields that show up in the contact editor like built-in ones and are stored as plain `X-…` vCard properties, so they survive sync with other CardDAV clients.

PR: #5659

No settings UI yet — that could be layered on top later if the approach fits. Feedback welcome.

Issue #303, opened 2017-08-27
Ability to create and remember custom fields


olantrust:
_Not to be confused with custom labels_ #27 These issues need two separate bugs.

Currently almost all common addressbook systems allow custom field creation in some way.
For example, a business user needs to create a field: customertype, and use it for every contact.
A political user needs a field: constituency, and an organisational user needs a date field: membership renewal.

Zimbra Gmail etc.. allow this, and Zimbra and Outlook allow the user to select the data type.

Our app should allow:
Create custom:

On selection, a pop-up appears to the User where they can set:
'Field name:' 
'Field type:' (single line, multiple lines, yes/no, date, url, number, random number, [maybe in future: sequential number]), 
'Occurs:'(single, or multiple per contact).
'Appears:' (each time on main screen, each time in 'Add field' dropdown, this time only)

**Example:**
User wants to create a 'member' field.
User selects 'Add custom'
Pop-up screen presents:
User enters:
Field name: Member
Field type: Yes/no
Occurs: Single
Appears: Every new contact

The new field will now appear for the current contact, and each new contact created, and the User can tick the Member:'Yes' or 'No' tickboxes (or leave blank for unknown).

skjnldsv:
I tend to aim for simplicity, but you have my attention! :)
Could you propose a mockup of what you suggest?
I'm not really sure how to implement this without breaking the nice fit we have between all of our properties.

jancborchardt:
Where does Gmail allow this? Can you post a screenshot?

olantrust:
![gmailcustom-screenshot_20170830_151041](https://user-images.githubusercontent.com/627140/29876742-77cc1388-8d95-11e7-94ee-c3d08ea3e1ba.png)

olantrust:
Can we remove the needsinfo label now, or is there anything else we can clarify? It's not clear what information is still needed. Thanks
@skjnldsv @jancborchardt

mhgithub20:
I need this too.  I also need CardDAV to be able to sync it to other systems.  I've found comments in other discussions indicating that the underlying core and CardDAV already support custom fields, but the web UI does not -- though I'm not in a position to verify that claim.  What is needed to transition this from a wish to an active project.  Do we need more votes in favor?  Do we need a bounty?

bpmartin20:
[...]

PR #5659, opened 2026-08-24
feat(contacts): admin-defined custom property types via app config

fberke:
### Summary

Allows admins to define custom property types (custom fields) for the contact editor via app config — no UI, a single `occ` command:

```
occ config:app:set contacts customProperties --value='[
  {"name": "x-customernumber", "label": "Customer number"},
  {"name": "x-region", "label": "Region", "force": "select",
   "options": [{"id": "NORTH", "name": "North"}, {"id": "SOUTH", "name": "South"}]}
]'
```

The fields show up in the "Add more info" menu and in the contact details, read-only and merge views like built-in fields, and are stored as regular `X-…` vCard properties, so they survive sync with other CardDAV clients.

Implements the config-based flavor of #3273 / #303.

### How it works

- A new `CustomPropertiesService` reads and validates the `customProperties` app config (JSON). Only `x-…` property names are accepted; invalid entries are dropped and logged, so a broken config can never break the app. The config schema is documented in the class docblock.
- `PageController` provides the sanitized list as initial state.
- The frontend merges the entries into the `rfcProps` registry at boot (`src/models/customProperties.js`), so all existing views pick them up without further changes.

Supported keys per entry: `label`, `force` (`text`, default, or `select`), `options` (TYPE parameter choices for text fields, value choices for select fields), `multiple`, `primary`, `icon`.

### Notes / limitations

- Labels are shown as configured (not translated) — they are admin-provided, instance-specific strings.
- Editor types are limited to `text` and `select` for now.
- Entries colliding with built-in property names are skipped.

### Testing

- Unit tests for the service (validation matrix) and the frontend merge function; `PageControllerTest` updated for the new constructor dependency.
- Manually tested on Nextcloud 35.0.0 beta 4: fields appear in the editor, values round-trip through save/reload and appear in the vCard export as `X-…` properties.

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -9,6 +9,7 @@ namespace OCA\Contacts\Controller;
 
 use OC\App\CompareVersion;
 use OCA\Contacts\AppInfo\Application;
+use OCA\Contacts\Service\CustomPropertiesService;
 use OCA\Contacts\Service\GroupSharingService;
 use OCA\Contacts\Service\SocialApiService;
 use OCP\App\IAppManager;
@@ -33,6 +34,7 @@ class PageController extends Controller {
 		private IAppManager $appManager,
 		private CompareVersion $compareVersion,
 		private GroupSharingService $groupSharingService,
+		private CustomPropertiesService $customPropertiesService,
 	) {
 		parent::__construct(Application::APP_ID, $request);
 	}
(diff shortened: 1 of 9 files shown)


SebastianKrupinski:
Hi @fberke 

Thank you for the PR. I had a quick look at the code, and noticed one thing, we require that any newly created front end files be created in typescript.

fberke:
Thanks for the quick look! Fair point — I converted `src/models/customProperties.js` and its test to TypeScript in bc86685a2, typed against the config schema. Kept it as a fixup commit for now; I'll squash once the review is through.

PR #5635, opened 2026-08-10
Removed teams UI

hiobjan:
Remove the teams functionality from contacts, as it has been moved to it's own app in https://github.com/nextcloud/circles/pull/2561

Closes #5403

## 🤖 AI (if applicable)

- [-] The content of this PR was partly or fully generated using AI

PR #5669, opened 2026-08-25
Fix/restore teams UI version gated

GVodyanov:
Reverts https://github.com/nextcloud/contacts/pull/5635 and makes it conditional, because 34 and lower still need the UI.

| 34 | 35 |
| - | - |
|  |  |

## 🤖 AI (if applicable)

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -4,7 +4,7 @@
  */
 
 import { Store } from 'vuex'
-import isCirclesEnabled from '../services/isCirclesEnabled.js'
+import isTeamManagementEnabled from '../services/isTeamManagementEnabled.js'
 import addressbooks from './addressbooks.js'
 import circles from './circles.js'
 import contacts from './contacts.js'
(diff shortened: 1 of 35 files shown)


src/store/index.js
```diff
@@ -20,8 +20,8 @@ const modules = {
 	importState,
 }
 
-// If circles is enabled let's init the store
-if (isCirclesEnabled) {
+// Only register the teams store when we manage teams ourselves
+if (isTeamManagementEnabled) {
```
GVodyanov, line 24: You may ask, why am I renaming this?

`isCirclesEnabled` used to check whether the circles "app" was installed and that's it, meanwhile `isTeamManagementEnabled` checks both if the app is installed and if the server version supports it. I figured seeing as it's a slightly different thing it would be best to separate the naming to not generate confusion

ChristophWurst (approved):
Linked the reverted PR in the PR description

PR #5688, opened 2026-08-29
fix: addressbook checkbox

GVodyanov:
Fixes this:

https://github.com/user-attachments/assets/ba5313e2-b81d-402a-ad07-4d89e2a6247e

## 🤖 AI (if applicable)

### DIFF
--- a/src/store/addressbooks.js
+++ b/src/store/addressbooks.js
@@ -108,25 +108,25 @@ const mutations = {
 	/**
 	 * Toggle whether a Addressbook is Enabled
 	 *
-	 * @param {object} context the store mutations
+	 * @param {object} state the store data
 	 * @param {object} addressbook the addressbook to toggle
 	 */
-	toggleAddressbookEnabled(context, addressbook) {
-		addressbook = state.addressbooks.find((search) => search.id === addressbook.id)
-		addressbook.enabled = !addressbook.enabled
+	toggleAddressbookEnabled(state, addressbook) {
+		const storedAddressbook = state.addressbooks.find((search) => search.id === addressbook.id)
+		storedAddressbook.enabled = !storedAddressbook.enabled
 	},
 
 	/**
 	 * Rename a Addressbook
 	 *
-	 * @param {object} context the store mutations
+	 * @param {object} state the store data
 	 * @param {object} data destructuring object
 	 * @param {object} data.addressbook the addressbook to rename
 	 * @param {string} data.newName the new name of the addressbook
 	 */
-	renameAddressbook(context, { addressbook, newName }) {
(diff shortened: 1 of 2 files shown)


src/store/addressbooks.js
```diff
@@ -108,25 +108,25 @@ const mutations = {
...
 	 * Toggle whether a Addressbook is Enabled
 	 *
-	 * @param {object} context the store mutations
+	 * @param {object} state the store data
 	 * @param {object} addressbook the addressbook to toggle
 	 */
-	toggleAddressbookEnabled(context, addressbook) {
-		addressbook = state.addressbooks.find((search) => search.id === addressbook.id)
-		addressbook.enabled = !addressbook.enabled
+	toggleAddressbookEnabled(state, addressbook) {
```
odzhychko, line 114: (A) Could not reproduce the described issue and (b) code changes seems to make a difference.

For A: I tested it without this change and could not reproduce the error you video shows.

For B: I tested also tested following:

```js
	toggleAddressbookEnabled(context, addressbook) {
		const addressbookFromStateNamedContext = context.addressbooks.find((search) => search.id === addressbook.id)
		const addressbookFromLexicalState = state.addressbooks.find((search) => search.id === addressbook.id)
		console.log("Compare", addressbook === addressbookFromStateNamedContext, addressbookFromLexicalState === addressbook, addressbookFromLexicalState === addressbookFromStateNamedContext)
		addressbookFromLexicalState.enabled = !addressbook.enabled
	},
```

The output was:

```
Compare true true true
```

So all address books are the same and this code change should make no difference...

odzhychko: Maybe this is an issue only observed with Vue 3?

odzhychko (changes requested):

GVodyanov:
Note: this only happens for addressbooks that were just created, forgot to mention

src/store/addressbooks.js
```diff
@@ -108,25 +108,25 @@ const mutations = {
...
-	 * @param {object} context the store mutations
+	 * @param {object} state the store data
 	 * @param {object} data destructuring object
 	 * @param {object} data.addressbook the addressbook to rename
 	 * @param {string} data.newName the new name of the addressbook
 	 */
-	renameAddressbook(context, { addressbook, newName }) {
-		addressbook = state.addressbooks.find((search) => search.id === addressbook.id)
-		addressbook.displayName = newName
+	renameAddressbook(state, { addressbook, newName }) {
```
odzhychko, line 127: Intentional?

odzhychko (approved):
> Note: this only happens for addressbooks that were just created, forgot to mention

Could reproduce now and the fix works.

PR #5708, opened 2026-09-07
fix(store): keep new contacts in the sorted index

### LINKED ISSUE #5158: Bug: Sort by "last modified" is not working

### LINKED ISSUE #5681: Contact not found error when pressing new contact button in Nextcloud Contacts App
### Describe the bug

Where is an error "contact not found" in Contacts web version

Even when I delete Address Book Contacts (it reappears by the way)  and trying to create contacts on empty book.

From unusual apps I use only Nextcloud Mail. I tried to disable it, but it didn't help

Nextcloud AIO version is **13.6.0**

If you need more logs or such I can collect them just told me from which service

### Steps to reproduce

1. Go to Contacts from menu
2. Click "+ New contact"
3. Where is Error Contact not found in interface (top right corner) instead of contact creation form

(I am trying to delete default Address Book Contacts and create new with my own name - same behavior) 

### Expected behavior

I see new contact form

### Actual behavior

I see contact not found error in interface (but my counter in settings in address book section increments with every New contact button click).

### Contact version

8.7.6

### Operating system

Debian GNU/Linux 13 (trixie)

### PHP engine version

PHP 8.3

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

When I delete my address book and default address book Contacts reappeared I usually can create some contacts before I refresh my page. After what new contact form opens no more and instead I can see "Contact not found error" and this error in console:

```
[ERROR] contacts: [Vue error]: Error in https://vuejs.org/error-reference/#runtime-m: Error: Debounced method called with different contexts of the same prototype. 
Object { app: "contacts", uid: "ethicalengineer", level: 2, error: Error, vm: Proxy, info: "https://vuejs.org/error-reference/#runtime-m" }
```

Yes I cleared my cache/cookies/localstorage

hamza221:
Fix #5681
Fix #5158

Assisted-by: ClaudeCode:claude-opus-5

## 🤖 AI (if applicable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contact sorting with consistent favorite prioritization and secondary ordering.
  * Preserved correct ordering when adding, updating, or replacing contacts in address books.
  * Fixed sorting behavior for large contact lists and contacts with irregular revision values.
  * Prevented errors when deleting contacts that are no longer present.

* **Tests**
  * Expanded coverage for contact insertion, updates, favorites, address book changes, deletion, and re-sorting across different contact fields.

### DIFF
--- a/src/store/contacts.js
+++ b/src/store/contacts.js
@@ -5,6 +5,7 @@
 
 import { showError } from '@nextcloud/dialogs'
 import ICAL from 'ical.js'
+import { toRaw } from 'vue'
 import Contact from '../models/contact.js'
 import logger from '../services/logger.js'
 import validate from '../services/validate.js'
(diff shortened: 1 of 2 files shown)


hamza221:
@coderabbitai review

GVodyanov:
Claude found this:

⚠️ fixes the reported bug, introduces two new ones

The old insertion loop had its condition inverted (sortData(other, sortedContact) >= 0 → continue), so it inserted before the first contact that sorts earlier, and dropped the contact entirely when every existing contact sorted after it. Confirmed in the live store on the parent commit and fixed by the PR:
```
┌─────────────────────────────────────────────────────┬──────────────────────────┬───────────────────────┐
│                      scenario                       │          parent          │        PR 5708        │
├─────────────────────────────────────────────────────┼──────────────────────────┼───────────────────────┤
│ new contact sorts before all existing (issue #5681) │ dropped from the index   │ present at index 0 ✅ │
├─────────────────────────────────────────────────────┼──────────────────────────┼───────────────────────┤
│ new contact sorts in the middle                     │ Name, Alice, Zoe (wrong) │ Alice, Name, Zoe ✅   │
├─────────────────────────────────────────────────────┼──────────────────────────┼───────────────────────┤
│ contact with an empty name                          │ dropped from the index   │ present, but first ⚠️ │
├─────────────────────────────────────────────────────┼──────────────────────────┼───────────────────────┤
│ sort by First name, N:…;Mary,Jane;;; (vCard 4.0)    │ works                    │ TypeError ❌          │
└─────────────────────────────────────────────────────┴──────────────────────────┴───────────────────────┘
```

1. Crash when sorting by First/Last name. Removing .toString() from the index entry means value is no longer always a string: contact.firstName is n[1], which ICAL returns as an array when a vCard 4.0 N has multiple given names. sortData then takes the .toUnixTime() branch → TypeError: b.value.toUnixTime is not a function. It's reachable through import, which parses the raw file and keeps VERSION:4.0 before upload. End-to-end with sorting set to First name:

- parent: importing N:Wayne;Bruce,Bat;;; → accepted: 1, contact appears.
- PR 5708: importing N:Parker;Peter,Pete;;; → accepted: 0, denied: 1, contact absent from the list — but a PROPFIND shows it was created on the server. createVCard succeeds, then the addContact commit throws inside the promise and the .catch counts it as denied. Silent data/UI divergence until reload.

(Contacts fetched from the server come back as vCard 3.0, where N components stay strings — that's why normal browsing doesn't hit it.)

2. Nameless contacts move from the bottom to the top. The old sortByFavoriteAndName pushed empty values last (if (!a.value) return 1); sortByFavoriteAndData delegates to sortData, where '' localeCompares first. Visible with real data: sorting by First name puts all 14 account contacts (empty given name) above the named ones, where the parent commit put them below.

3. Undocumented behavior change: "Last modified" flips direction. rev now stays an ICAL.VCardTime, so sortData uses nameB - nameA → newest first, where the parent string-compared the ISO date → oldest first. A contact with no REV mixes a string into that comparison and sorts inconsistently against the typed ones.

Worth noting on the plus side: the PR removes a real pre-existing inconsistency — sortContacts built values without .toLowerCase() while setContacts/addContact applied it, so the index held mixed-case entries after a sort-order change.

SebastianKrupinski (dismissed):

hamza221:
@coderabbitai review

SebastianKrupinski (approved):

hamza221:
@coderabbitai final review por favor 🙏🏼

hamza221:
@coderabbitai review please

PR #5724, opened 2026-09-14
fix(OCM invites): check for addressbook or lack thereof in route.

redblom:
The OCM invites routes lack the addressbook part. That Irregularly causes a `group not found` error when navigating from any of the Address book contacts lists to 'External invitations' list.

Screenshot from 2026-09-15 10-24-18
<br><br>
This PR fixes this issue by checking if `this.addressbook` is actually set in the relevant watcher.


src/views/Contacts.vue
```diff
@@ -398,7 +398,7 @@ const _default = {
 
 		// watch url change and address book select
 		selectedAddressbook() {
-			if (!this.isMobile && !this.selectedChart) {
+			if (this.addressbook && !this.isMobile && !this.selectedChart) {
```
hamza221, line 401: Is this leftover ? `this.addressbook`is always underfined

redblom: No it's the actual fix. The reason is as follows: The 'External invitations' route does not have the `addressbook` part so when navigating to it from any of the Address books lists `this.addressbook` changes to undefined and the `selectedAddressbook()` watcher is triggered. And then `selectFirstContactIfNone()` is called (causing the `group not found` error) instead of `selectFirstOcmInviteIfNone()`.

PR #5733, opened 2026-09-17
Feat: Make use of the Nextcloud default Mailtemplate for recognition.

T0mWz:
When an OCM invitation request is sent, it is currently a plaintext email. Since we will be exchanging data here and the outside world isn’t always trustworthy, we want to provide the user with some identifying features. It is therefore recommended to use the Nextcloud standard email template, which the user may already recognize from the service and which also displays the service’s logo, color, and name.

We are expanding the OCM invitation flow using the standard Nextcloud email template.
CurrentMailTemplate

The email now uses Nextcloud core's standard `IEMailTemplate` (same as sharebymail, IMipPlugin, NewUserMailHelper, etc.), so the branded header/footer, button styling, and light/dark-safe HTML come from core instead of hand-rolled `<br>` markup.

NextcloudMailtemplate

## 🤖 AI (if applicable)

- [X] The content of this PR was partly or fully generated using AI (N/A)
With use of Claude was this PR created


T0mWz:
@mickenordin , @redblom , FYI :-)

mickenordin (approved):
Looks good to me, although you need to sign off all commits, dco test is failing.

